PR #2, opened 2015-09-20
Initial import of FreeBSD bits

Sp1l:
Still need to merge the master branch into my changes
Perl changes are incompatible... `#!/usr/bin/env perl` would work I guess

Primarily to elicit feedback for now.


ronys:
A few of questions:
1. Is the env command in /usr/local/bin rather than /usr/bin? This defeats the value of using it in scripts...
2. What about packaging? I see that you copied the Linux Makefile as-is, and this supports both deb and rpm packaging. What does FreeBSD use?
3. For the os files, do you think we can use a POSIX directory instead of duplicating many of the files?

Sp1l:
On 2015-09-22 20:29, ronys wrote:
> A few of questions:
> 1. Is the env command in /usr/local/bin rather than /usr/bin? This
> defeats the value of using it in scripts...
> 2. What about packaging? I see that you copied the Linux Makefile
> as-is, and this supports both deb and rpm packaging. What does FreeBSD
> use?
> 3. For the os files, do you think we can use a POSIX directory instead
> of duplicating many of the files?
> 
> --
> Reply to this email directly or view it on GitHub [1].
> 
> 	*
> 
> Links:
> ------
> [1] https://github.com/pwsafe/pwsafe/pull/2#issuecomment-142373462

Hi Rony,

1. On FreeBSD there's a strict separation of the base OS and anything 
installed via ports. env is in the base system and therefor lives in 
/usr/bin

2. FreeBSD has ports (build your own) and packages (create your own or 
use binary). I will have to wrap this into a port (devel/pwsafe) which 
users can use to build or to download a binary package for using that 
name.

3. I was working on that already... Most of it is duplicated. I think I 
have it worked out pretty far now. did an EXPORT PLATFORM=FreeBSD in 
Makefile.freebsd so I can use this for ifeq ($(PLATFORM),FreeBSD) in 
sub-Makefiles, In C/CPP files I can rely on #ifdef __FreeBSD__ for 
detection.

Struggling a bit with the way that the port installs wxgtk2-3.0. It 
seems things are not in places where they are expected (compare 
wx-config), but I think I'm getting there. After the refactor using the 
linux dir and conditionals I just ran into some snag which I will deal 
with somewhere this weekend I hope, so I can push that to GitHub again!
File '../../../lib/unicoderelease/libos.a' does not exist.
     Must remake target '../../../lib/unicoderelease/libos.a'.
gmake[1]: *** No rule to make target 
'../../../lib/unicoderelease/libos.a', needed by 
'GCCUnicodeRelease/pwsafe'.  Stop.

Managed to "fix" the Yubikey dep (NO_YUBI=yes gmake)

Progress, but no cigar just yet!

Thanks for the feedback,

Bernard.

ronys:
Sounds good. Re env, I asked because Misc/check_po.py was changed from /usr/bin/env python to /usr/local/bin/env python. Mistake?

Sp1l:
Yes, I clobbered /usr/bin/env with a too wide sed -ie 's|/usr/bin|/usr/local/bin|'. Will fix that!

ronys:
Fine.

I'm planning to release a new version of PasswordSafe this Sunday. Any
chance that I'll be able to merge in your work by the?

Cheers,

  Rony

Sp1l:
I'll try getting something up. Do you want the os/linux dir refactored or just modified with some ifdefs for FreeBSD specific things?

I'll be travelling to EuroBSDcon as of tonight, should have some time on board trains, plane and busses. I'll let you know if I succeed!

ronys:
EuroBSDcon - oooh - take me with you :-)

I'd be much happier with refactoring to a new "os/posix" directory that the
os/linux, os/freebsd and perhaps os/mac Makefiles will reference to as
needed.

Enjoy the conference.

Rony

Sp1l:
Closing this one, opening a new one which is aligned with upstream master

PR #4, opened 2015-10-02
Update paths to match where they go in Debian

bblough:
Paths returned by path helper functions don't match Debian conventions.  I don't know how this will affect other distros, so it may or may not make sense to merge it.  Regardless, I'm providing it in case it proves useful.


ronys:
Thanks. This is currently incomplete, even for Debian. At the very least, install/deb/Makefile should be updated to place the relevant files in the new directories.
Same for install/rpm/pwsafe.spec.tmpl, install/slackware/pwsafe.SlackBuild

bblough:
Honestly, since I'm not using them to generate the official Debian packaging, I totally forgot about the makefiles, etc. under under install/.  I was just forwarding the patch we're using to adjust the paths in the code.

If this is something you think should be pursued, I'll go back through and make sure it's complete, even though it's not necessary for my packaging work.  Caveat: I have some other stuff I need to finish first, so it might be a week or three before I get a chance to do it.

bblough:
I've adjusted the install paths for deb/rpm/slack to work with the original patch.

ronys:
Looks better, thanks. Need to test what will happen when upgrading a package. Will the old (pwsafe) directories be orphaned? Any idea?

bblough:
When removing a package in Debian, the files installed by the package will be removed, then the directories installed by the package will be removed only if empty (to avoid deleting files installed by another package, or added/changed by a user).  My (untested) assumption is that upgrades happen the same way.  So I think it should be fine.

I would expect other distros to behave in a similar fashion, but I don't know that for certain.

pm-kan:
Slackware remove old empty folders on upgrade too. I'll update SlackBuild if/when pull request will be applied.

PR #5, opened 2015-10-02
Generate position independent code.  This fixes FTBFS on Debian

bblough:
0.95 and later fails to build from source on Debian systems.  Enabling position independent code (-fPIC) resolves the issue.


ronys:
Will merge this and previous after cross-checking against other Linux distros, 32 and 64 bit architectures.

Thanks!

ronys:
Tested on Debian 8.2, Ubuntu 14.04 and Fedora22, 32 and 64 bits, merging into master - thanks!

PR #7, opened 2015-10-08
Encryption Changes

jpvasquez:
This includes two commits for performance enhancements.

The first one https://github.com/pwsafe/pwsafe/commit/b88f60487806bcfddf2e22846e6622d7a470db8e simply adds syntax changes to avoid useless virtual indirection in leaf classes and adds const markers to the encryptors methods which effectively do not change its internal state. This last change will be needed on the next commit.

The second one https://github.com/pwsafe/pwsafe/commit/198b6cdb33d3dc501220c557252794bab5df54f6 has been made due to a finding in the profiler, which showed that most of the CPU time spent opening huge safes was creating and destroying equivalent Blowfish objects. On an iPhone 4s this change make a 1500-entries safe open in 0,5 seconds instead of 2,5, which is a huge UX gain.

### DIFF
--- a/src/core/BlowFish.cpp
+++ b/src/core/BlowFish.cpp
@@ -301,7 +301,7 @@ const uint32 BlowFish::tempbf_S[4][256] =
 };
 
 void BlowFish::Blowfish_encipher(uint32 *xl,
-                                 uint32 *xr)
+                                 uint32 *xr) const
 {
   union aword Xl;
   union aword Xr;
@@ -325,7 +325,7 @@ void BlowFish::Blowfish_encipher(uint32 *xl,
 }
 
 void BlowFish::Blowfish_decipher(uint32 *xl,
-                                 uint32 *xr)
+                                 uint32 *xr) const
 {
   union aword Xl;
   union aword Xr;
@@ -424,7 +424,7 @@ BlowFish::~BlowFish()
   trashMemory(bf_S, sizeof(bf_S));
 }
 
-void BlowFish::Encrypt(const unsigned char *in, unsigned char *out)
+void BlowFish::Encrypt(const unsigned char *in, unsigned char *out) const
 {
   for (int x = 0; x < 8; x++)
     out[x] = in[x];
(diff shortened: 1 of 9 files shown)


ronys:
1. Looks promising. The downside is that the Blowfish objects are around slightly longer than necessary, but in the context of entry fields, this isn't really an issue.
2. 'virtual' in the leaf class's function declarations is a matter of taste. I usually leave them in in order to clearly indicate that the function's virtual.
3. More importantly, this fails to build under Windows:

1>TwoFish.cpp(497): error C2440: '=' : cannot convert from 'const uint32 [256]' to 'uint32 *'
1>          Conversion loses qualifiers
1>TwoFish.cpp(498): error C2440: '=' : cannot convert from 'const uint32 [256]' to 'uint32 *'
1>          Conversion loses qualifiers
1>TwoFish.cpp(499): error C2440: '=' : cannot convert from 'const uint32 [256]' to 'uint32 *'
1>          Conversion loses qualifiers
1>TwoFish.cpp(500): error C2440: '=' : cannot convert from 'const uint32 [256]' to 'uint32 *'
1>          Conversion loses qualifiers
1>TwoFish.cpp(567): error C2440: '=' : cannot convert from 'const uint32 [256]' to 'uint32 *'
1>          Conversion loses qualifiers
1>TwoFish.cpp(568): error C2440: '=' : cannot convert from 'const uint32 [256]' to 'uint32 *'
1>          Conversion loses qualifiers
1>TwoFish.cpp(569): error C2440: '=' : cannot convert from 'const uint32 [256]' to 'uint32 *'
1>          Conversion loses qualifiers
1>TwoFish.cpp(570): error C2440: '=' : cannot convert from 'const uint32 [256]' to 'uint32 *'
1>          Conversion loses qualifiers

Looking at how to fix this. Suggestions welcome.

ronys:
Found a simple fix for Windows build. Testing, and will pull into master once tests pass. Thanks!

jpvasquez:
Thank you! I'll rebase and send you the other changes I made over time one by one, so that you can incorporate the modifications which are universal improvements. 

Regarding the removal of `virtual` marker, this brings a (most of the time negligible, I know) performance gain, due to the fact that the compiler can omit the indirection and even inline functions when it knowns the type of the object to be a leaf type. That's why my personal preference is to only add `virtual`  when needed, removing it from leaf types.

ronys:
Thanks!

If you could send your modifications against the v4/2015-10-12 branch, that would be great! My plan is to merge this into master once the basic attachment functionality is working (almost there!), and release the next v3 version with V4 as an exportable format (e.g., not the default, and clearly marked alpha/experimental/use-at-your-own-risk).

PR #11, opened 2015-10-14
1193: Ctrl-F would set focus to search textbox if searchbar is already visible

sauravg:
Only 233c37c is required to fix the focus issue.  The rest are required to build with CodeLite IDE (I guess I alone use it).


ronys:
Tested the focus fix, accepted into master - thanks!

PR #13, opened 2015-10-14
Part 2 of searchbar focus fix: for OSX

sauravg:
Don't hide "find..." menu item while searching

The only reason wxWidgets sends us Activate event on Mac on
hitting Cmd-F is that menu item. If it is removed, wxWidgets
won't generate Activate events on Mac anymore and Cmd-F will
do nothing until we close and re-open the search bar.

I have no idea why this isn't an issue on Linux as well.


ronys:
Could be that this is also an issue on Linux, see https://sourceforge.net/p/passwordsafe/bugs/1193/#9209

I'm merging this pull request after review, but will not have time to test before this evening. Perhaps yshirman can help with the testing?

PR #17, opened 2015-10-20
Hide/Unhide pwsafe during autotype using timers

sauravg:
This should fix http://sourceforge.net/p/passwordsafe/bugs/1178/. However, I don't know what broke it in the first place, even after debugging. The hiding/unhiding code has undergone a lot of changes by multiple authors and I even see use of mutexes in some places. However, I could find no difference between pwsafe's code flow and a simple test app. The test app worked (it could hide itself) but pwsafe didn't, with exactly the same set of function calls.

So this new code doesn't really fix what's broken (I don't even know  the what/when of it), but rather somehow ensures that the breakage cannot keep pwsafe from hiding itself.


ronys:
Lovely code!
- One could argue that TimedTaskChain really belongs in os, event though it's currently wx-specific.
- The unit test code should be moved under src/test in the merge to V4, and put under the gtest framework.

PR #18, opened 2015-10-20
Implement Find support for attachment filename

c-273:
(Need to fix that Find no longer saves user selection within the same session)


ronys:
Reviewed, tested & accepted - thanks!

Note that the attachment-specific fields in advanced and filters should only be shown when a V4 database is open...

PR #24, opened 2015-10-26
Replace toobar images on dialog button by normal text.

c-273:


ronys:
We haven't received any complaints about the clipboard graphic on the existing dialog boxes (which have been there for quite a while), and I don't see any reason to either revert to text or be inconsistent in the add-edit entry with the other d'boxes.

So for now I'm not merging in this pull request.

ronys:
Cherry-picked the second commit, leaving first out of master for now.

PR #29, opened 2015-10-29
Optimize HTML PNGs

kainjow:
Files were optimized via ImageOptim (http://imageoptim.com). Average save of 95% file size. This also fixes a warning dialog "iCCP: known incorrect sRGB profile" on OS X when some pages were viewed.


ronys:
Tested on Windows, no noticeable difference. Thanks!

PR #30, opened 2015-10-30
Fix compiler error (regression) and various warnings

kainjow:


ronys:
Merged into master - thanks!

PR #32, opened 2015-10-31
Fix couple assertions in sizer code.

vovkasm:
wxEXPAND conflicts with wxALIGN_* sizer flags. Use only wxEXPAND for layout sizers.


ronys:
Thanks - tested under Debian. clean build, no assertions.

PR #33, opened 2015-10-31
Support filter of attachment fields - WIP

c-273:
Compiles cleanly but currently doesn't filter them
TODO:
- add support of filter export/Import
- display of attachment filter properties
- implement proper ComboBox separators (filter fields)


ronys:
to be replaced.

PR #40, opened 2015-11-10
Fix autofocus on master password field in the startup dialog on Mac.

vovkasm:
wxWindow->SetFocus on Mac OS X works only if window already visible on screen.
So:
- move code for autoselect the master password field to EVT_ACTIVATE event handler for the dialog.
- guard the code to be called only once (in case user will change focus and switch between applications).


ronys:
Note to self: Needs regression testing on other wx platforms.

Thanks!

Issue #45, opened 2015-11-29
Linux system tray and -m / -s switch behavior


tknarr:
On Linux the -m switch with the system tray icon turned on causes the main window to minimize like a normal program and not show the system tray icon until it's unminimized and closed. The -s switch with a database named always causes the system tray icon to appear, but as often as not the main window remains visible with no database in it.

I'm going to take on tracking this oddness down. I've got a couple of ideas but need to confirm what happens at run-time.

PR #46, opened 2015-11-30
Fix window visibility on initial open, fix showing of system tray icon

tknarr:
Issue #45
When using the -s switch, don't show the main window. This fixes the problem of the main window sometimes remaining visible when the program starts.
If the systray is being used, show the tray icon after all other initialization is done. This insures the tray icon's visible from the start even if the main window is never shown, rather than only being shown after the main window is closed for the first time.


ronys:
Looks good - thanks!

PR #54, opened 2016-01-09
Changes in build/link options

pm-kan:


ronys:
I've cherry-picked the first commit and added the same flags to the vs12 project files.
The second commit (multiprocessor compilation is redundant, as we already support this via the UserVariables-1[24].props configuration file. You can run the configuration script or just add
  <ItemDefinitionGroup>
    <ClCompile>
      <MultiProcessorCompilation>true</MultiProcessorCompilation>
    </ClCompile>
  </ItemDefinitionGroup>
to the file manually.

(this is because some programmers preferred not to bog down their PC while compiling)

PR #64, opened 2016-01-19
Fix cli build

sauravg:
Specify c++11, link with core twice to satisfy libos, fix an apparent typo in "clean" target.


ronys:
Odd. I recall adding core lib twice recently. Perhaps in the CMake file?
Thanks!

PR #65, opened 2016-01-22
Fix errors in ::GetFileTimes for Mac OS X

jeffsf:
Parallel other use of 'statbuf' rather than 'info' in
::(fn, &statbuf)

As described in [Ticket #1321](https://sourceforge.net/p/passwordsafe/bugs/1321/), *src/os/mac/file.cpp* does not compile under Mac OS X 10.11.2 with Xcode 7.2
* *&statbuf* is not defined
* *status* is declared twice

This change allows this specific file to compile.

Not that I cannot yet verify the functionality as the project does not successfully build. *Generate version.h* does not appear to run, causing problems with *About.cpp* missing *version.h*


ronys:
Looks good - thanks!

PR #67, opened 2016-01-29
Fix cmake build on Linux without yubikey lib (NO_YUBI option)

vovkasm:
When I try to build project on Linux with cmake
```sh
mkdir build && cd build
cmake -GNinja -DNO_YUBI=YES ..
```
, it failed to compile `src/os/linux/PWYubi.c`. `src/os/linux/Makefile` has logic to exclude this file from build with yubikey support disabled.
So this patch implement same logic for cmake build.


ronys:
Thanks!

PR #68, opened 2016-02-02
Fix saving current (V3 or V4) DB after previously exporting to V1 or …

c-273:
…V2 format

Incorrectly calculating file signature causing issue creating intermediate backup


ronys:
Thanks!

PR #72, opened 2016-02-09
Simpler, faster and more accurate wchar => keysym map

sauravg:
1. No need to use a C++ map. A switch/case statement is good enough, and faster.
2. Ignores the deprecated mappings if they are superceded by anything else.
3. Includes a few new keysyms from the latest X11/keysymdefs.h


ronys:
Thanks!

PR #74, opened 2016-02-19
Prevent double click on CRichEditExtn doesn't include trailing white …

c-273:
…space

See SR 463 double click puts blank into the clipboard unintendedly


ronys:
Got it - Thanks!

PR #77, opened 2016-02-28
gtest-related makefile enhancements:

bblough:
* Automatically build gtest lib in-tree during build (instead of
    requiring it be manually built beforehand)
* Allow test build/run for release builds
* Add test dir to top-level clean targets


ronys:
Thanks!

PR #84, opened 2016-03-13
Fix export of Named Password Policies to XML and implement FR to view associated entries

c-273:
Whilst implementing FR 806, found & fixed bug re: export of named password policies to XML. Then decided I needed a way to easily see what entries used a partiular named password policy and so implemented it as a FR.
Note: only tested under VS2015. VS2013 files edited manually as I have now standardised on VS2015 and uninstalled VS2013.


ronys:
Got it - thanks!

Issue #85, opened 2016-03-15
CMakeLists.txt: mixes Windows and Unix stuff ? adding more info 


yopito:
Hello, quoting `CMakeLists.txt`:

```
 26 if (NOT WIN32)
 27   # TODO - add support for wx build under Windows
 28   set(wxWidgets_CONFIGURATION mswu)
```
`wxWidgets_CONFIGURATION` (and `mswu`) is related to Windows, not Unix/Linux ?

Suggested patch:
```
@@ -25,7 +25,14 @@
 
 if (NOT WIN32)
   # TODO - add support for wx build under Windows
-  set(wxWidgets_CONFIGURATION mswu)
+  # help people about wx-config
+  find_program(MYHINT_wxconfig wx-config)
+  if(NOT(MYHINT_wxconfig))
+    message(STATUS
+      "Failed to find 'wx-config' executable:\n"
+      "   Hint: add this on command-line: -D wxWidgets_CONFIG_EXECUTABLE=/path/to/wx-config"
+    )
+  endif()
   find_package(wxWidgets COMPONENTS adv base core html net REQUIRED)
   include(${wxWidgets_USE_FILE})
```
Checking `wxWidgets_CONFIG_EXECUTABLE` (via another macro name) will make life easier for unusual platforms (like MacOS X, see #55 )

ronys:
Can you submit a pull request for this?

yopito:
yes. You want one or 2 PR since 2 distincts problems ? or a PR with 2 commits to merge ?

yopito:
here it is: one PR ( #86 ) with 2 commits. I close this issue then.

PR #86, opened 2016-03-16
fix cmake/wx win32 in non-win32 section, hint on wx-config usage from command-line

yopito:
see issue #85 for more explanations.


ronys:
Looks good - thanks!

PR #93, opened 2016-04-02
Logic mismatch between gmtime64_r/gmtime_r and gmtime64_s

bblough:
gmtime64_s is expecting gmtime64_r to return true on success, false on failure.
gmtime64_r is returning ( gmtime_r() ==  0), but gmtime_r returns pointer to struct on success or NULL on error.  So it appears that the check in gmtime64_r is backwards.


ronys:
Funny - I caught this as well at about the same time - thanks!

PR #94, opened 2016-04-05
Patch from Ubuntu to fix link failure when built with -Wl,--as-needed

bblough:
Reorder libs so uuid comes after other libs that require it


ronys:
Hm,

Looks like whatever required the double inclusion of libcore.a is no longer relevant...

Thanks!

PR #96, opened 2016-04-22
Fix BR1334 & implement FR806

c-273:
FR806 take two - changing the action when deleting last entry with a group value.


ronys:
Looks good - thanks!

PR #97, opened 2016-04-26
Fixes for FreeBSD for 0.98.1

Sp1l:
1. Rename src/os/linux to src/os/unix
2. Refactor linux pathnames in Makefiles, incluide statements etc
3. Merge 0.98.1 changes into Makefile.FreeBSD
   - Remove Linux packaging targets
4. Don't overwrite CC/CXX when already set in Make env in wxWidgets Makefile
5. Use more fool-proof way of detecting HOST_NAME_MAX for env.cpp
6. Use methods available in FreeBSD to prevent trace and limit dumping core


ronys:
Looks good - thanks!

PR #103, opened 2016-05-21
Fix osx builds done using Xcode

sauravg:
Misc/mkversion.pl has changed and now expects some additional args.
Without those, it won't generate version.h. We pass some dummies
to keep it happy.


ronys:
Thanks!

PR #111, opened 2016-06-11
Update wxWidgets Windows project files for new files. Separate cmake and current wxWidgets version generation

c-273:
This doesn't cause cmake to affect the source tree as you require


ronys:
I'm OK with this for now, as the whole version marking needs to be rethunk.

Thanks!

PR #114, opened 2016-06-16
Fix osx & linux builds

sauravg:
Breakage caused by moving of RUEList.{h,cpp} to src/core, and the new mainView.cpp being missing from Makefile


ronys:
Thanks!

PR #125, opened 2016-07-07
Update ResText to latest provided by TortoiseSVN

c-273:


ronys:
Has this been tested against our code base? I remember making non-trivial changes a while ago since the processing time was way too long...

Issue #122, opened 2016-06-27
Fedora 24: some issues


sguazt:
Hello,
Just installed Fedora 24 x86_64.
I'm facing some issues with 0.98.1BETA.

- If I install the RPM provided by this site (pwsafe-0.98.1BETA.x86_64.rpm), I get the following message:

```
10:20:53 PM: Warning: Mismatch between the program and library build versions detected.
The library used 3.0 (wchar_t,compiler with C++ ABI 1010,wx containers,compatible with 2.8),
and your program used 3.0 (wchar_t,compiler with C++ ABI 1009,wx containers,compatible with 2.8).
Could not initialize help subsystem.
```

the program will start but:
   1. it doesn't remember the history (every time I run it, it starts with an empty history)
   2. it doesn't install an entry in the system menu
   3. the UI is weird (instead of a window with white background I see a window with a grey background); but maybe this is the new look&feel of WxGTK3...?

- If I try to create my own RPM (I've used the spec file in the install/rpm directory), install it and run, I get the following message:

```
/usr/include/wx-3.0/wx/strvararg.h(456): assert "(argtype & (wxFormatStringSpecifier<T>::value)) == argtype" failed in wxArgNormalizer(): format specifier doesn't match argument type
pwsafe:[14682]: PWSprefs - using existing config file: /home/sguazt//.pwsafe/pwsafe.cfg [R/W]
Segmentation fault (core dumped)
```

and the program won't start

Any idea?

Thank you very much!

Marco

sguazt:
The new release (v.0.99) fixes most of the above issues.
Still we have:
* no application menu entry (I use XFCE desktop manager)


A possible fix is to copy file:
```
/usr/share/pwsafe/pwsafe.desktop
```
to
```
/usr/share/applications/pwsafe.desktop
```
and change the `Icon` property from:
```
Icon=pwsafe
```
to
```
Icon=/usr/share/pwsafe/pwsafe.png
```

This works for me.

ronys:
Thanks. Care to create a pull request?

sguazt:
OK. I'll work on it ASAP.

Cheers

sguazt:
Just a question.
For creating the RPMs (the ones available in the GitHub "releases" page), are you using the `install/rpm/pwsafe.spec.tmpl`?
I'm asking this because this file seems OK (or at least seems to install the files in the right place):

```
install -d -m 755 ${RPM_BUILD_ROOT}/usr/share/applications
[...]

PR #126, opened 2016-07-10
Fix rpm spec

sguazt:
Hi,

Please, consider to merge my pull.

With this pull, I'm trying to fix these two issues (see #122):
1. The CPack problem that generates a bad RPM which conflicts with the `filesystem` package.
2. The location of `pwsafe.desktop` and `pwsafe.png`.

Regarding the last point, I'm not sure if my fix conflicts with other non-RPM-based environment.
Unfortunately, I didn't find an easy way to tell CPack to copy these files in the new locations only if the generator is RPM.
That is, a safer fix would be:

```
if(CPACK_GENERATOR MATCHES "RPM")
    install (FILES "install/desktop/pwsafe.desktop" DESTINATION "share/applications")
    install (FILES "install/graphics/pwsafe.png" DESTINATION "share/pixmaps")
else ()
    install (FILES "install/desktop/pwsafe.desktop" DESTINATION "share/pwsafe")
    install (FILES "install/graphics/pwsafe.png" DESTINATION "share/pwsafe")
endif(CPACK_GENERATOR MATCHES "RPM")
```

but putting this code in `CMakeLists.txt` doesn't work.

Best,

Marco

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -318,8 +318,8 @@ install (TARGETS pwsafe RUNTIME DESTINATION "bin")
 if (NOT WIN32)
    install (DIRECTORY xml DESTINATION "share/pwsafe/")
    install (FILES "docs/pwsafe.1" DESTINATION "share/man/man1") # gzip'ed in postinst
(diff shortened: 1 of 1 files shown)


ronys:
Looks good - thanks!

PR #131, opened 2016-07-20
Update to build via cmake on FreeBSD 10.3-RELEASE

derekschrock:
- Include /usr/local/include so we can find uuid/uuid.h
- Tested on 10.3
- Update requirements and instructions
- Share the same source files a Linux builds


ronys:
Looks good - thanks!

PR #134, opened 2016-07-22
Fix empty group export and included empty groups when exporting a non-empty group

c-273:


ronys:
Looks good, will test it over the weekend.

PR #138, opened 2016-07-25
Make the build reproducible

bblough:
Debian has a reproducible build project (https://wiki.debian.org/ReproducibleBuilds) that strives to make package builds byte-for-byte reproducible in order to allow for easier verification.
The attached patch fixes some issues in pwsafe that cause builds to differ even when built under the identical environment/compiler conditions.
 - strip extra fields from zip file containing help files
 - don't honour locale in rc2cpp.pl, or else keys written
   into generated cpp source file will be sorted differently
   under different locales.

### DIFF
--- a/Misc/rc2cpp.pl
+++ b/Misc/rc2cpp.pl
@@ -18,7 +18,6 @@
 
 use strict;
 use warnings;
-use locale;
 use File::Basename;
 
 sub usage {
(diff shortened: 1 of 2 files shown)


ronys:
Thanks!

I've added the help/CMakeLists.txt to the modified files (Added '-X' flag to zip there too), as we're slowly migrating the build process to cmake.

Merge gratefully accepted, in any case.

PR #141, opened 2016-07-29
More empty groups fixes

c-273:
Not sure how I managed to get the same changed committed with different IDs in the same branch.

I thought I had issued this request earlier.


ronys:
Looks good, despite duplicate commits (odd).

PR #142, opened 2016-07-29
Enhanced Save immediately

c-273:
See comments in DboxMain.cpp lines 1365-7 & 1392-4


ronys:
Did you consider marking the data changes - SetChanged(DATA) - in the core/commands Execute?
Seems to me that this would be easier to maintain and work for both wx and MFC?

ronys:
Functionally this seems to work fine.

Main issue I see is extraneous refreshes of tree when save-immediate is set compared to when it's not.

Perhaps this is due to the modified (italic) font being set and unset, which isn't really needed under save-immediate?

ronys:
Merged manually.

PR #149, opened 2016-08-09
Added QR code entry field in v3 and v4.

medo64:
Currently unimplemented in PasswordSafe, but defined for use in a
compatible program.


ronys:
Excellent - thanks!

PR #154, opened 2016-08-21
Fix rename of group (adding unnecessary rename of empty subgroups if …

c-273:
…none exist)


ronys:
VeryLongName doesn't bother me so much as logic in ui instead of in core, although certainly a challenge doing this with multi-command. Leaving as-is for now.

PR #155, opened 2016-08-21
Fix missing function arguments in XML import/export cli + some refactoring

dskrvk:


ronys:
Verified under Windows and 64-bit Linux.
Thanks!

PR #161, opened 2016-08-26
Change List view column picker D&D to be like Tree list

c-273:
Reported ASSERT in CmdTarget, although I could not reproduce it

ASSERT is in the same place as I had in TreeCtrl but I couldn't reproduce it - so changed anyway
to be like TreeCtrl


ronys:
OK, merging in and testing.

ronys:
Still hit assert, but not consistently, upon program exit after using column picker.

c-273:
I did what you did a number of times with never an assert. 

I can remove the change apart from CPWTreeCtrl where I did get the assert in exactly the same place CCmdTarget. 

Let me know. 

I could do this via a diff so you can test rather than a commit. For all I know VS2015 has changed something and it could happen before I did the change. Have you tried 3.39?

David 

PS No laptop time until this afternoon my time ~3pm

ronys:
Let's leave it for now. Will try on release build as well as checking if this is a regression from 3.39.1.

PR #164, opened 2016-08-27
List view changes and DB Properties group counts

c-273:
Feel free to ask me to revert the DB Properties group count change


src/ui/Windows/Properties.cpp
```diff
@@ -46,9 +48,18 @@ BOOL CProperties::OnInitDialog()
...
+  CString ngroups, cs_numGroups;
+#if 0
+  // Use groups based on the entries
   ngroups.Format(IDS_NUMGROUPS_E,
      m_pdbp->numgroups.c_str(), m_pdbp->numemptygroups.c_str());
+#endif
+  // Use groups based on the groups that are displayed in Tree view
+  // cf. Windows Explorer
+  std::vector<std::wstring> vGroups;
+  GetMainDlg()->GetAllGroups(vGroups);
```
ronys: Perhaps this should be the semantics of m_pdbp->numgroups? Is m_pdbp->numgroups used anywhere else that might be broken by changing this?
In any case, wx's property display code should be similarly changed, otherwise the two versions will show different counts for the same DB...

docs/ReleaseNotes.html
```diff
@@ -82,6 +82,26 @@ <h1 align="center">Password Safe 3.39.2 Release Notes ???? 2016</h1>
...
+positions or columns are added or removed.</td>
+</tr>
+<tr>
+<td align="center" width="100">[]</td>
+<td>The user specified Notes font is now also applied to the Notes column in the
+List view.</td>
+</tr>
+<tr>
+<td align="center" width="100">[]</td>
+<td>Entry Protected sttaus and AutoType string is now correclty displayed in the
```
ronys: Should be
"Entry Protected state and AutoType string are now correctly displayed in the"

docs/ReleaseNotes.txt
```diff
@@ -13,6 +13,14 @@ Bugs fixed in 3.39.2
 [] Correctly import empty groups from a XML file when an imported group has
 been specified.
 [] Empty groups are now added when merging one PasswordSafe database into another.
+[] List view column widths are no longer reset when dragging columns to different
+positions or columns are added or removed.
+[] The user specified Notes font is now also applied to the Notes column in the
+List view.
+[] Entry Protected sttaus and AutoType string is now correclty displayed in the
```
ronys: Should be
"Entry Protected state and AutoType string are now correctly displayed in the"

PR #165, opened 2016-08-28
Fix type mismatch issues; remove unused variables

dskrvk:


ronys:
Tested both on Linux 64 bit and Windows - good work. Thanks.

I wonder if it isn't better to use the 'auto' feature for local variables, e.g.,
auto n = strlen(foo);
Any thoughts?

dskrvk:
Yeah, we should use it more in the codebase. Will keep it in mind next time I'm making changes.

PR #167, opened 2016-08-31
Fix ASSERT during Synch & incorrect test

c-273:


ronys:
Will investigate as you suggested.

Thanks!

PR #171, opened 2016-09-05
Precursor to new attachment viewer (very soon)

c-273:


ronys:
Thanks. Would be great if you could commit the viewer code in another branch.
(you can use "git stash" and "git stash pop" to save the uncommitted work, create a new branch and then put your work there).

ronys:
Yes, that's understood.

PR #178, opened 2016-09-13
V4 enhancements

c-273:


ronys:
Keeping this on hold for the minute. Certainly not for 3.40.

ronys:
I'm merging this manually due to a couple of conflicts. Pushing it in a separate branch for now: V4AttMngmt

- The Manage Attachments dialog is a (very) useful development tool, but I'm not convinced that users should have this level of detail/control. Probably a simplified version.
- I think that CItemAtt::m_bOrphaned is in principle redundant, and should be true iff refcount is zero
- Would much rather not allow users explicit control on purging attachments.
-  ReuseAttachment is also redundant. If the attachment's identical in all parameters, it should be referenced, otherwise, should be imported separately. Can't think of a scenario where the user will decide one way or the other in a meaningful manner, we should decide for him/her.
- Buggette: Duplicating an entry with an attachment didn't bump refcount.

PR #179, opened 2016-09-13
Fix Find toolbar issues

c-273:
No idea why commits are there twice!


ronys:
Reviewed, tested and merged - thanks!
(rephrased relnotes, but you expected that, didn't you...)

PR #195, opened 2016-11-04
Try to solve flush of data to removable devices

c-273:
Not added flush to ReleaseNotes or updated discussion item
Unix/Mac need additional equivalent code (fsynch ?)
Also, minor documentation updates

### DIFF
--- a/help/default/html/autotype.html
+++ b/help/default/html/autotype.html
@@ -81,9 +81,10 @@ of the Notes field</li>
 <ul>
 <li><i>NNN</i> can be one to three digits (e.g., \o5 for 
 the 5<sup>th</sup> line, \o45 for the 45<sup>th</sup>, etc.</li>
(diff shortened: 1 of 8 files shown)


ronys:
Two thoughts:
- Why not apply the FlushFileBuffers() logic in any case? This would remove the need to detect whether or not media is removable.
- FClose() should return bool, indicating failure, not just ASSERT(0). Consider the user removing media - we want the write to fail and prompt to try again. As written, I think this will silently fail in Release build.

c-273:
Feel free to change. I can either rework these commits (cancel the MR) or you can change with a new commit. 

I would remove the ASSERTs as well personally. 

As to returning bool,  the original PWSfile:Close didn't - it always returned SUCCESS even if fclose failed! 

David ⁣​

ronys:
Not checking the return value of close/fclose is a historical error.
(Closing per your request)

Issue #196, opened 2016-11-05
Build failure on i386 FreeBSD


smeredith:
I followed the instructions in the FreeBSD readme and am getting a compile error:

``` shell
$ gmake-lite
[  0%] Building CXX object src/core/CMakeFiles/core.dir/Command.cpp.o
In file included from /usr/home/smeredith/pwsafe/src/core/Command.cpp:11:
In file included from /usr/home/smeredith/pwsafe/src/core/CommandInterface.h:10:
In file included from /usr/home/smeredith/pwsafe/src/core/coredefs.h:18:
/usr/home/smeredith/pwsafe/src/core/ItemData.h:230:8: error: class member cannot be redeclared
  bool Matches(time_t time1, time_t time2, int iObject,
       ^
/usr/home/smeredith/pwsafe/src/core/ItemData.h:228:8: note: previous declaration is here
  bool Matches(int num1, int num2, int iObject,
       ^
1 error generated.
gmake-lite[2]: *** [src/core/CMakeFiles/core.dir/build.make:144: src/core/CMakeFiles/core.dir/Command.cpp.o] Error 1
gmake-lite[1]: *** [CMakeFiles/Makefile2:179: src/core/CMakeFiles/core.dir/all] Error 2
gmake-lite: *** [Makefile:161: all] Error 2
```

smeredith:
As far as I can tell, no code is depending on overloading to resolve the name, so one of these could just be renamed. I will make a patch.

PR #197, opened 2016-11-05
Fix compile error on 32-bit FreeBSD

smeredith:
There are multiple functions named Matches() that differ only by
parameter type. In 32-bit BSD, the params for two of these are
both actually int. No callers are depending on the correct
overload to be bound based on type, so the easy fix is to rename one of
them, and change the call site to use the new name.

This fixes issue #196.

I built FreeBSD 32-bit and 64-bit and OS X and smoke-tested the app. I did not build Windows.


smeredith:
You are correct. There is only one caller. Fix it.

PR #200, opened 2016-11-05
Updates in Russian .po and Help

pm-kan:


ronys:
Much appreciate the effort. Note that some of the English texts will probably be changed slightly before release.

PR #206, opened 2016-11-08
Changes to build deb package "lintian-clean" using cmake + cpack

thompsonm-ANL:
- only run deb steps if we're on a deb based system
- specify specific depends for ubuntu, mint, debian
- get rid of old control files to alleviate confusion (let me know if you think these should stay)


ronys:
Looks good. Merging into master and testing it directly from there.

Thanks!

PR #208, opened 2016-11-09
Fix password history processing for aliases

c-273:


ronys:
Closed per submitter's request.

PR #211, opened 2016-11-14
Multiple fixes including a crash, aliases & PWH, empty groups

c-273:
Note commit 577614b now purely adds comments to mark contensious code after other changes made atomic in other commits.
I really think that BR1124 should not have been implemented as the wrong dependent was being used.  Similarly, when implemented, the Help file was not updated to indicate this change (both in 2013 c22bafa & 2015 475a594). The use of SafeGetBaseEntry(ci, ci) in DBoxTray.cpp (introduced in 2011 480a8f4 and unchanged since then) also violates BR1124.


ronys:
Looks good - merging for testing. Thanks!

PR #212, opened 2016-11-14
Fix missing notification after changing commands within commands by h…

c-273:
…elper functions.


ronys:
OK, but I agree that this is unrelated to the unit test failure, which I didn't have time to nail down (yet).

PR #214, opened 2016-11-15
Resolve various empty group/filter issues:

c-273:
1. All empty groups are no longer shown when the user filters for only entries with expired passwords.
2. Empty groups added but not yet saved are now included when the user filters on unsaved changes.
3. Empty groups, whose names satisfy a filter that include tests on group name values, are now included in the display when the filter is active.


ronys:
Looks good - thanks!

PR #215, opened 2016-11-16
Some more fixes

c-273:
I have made PWS consistent throughout wrt username and shortcuts. I have also backed out BR1124 but via an #ifdef/#else/#endif.  As it stands, BR1124 is undefined in the project.  Depending on your decision, you can either define it (not recommended), swap the code around and change #ifdef to #ifndef to re-instate BR1124 consistently or remove the #ifdef/#else#/endif and make the change you want permanent.  Search on BR1124 over the entire solution.


ronys:
OK, I see some refactoring I'd like to do, but the general approach seems good.

PR #217, opened 2016-11-18
Implement Recent Entries Run Command, change GetValues and its callers, fix out of bound error

c-273:


ronys:
Cherry-picked first 2 commits, refactoring GetValues as discussed.

PR #228, opened 2016-11-28
I18n

pm-kan:
Sorry, haven't time to do it in time(


ronys:
Better late than never - thanks!

PR #232, opened 2016-12-11
Rebase to master

c-273:
Hopefully correct but one test fails in V4 CoreRWTest - to be investigated


ronys:
Rebased in main project repo independently after discussion.

PR #238, opened 2016-12-29
Removed an unneeded Winsock include

os12:
- security software with socket headers makes me uneasy...
 - compiled with Visual Studio 2015


ronys:
Thanks!

PR #242, opened 2017-01-03
FR819, BR1392, menu tweak and fix empty group validation

c-273:


docs/ReleaseNotes.html
```diff
@@ -115,9 +115,14 @@ <h1 align="center">Password Safe 3.41.1 Release Notes ???????? 201?</h1>
 <h4><a name="V3.41.1Changes"></a>Changes to Existing Features in V3.41.1</h4>
 <table>
 <tbody>
-<tr>
-<td align="center" width="100"><a href="nojs.html" onclick="FR(816);return false;">816</a></td>
-<td>Set installed version in uninstall section of Registry for Windows version.</td>
+<tr>
+<td align="center" width="100"><a href="nojs.html" onclick="FR(819);return false;">819</a></td>
+<td>When an entry is duplicated directly or by being in a group being duplicated, its
```
ronys: Phrasing is confusing, as "its" is ambiguous. Suggest something like "a duplicated entry's creation time is set to the time of duplication. All other times are copied from the original."

docs/ReleaseNotes.txt
```diff
@@ -26,6 +26,8 @@ input focus when clicking on the icon to unlock the database.
 
 Changes to Existing Features in 3.41.1
 ======================================
+[819] When an entry is duplicated directly or by being in a group being duplicated,
```
ronys: Ditto

src/ui/Windows/MainEdit.cpp
```diff
@@ -466,6 +466,22 @@ void DboxMain::OnDuplicateGroup()
         // Make copy normal to begin with - if it has dependents then when those
         // are added then its entry type will automatically be changed
         ci2.SetNormal();
+
+        // Set creation time to now but keep all others
+        // Note: potential date/time inconsistencies that should not be "fixed"
+        // during open validation i.e. fields changed before the entry was created!
+        ci2.SetCTime();
```
ronys: Would be nice to have this section refactored into CItermData::SetDuplicateTimes(const CItemData &src);
for use in group and later in wx.

docs/ReleaseNotes.txt
```diff
@@ -16,6 +16,10 @@ The ReleaseNotes.html file has links to the relevant tickets/issues.
 Bugs fixed in 3.41.1
 ====================
 [] Fix issue when locking database when not using the System Tray.
+[1392] If the command line parameter '-g' is used to specify a configuration
+file by its relative path, PasswordSafe will now look for it in the same
+directory as the executable unless overridden by the PWS_PREFSDIR environmental
```
ronys: We use "environment variable" in the documentation. Both are probably acceptable, but we should be consistent.

src/ui/Windows/ThisMfcApp.cpp
```diff
@@ -1022,15 +1086,28 @@ BOOL ThisMfcApp::InitInstance()
 
     if (!sxUser.empty())
       SysInfo::GetInstance()->SetEffectiveUser(sxUser.c_str());
+  } else {
+    // Issues - get out
+    return FALSE;
   }
 
+  // If we set the languague for the command line, free it now
```
ronys: typo

PR #243, opened 2017-01-04
FR819, BR1392, menu tweak and fix empty group validation

c-273:
Updated/replaced as discussed


ronys:
Looks good - merging directly into github repo!

PR #256, opened 2017-01-15
[wx] few fixes for alias/shortcut entries

pm-kan:
Attempt to fix D'n'D from Dragbar for alias/shortcut items: tried to use the same checkers/getters as in MFC ver. (haven't found quick and clean way without adding PasswordSafeFrame::GetBaseEntry).
Also quick and dirty fix for Add/Edit dialog, to show "base reference" for Alias entries (without check for alias correctness and other alias/shortcuts improvements, that were recently added to MFC version).


ronys:
Nice - thanks!

PR #261, opened 2017-01-22
Prevent annoying GUI refreshes during multiple action commands

c-273:


ronys:
Looks good - need to check against Linux build too.

PR #273, opened 2017-02-04
Fix build error on macOS 10.12.3 with Xcode 8.2.1

dhocker:
This is a minor change to fix a build error on Xcode 8.2.1.


ronys:
Thanks!

PR #278, opened 2017-02-08
gh270b - VS2015 compiler warnings for DebugM/ReleaseM x64 build

dhocker:
This set of changes resolves all of the compiler warnings (known as of this instant in time) that were produced when building DebugM/ReleaseM x64 builds under VS2015. The changes were checked in both x86 and x64 builds.

A good number of the warnings came from the use of the size_t data type. In x86 size_t is a 32 bits, but in x64 it is 64 bits. Developers working in x64 builds need to be cognizant of this detail. It would be easy to create a situation where incompatibilities were introduced by using a size_t variable in a data structure and then writing the data structure to a file.

Developers working on VS2015 should probably build both x86 and x64 configurations to verify that they have not introduced any compiler warnings.


ronys:
Nice work - thanks!

PR #281, opened 2017-02-13
 	Rationalise & simplify VS solution and project configurations (MFC & vwxWidgets)

c-273:
Fix x64 coretest Debug build


ronys:
Needed some work to resolve conflicts with Dave's changes, but looks worth it!

PR #283, opened 2017-02-16
Installer 64 - build x86 and x64 versions of .exe and .msi installer

dhocker:
This pull request modifies the NSIS installer script so that it can produce either a 32-bit or 64-bit .exe installer. It also modifies the Wix installer script so it can produce a 32-bit or 64-bit .msi installer. Both scripts accept a parameter that defines the architecture to be built.

This pull request also includes two batch files (BuildNSISInstaller.bat and BuildWixInstaller.bat). These batch files can be used by developers to produce installers. They also serve as examples of how to use the NSIS and Wix tools against their respective scripts. As such, they can be used to adapt existing build processes.


ronys:
Thanks - I'll probably tweak these a bit to work with my release process, but the code looks solid.

PR #293, opened 2017-03-06
FR824 Autotype hotkey

c-273:


src/core/XMLprefs.cpp
```diff
@@ -108,11 +108,27 @@ bool CXMLprefs::Load()
   return true;
 }
 
-bool CXMLprefs::Store()
+void SortPreferences(pugi::xml_node parent)
```
ronys: Shouldn't this be SortPreferences(pugi::xml_node &parent) ?

Misc/Xerces 3.1.4/Notes.txt
```diff
@@ -0,0 +1,35 @@
+In order to build gtest 1.7.0 using VS2017, please extract the files in
```
ronys: Copy/paste error

PR #304, opened 2017-03-31
Add some Hungarian translations.

NAR:
Hello!

These are far from complete, but at least the GUI elements (except the synchronization and YubiKey dialogs) are translated.


ronys:
Thanks - added a couple of lines in code to support this, and it seems to work fine. Will be in the next release, hopefully in the next couple of days.

PR #308, opened 2017-04-23
My changes since 3.42.1

c-273:


src/ui/Windows/PasswordPolicyDlg.cpp
```diff
@@ -190,9 +197,13 @@ BOOL CPasswordPolicyDlg::OnInitDialog()
   // If started with Tall and won't fit - return to be called again with Wide
   if (m_bLongPPs && !GetMainDlg()->LongPPs(this)) {
     EndDialog(-1);
-    return FALSE;
+    return TRUE;
```
ronys: Why was this changed? Need to test with both tall and wide screens...

src/ui/Windows/PasswordPolicyDlg.cpp
```diff
@@ -903,22 +926,40 @@ void CPasswordPolicyDlg::OnGeneratePassword()
...
+  // Enable/Disable Copy to Clipboard
+  m_pCopyBtn->SetBitmap(m_bCopyPasswordEnabled ? m_CopyPswdBitmap : m_DisabledCopyPswdBitmap);
+
   UpdateData(FALSE);
 }
 
 void CPasswordPolicyDlg::OnCopyPassword()
 {
+  if (!m_bCopyPasswordEnabled)
+    return;
```
ronys: Can we get here at all if !m_bCopyPasswordEnabled ? Can this be an assert(0)?

src/ui/Windows/res/PasswordSafe3.rc2
```diff
@@ -963,6 +964,8 @@ BEGIN
   IDS_EDIT_NAMED_POLICY        "Change/View a Database Named Password Policy"
   IDS_TESTPOLICY               "Generate Password using selected Password Policy"
   IDS_CLICKTOCOPYGENPSWD       "Click to copy generated password to clipboard"
+  IDS_COPYPASSWORDHELP1        "Position 1 is the first character, 2 is the second etc. up to N for the last character of a password of length N. -1 is the last character, -2 the next to last etc. up to -N for the first password character. Positions must be separated by either spaces, commas or semi-colons to be recognised. There must also be a space, comma or semi-colon at the end of the list."
```
ronys: The last sentence is surprising. Does the user really have to enter "1;3;5;" instead of "1;3;5"?

src/ui/Windows/res/PasswordSafe3.rc2
```diff
@@ -963,6 +964,8 @@ BEGIN
...
+  IDS_COPYPASSWORDHELP2        "When a subset is visible and there is no warning message, the button will be enabled and can be clicked to copy the subset to the clipboard."
```
ronys: Isn't this explaining the obvious?

src/core/PWSprefs.h
```diff
@@ -147,7 +147,7 @@ class PWSprefs
     MainToolBarButtons, PasswordFont, TreeListSampleText, PswdSampleText,
     LastUsedKeyboard, VKeyboardFontName, VKSampleText, AltNotesEditor,
     LanguageFile, DefaultSymbols, NotesFont, NotesSampleText, AutotypeTaskDelays,
-    AddEditFont, AddEditSampleText,
+    AddEditFont, AddEditSampleText, AltNotesEditorCmdLineParms,
```
ronys: I'd much rather have the AltNotesEditor's semantics change s.t. it can contain the params as well as the command name, e.g.,
"Path to editor maybe with spaces .exe" "param1" "param2"

rather than adding a separate preference.

src/ui/Windows/MainMenu.cpp
```diff
@@ -974,10 +974,22 @@ void DboxMain::CustomiseMenu(CMenu *pPopupMenu, const UINT uiMenuID,
         case CItemData::ET_ALIAS:
         case CItemData::ET_SHORTCUT:
           // Allow going to/editing the appropriate base entry
-          pPopupMenu->AppendMenu(MF_ENABLED | MF_STRING,
-                                 ID_MENUITEM_GOTOBASEENTRY, tc_dummy); 
-          pPopupMenu->AppendMenu(MF_ENABLED | MF_STRING,
-                                 ID_MENUITEM_EDITBASEENTRY, tc_dummy);
+          if (m_bFilterActive) {
```
ronys: Wouldn't this be correct regardless of the value of m_bFilterActive?

src/ui/Windows/MainMenu.cpp
```diff
@@ -1431,6 +1443,17 @@ void DboxMain::OnContextMenu(CWnd * /* pWnd */, CPoint screen)
       case CItemData::ET_ALIAS:
       case CItemData::ET_SHORTCUT:
         pPopup->RemoveMenu(ID_MENUITEM_CREATESHORTCUT, MF_BYCOMMAND);
+        if (m_bFilterActive) {
```
ronys: Ditto.

help/default/html/preferences.html
```diff
@@ -769,6 +769,12 @@
 </tr>
 
 <tr>
+<td>AltNotesEditorCmdLineParms</td>
```
ronys: See review notes for commit implementing this.

src/ui/Windows/DboxMain.cpp
```diff
@@ -3115,14 +3118,24 @@ struct NoDuplicates{
 // Returns a list of entries as they appear in tree in DFS order
 void DboxMain::MakeOrderedItemList(OrderedItemList &OIL, HTREEITEM hItem)
 {
-  // Walk the Tree - either complete tree or only this group!
+  // Walk the Tree - either complete tree or only this group
```
ronys: How will this work if exporting a filtered subset of the entries under list view?

src/ui/Windows/MainEdit.cpp
```diff
@@ -2065,6 +2065,21 @@ void DboxMain::OnGotoBaseEntry()
     CItemData *pci = getSelectedItem();
     ASSERT(pci != NULL);
 
+    if (m_bFilterActive) {
```
ronys: I'd expect the code below to work regardless of the value of m_bFilterActive

src/ui/Windows/PasswordPolicyDlg.cpp
c-273: This commit corrects and documents the return code from OnInitDialog throughout PWS. The return code from OnInitDialog is to indicate where focus is to be placed (TRUE - default, FALSE - it has been changed) - it does not indicate whether the dialog is shown. In this case, the focus wasn't changed and so the correct return code is TRUE.  In reality, it doesn't matter in this case as the call to EndDialog prevents the dialog ever being shown.

src/ui/Windows/res/PasswordSafe3.rc2
c-273: I am only documenting the current process. But yes - how else does the program know that, in your example, the user wants the 5th character rather than the 50th?  OK - unlikely for this case but what if the user typed just "2;3;1" for a 12 character password? Is the '1' the first or maybe the 10th, 11th or 12th? (Some of my online accounts do not have the characters in ascending order and so "2 7 1" is possible as would be "6 2 10".

src/ui/Windows/res/PasswordSafe3.rc2
c-273: Yes for us but not for all.

src/core/PWSprefs.h
c-273: I suspect that it would not be applicable to all editors. Where would the file name go - before "param1" or after "param2".  Would that be true for all external editors? This semantic mirrors the same additional parameters that we already provide for browser additional parameters.

src/ui/Windows/DboxMain.cpp
c-273: You can only export a group if you can select a group to export. Since groups are not displayed entities in the List view, the question is meaningless.

src/ui/Windows/PasswordPolicyDlg.cpp
c-273: m_bCopyPasswordEnabled is enabled if the password length > 0 i.e. one has been generated. This test prevents copying an empty string and the incorrect update of the Status bar stating a password has been copied.

c-273:
Ditto J

src/ui/Windows/MainEdit.cpp
c-273: As the comments state, if a filter is active so that the base entry is not in the current Tree/List view, then even though we may have removed menu the option to go to the base entry, if the user has defined an accelerator key to do this, MFC will still try to do it. This defensive code prevents failures do to the entry not being in the displays.

src/ui/Windows/MainMenu.cpp
c-273: If a filter is active, the base entry may not be in the current Tree/List view and so can't be displayed.

src/ui/Windows/MainMenu.cpp
c-273: If a filter is active, the base entry may not be in the current Tree/List view and so can't be displayed.

c-273:
To be done another way

PR #319, opened 2017-04-29
Implement exporting of the current results of a filter to a DB. Fix plain text import

c-273:


ronys:
Works fine - nice!

What do you think about reporting (that is, writing out to the generated report) when a base entry is added to the exported set? I think this might be of interest.

PR #320, opened 2017-04-29
Implement additional command line parameters to the external editor u…

c-273:
…sed to edit the Notes field from the Add/Edit Basic dialog.


src/ui/Windows/AddEdit_Basic.cpp
```diff
@@ -1363,7 +1365,9 @@ UINT CAddEdit_Basic::ExternalEditorThread(LPVOID me) // static method!
   CString cs_CommandLine;
 
   // Make the command line = "<program>" "file"
```
ronys: Comment needs to be updated

src/ui/Windows/AddEdit_Basic.cpp
```diff
@@ -1363,7 +1365,9 @@ UINT CAddEdit_Basic::ExternalEditorThread(LPVOID me) // static method!
...
-  cs_CommandLine.Format(L"\"%s\" \"%s\"", sxEditor.c_str(), self->m_szTempName);
+  cs_CommandLine.Format(L"\"%s\" \"%s\" %s", sxEditor.c_str(), self->m_szTempName,
```
ronys: 1. I think most editors prefer the filename to be last, e.g. editor.exe /flag file.txt
2. This will very likely break if more than one command line flag is given. Not sure how you want to handle quotes in the params string, but if the value is "/flag1 /flag2", then this should be passed as \"/flag1\" \/flag2\", and not as \"flag1 /flags\"

PR #321, opened 2017-04-29
Prevent user using accelerator of a deleted menu item

c-273:


src/ui/Windows/DboxMain.cpp
```diff
@@ -2688,14 +2689,343 @@ LRESULT DboxMain::WindowProc(UINT message, WPARAM wParam, LPARAM lParam)
...
+  if (message == WM_COMMAND &&  wNCode == 1 && wID < ID_LANGUAGES) {
+    if (CheckCommand(wID))
+      return 1L;
+  }
+
 exit:
   return CDialog::WindowProc(message, wParam, lParam);
 }
 
+bool DboxMain::CheckCommand(const WORD wID)
```
ronys: Is there a way to check the menu item's existence/state directly from the Menu rather than 300+ lines of code that need to be maintained in addition to CustomiseMenu? - This duplication worries me a lot, as it's very hard to remember to fix this as well.

The name CheckCommand() is a bit too generic - please rename to CheckAcceleratorValid() s.t. the context will be clearer.

c-273:
Will do post other merges

PR #323, opened 2017-04-29
Implement FR830 Find & Replace

c-273:


c-273:
I don't know what went on with these 3422-nn branches but it seems a mess. I had to send you the bitmap and when I switch to this branch, half the files are missing. I am closing this and re-issuing against master that should make it much much easier!

ronys:
Closed per discussion.

PR #325, opened 2017-04-29
Implement FR832

c-273:


ronys:
1. I like the System Tray ownership change - good refactoring!
2. On my system, the index isn't visible on the system tray icon - only on the tooltip.
3. Also, icon remains red even when db is locked (regression)
4. Finally, instead of an entry in the Manage menu that's enabled/disabled, I think it would be much more natural to have this in the systray popup.

PR #331, opened 2017-05-21
Implement FR830 - Find & Replace

c-273:


c-273:
Issue with repro - closing MR and re-doing it

PR #332, opened 2017-05-23
Implement FR830 - Find & Replace

c-273:


src/ui/Windows/FRHdrCtrl.cpp
```diff
@@ -0,0 +1,165 @@
+/*
+* Copyright (c) 2003-2017 Rony Shapiro <ronys@pwsafe.org>.
+* All rights reserved. Use of the code is allowed under the
+* Artistic License 2.0 terms, as specified in the LICENSE file
+* distributed with this code, or available from
+* http://www.opensource.org/licenses/artistic-license-2.0.php
+*/
+// GEHdrCtrl.cpp : implementation file
```
dskrvk: Copy-paste error? Wrong file name

ronys: Indeed.

src/ui/Windows/FindReplaceDlg.cpp
```diff
@@ -0,0 +1,786 @@
...
+  // Now do search
+  DboxMain *pDbx = (DboxMain *)GetParent();
+  size_t num_found = pDbx->DoFindReplaceSearch(m_ft, m_rule, m_secOldText, m_bCaseSensitive,
+                                               m_vFRResults);
+
+  // Allow buttons again
+  m_btnSearch.EnableWindow(TRUE);
+  UpdateButtons(TRUE);
+  
+  if (num_found == 0) {
```
dskrvk: Maybe otherwise display the number of replacements made?

src/ui/Windows/FindReplaceDlg.cpp
```diff
@@ -0,0 +1,786 @@
...
+    break;
+  case 5:
+  {
+    StringX sxOldText = self->m_secOldText;
+    StringX sxNewText = self->m_secNewText;
+
+    StringX sxLHSOldFieldValue = pLHS_pci->GetFieldValue(self->m_ft);
+    StringX sxLHSNewFieldValue = self->ChangeField(sxLHSOldFieldValue, sxOldText, sxNewText);
+
+    StringX sxRHSOldFieldValue = pLHS_pci->GetFieldValue(self->m_ft);
```
dskrvk: Shouldn't this be `pRHS_pci->GetFieldValue`?

src/ui/Windows/FRListCtrl.cpp
```diff
@@ -0,0 +1,337 @@
+/*
+* Copyright (c) 2003-2017 Rony Shapiro <ronys@pwsafe.org>.
+* All rights reserved. Use of the code is allowed under the
+* Artistic License 2.0 terms, as specified in the LICENSE file
+* distributed with this code, or available from
+* http://www.opensource.org/licenses/artistic-license-2.0.php
+*/
+// GEListCtrl.cpp : implementation file
```
dskrvk: Another name mismatch

src/ui/Windows/FRHdrCtrl.cpp
dskrvk: Another name mismatch

dskrvk:
Question - why be so specific with the matching rules? Most of the search dialogs nowadays only ask if you want case sensitive and/or whole word match. What is the use case for only wanting to match a beginning/end of a string?

ronys:
I agree with the last comment. In general, I'm unhappy with the interface for this feature, which is the main reason I haven't pulled it in yet.

PR #333, opened 2017-06-04
pwsafe-cli: enhance the pwsafe command-line utility with some more operations

sauravg:
Doc at https://gist.github.com/sauravg/bf8a53b0da747dd194f0


ronys:
Can't get this to compile under Linux using g++ (GCC) 4.8.5 (see attached). Please advise.
[make.txt](https://github.com/pwsafe/pwsafe/files/1050316/make.txt)

sauravg:
Oops. Whole bunch of mistakes like not build pwsafe itself, not doing “make clean”, building only on OSX but not on Linux, etc.

Should be fixed now but I’m not sure if d843d2c is actually a part of the PR itself. Please check.

PR #337, opened 2017-06-25
Test Makefile Fixes

bblough:
The autotools-based makefile for the test driver is missing LDFLAGS for the X11 and Xtst libs, causing linking to fail on Debian.  In addition, it does not honor LDFLAGS passed in from the build process.


ronys:
Thanks!

The cli build has managed to do without the X11 libraries, it would be nice to apply the same tweak for the test build.,,

bblough:
Noted.  I'll add it to my TODO list.

PR #340, opened 2017-08-02
Add a feature to display a password as QR code

sauravg:
Add an item to the popup menu brought up by right-clicking an item to show the entry's password as a QR code. For security reasons, the dialog closes automatically after 15 seconds.

display-password-as-qrcode

qrcode-dlg

This is useful for quickly transferring a password to a mobile device securely using a [QR-code scanning app](https://play.google.com/store/apps/details?id=com.google.zxing.client.android&hl=en)

Cons: adds yet another build dependency: libqrencode

Currently its linux-only. Stil researching on osx.

### DIFF
--- a/CodeLite/pwsafe.project
+++ b/CodeLite/pwsafe.project
@@ -131,6 +131,8 @@
     <File Name="../src/ui/wxWidgets/mainView.cpp"/>
     <File Name="../src/ui/wxWidgets/passwordsubset.cpp"/>
     <File Name="../src/ui/wxWidgets/passwordsubset.h"/>
+    <File Name="../src/ui/wxWidgets/pwsqrcodedlg.cpp"/>
+    <File Name="../src/ui/wxWidgets/pwsqrcodedlg.h"/>
   </VirtualDirectory>
   <Dependencies Name="Release">
     <Project Name="core"/>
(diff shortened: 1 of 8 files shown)


dskrvk:
Could you elaborate on the proposed use case? If this is to get a password into a phone, what happens next? Why not use an app such as PasswordSafe Sync?

sauravg:
> If this is to get a password into a phone, what happens next?

After that, you copy/paste the password from the qr code scanning app to
wherever you need it.

> Why not use an app such as PasswordSafe Sync?

That is up to the user. In sme cases, one may not have password sync set up
at that moment. One may not want to sync his passwd db to cloud at all. One
may not want that passwd db on his phone at all. Etc. Etc.

Basically, one may need to copy just one password from his laptop/desktop
to another device. Securely and conveniently. That's exactly what this
feature provides.

ronys:
Interesting idea - could you wrap it in `#ifndef NOQRENCODE...#endif` as we do for Yubikey support, s.t. I can make it an optional component?

ronys:
Added NO_QR conditional build, analogous to NO_YUBI, plus cmake support for Debian, including packaging.

Pushed to separate "qrencode" branch on main repo to work on support for other distros.

PR #343, opened 2017-08-15
Some fixes for wx version

pm-kan:
- Updates in Code::Blocks project
- Updated SlackBuild, so it can extract version info from version.wx and find .xsd after installation
- c274af3 make all wx includes as system (<>), and remove excessive path specification for os and core includes
-  0817745 - minor cleanup: wxT/_T changed to L"", also changed few strings to "wide"
(c274af3 and 0817745 aren't necessary: I've started to do it when I got some strange compilation warnings, but real problem was in incorrect local wx-config settings)
- fixed displaying of default DCA values in Add/Edit dialog
- also minor fixes to allow using tray icon in slackware and open pwsafe.org site from "About" dialog ("check version" doesn't work because wx don't support https)


ronys:
Merged manually without c274af3 and 0817745.

PR #344, opened 2017-08-16
Dragbar fixes for wxWidgets

hdfssk:
This fixes some cosmetic problems with the dragbar.

- The dragbar’s background is now repainted.
Currently, just after unlocking a safe the dragbar appears as expected:
bg initial paint
After minimizing & restoring passwordsafe, whatever was behind the dragbar is the new background:
bg failed repaint

- The dragbar now resizes with the passwordsafe frame.


ronys:
Thanks!

PR #348, opened 2017-09-30
README.LINUX updated for debian users

libBletchley:
.deb files are no longer in the release directory, so README.LINUX is updated to refer users to the developer build process.


ronys:
Thanks - I've taken a slightly different approach, removing references to specific releases.

PR #349, opened 2017-09-30
update README.LINUX.DEVELOPERS.txt for debian

libBletchley:
The original instructions simply point developers to a page that neglects to instruct users to obtain the relevant public key for the repository, and also there is a tendency for users to follow that page entirely and install too many packages.  Detailed instructions are now given.  (although perhaps my use of `aptitude` should be replaced with `apt-get` for the general public.. not sure)

### DIFF
--- a/README.LINUX.DEVELOPERS.txt
+++ b/README.LINUX.DEVELOPERS.txt
@@ -28,7 +28,7 @@ git
 libgtest-dev
 libqrencode-dev
 libuuid1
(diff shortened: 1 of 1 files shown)


ronys:
I'd rather not go into the specifics of the wxWidgets package installation. Actually, these instructions were written when 3.0 was relatively new. In the meanwhile, most distros have included 3.0, so this section is less relevant.
Rephrased accordingly.

PR #350, opened 2017-09-30
setup-deb-dev-env.sh modded to add repo if needed

libBletchley:


ronys:
I'd rather not automate the addition of keys and new package repositories for wxWidgets, as this is probably no longer needed. (See my comment on pull request 349)

PR #354, opened 2017-10-01
Few fixes for wx build 

pm-kan:


ronys:
Putting PWSUtil::loadFile() under core and creating a dependency from os to core is odd, but I see why you did it that way - the lesser evil, I guess...

PR #361, opened 2017-10-17
[wx] Implement Password History Options

bblough:
Implements the functionality required by the Password History tab on the
options window.

### DIFF
--- a/src/ui/wxWidgets/optionspropsheet.cpp
+++ b/src/ui/wxWidgets/optionspropsheet.cpp
@@ -78,6 +78,7 @@ EVT_BUTTON( wxID_OK, COptions::OnOk )
   EVT_RADIOBUTTON( ID_PWHISTSTOP, COptions::OnPWHistRB )
   EVT_RADIOBUTTON( ID_PWHISTSTART, COptions::OnPWHistRB )
   EVT_RADIOBUTTON( ID_PWHISTSETMAX, COptions::OnPWHistRB )
+  EVT_RADIOBUTTON( ID_PWHISTCLEAR, COptions::OnPWHistRB )
   EVT_BUTTON( ID_PWHISTNOCHANGE, COptions::OnPWHistApply )
   EVT_CHECKBOX( ID_CHECKBOX29, COptions::OnLockOnIdleClick )
   EVT_CHECKBOX( ID_CHECKBOX30, COptions::OnUseSystrayClick )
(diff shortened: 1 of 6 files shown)


bblough:
This is functional based on my testing, though there are a few minor outstanding items that I still need to address.  But there's one thing in particular that I would like some direction on, if possible.

On the Windows version, the "Manage password history of current entries" selections are applied by clicking the main OK on the options dialog.  However,  on the WX version, there is a separate Apply button specific to those controls.  

Should I remove the Apply button to make it function more like the Windows version, or should I leave it  there?  My instinct says to remove it, but the fact that it's there means that someone put it there intentionally (and I don't want to step on anyone's toes).

bblough:
I should also add that as it currently stands, clicking the apply button doesn't appear to do anything, as the logic doesn't actually happen until the dialog exits.  So if the apply button stays, I need to move the logic into the options dialog, or if it doesn't, I need to move the apply functionality into the apply for the entire options page.

I just made the last couple of outstanding changes save for the stuff related to the apply button, so once I make the apply-related adjustments, I think it will be good-to-go.

src/ui/wxWidgets/optionspropsheet.cpp
```diff
@@ -1103,9 +1123,32 @@ void COptions::OnPWHistSaveClick( wxCommandEvent& /* evt */ )
...
+    // Reset entries to HISTORY OFF
+    m_pwhistaction = 2 * applytoprotected;
+
+  } else if (m_pwhistsetmaxRB->GetValue()) {
+    // Don't reset history setting, but set history number
+    m_pwhistaction = 3 * applytoprotected;
+
+  } else if (m_pwhistclearRB->GetValue()) {
+    // Don't reset history setting, but clear all history items
+    m_pwhistaction = 4 * applytoprotected;
```
dskrvk: Are these coefficient values (1,2,3,4) actually flags? Are they declared somewhere? If not, it would be nice to make them named constants.

bblough: Not that I've seen.  I based this implementation on the Windows version, and this is how it handles it.  But I can go back and do that to both versions - that sounds like a good idea.

ronys: - The 'apply' button is there as an attempt to keep close to the Windows implementation, good to keep it.
- There are a few places where numerical values are used instead of well-named constants. I wouldn't go through the entire code base looking to clean these up, but if you have your editor open on the relevant section, I'd appreciate the extra effort. Bonus points for doing this cross-platform :-)

bblough:
OK, I think that should do it, unless anyone sees additional changes/improvements/issues.

PR #364, opened 2017-10-19
Help: update security+settings 

grinapo:
with information about the specifics of the key stretching. If I see right PBKDF2 isn't yet used, if it is, the PR shall be slightly modified (unless the iteration numbers would differ).


grinapo:
As a sidenote I wanted to append the Aliases help with "this doesn't work under Linux at all" but I am not sure whether this is intentional, or considered to be a bug. Please advise.

ronys:
Thanks,

PBKDF2 is used in the V4 format, which is still not ready for prime time.
V3 uses the algorithm listed, which is the "moral equivalent" (and predates PBKDF2, which wasn't available when the program was first written).

Alias support will eventually make it into the wx version, long with other missing features...

Issue #367, opened 2017-10-22
database corruption w/total loss of all passwords after successful import and rewrite


ghost:
For the first time, I went to view/edit a record and used the `Generate` button to replace the current password with a password that is generated by the tool.  I recall that the generated password started with a `+` and contained a dot.  I clicked OK, then alt-F4 to terminate.  I was then asked if I wanted to save the file, and chose 'yes'.  The next time I ran pwsafe, it prompted for the master password as usual, but the window simply terminated.  Terminal shows:
```
pwsafe
pwsafe:[9165]: PWSprefs - using existing config file: /home/user//.pwsafe/pwsafe.cfg [R/W]
pwsafe:[9165]: Couldn't load default language catalog for Chinese
pwsafe:[9165]: Couldn't load pwsafe language catalog for Chinese
pwsafe:[9165]: Couldn't load default language catalog for Danish
pwsafe:[9165]: Couldn't load pwsafe language catalog for Danish
pwsafe:[9165]: Couldn't load default language catalog for Dutch
pwsafe:[9165]: Couldn't load pwsafe language catalog for Dutch
pwsafe:[9165]: Couldn't load default language catalog for French
pwsafe:[9165]: Couldn't load pwsafe language catalog for French
pwsafe:[9165]: Couldn't load default language catalog for German
pwsafe:[9165]: Couldn't load pwsafe language catalog for German
pwsafe:[9165]: Couldn't load default language catalog for Italian
pwsafe:[9165]: Couldn't load pwsafe language catalog for Italian
pwsafe:[9165]: Couldn't load default language catalog for Korean
pwsafe:[9165]: Couldn't load pwsafe language catalog for Korean
pwsafe:[9165]: Couldn't load default language catalog for Polish
pwsafe:[9165]: Couldn't load pwsafe language catalog for Polish
pwsafe:[9165]: Couldn't load default language catalog for Russian
pwsafe:[9165]: Couldn't load pwsafe language catalog for Russian
pwsafe:[9165]: Couldn't load default language catalog for Spanish
pwsafe:[9165]: Couldn't load pwsafe language catalog for Spanish
pwsafe:[9165]: Couldn't load default language catalog for Swedish
pwsafe:[9165]: Couldn't load pwsafe language catalog for Swedish
pwsafe:[9165]: Selected language: menu id= 10245
*** Error in `pwsafe': free(): invalid next size (fast): 0x0000561068b7f260 ***
======= Backtrace: =========
[...]

PR #369, opened 2017-10-23
Fix mbstowcs-related heap overflow

bblough:
Add additional call to mbstowcs to check needed size prior to allocating
buffer.  Related to issue #367 

Note: mbstowcs calls OS specific code.  I have NOT tested this on Windows, but a review of the MSDN docs on MultiByteToWideChar leads me to believe it's OK.  Someone with a Windows machine should verify.

### DIFF
--- a/src/core/PWSfileV1V2.cpp
+++ b/src/core/PWSfileV1V2.cpp
@@ -377,16 +377,21 @@ size_t PWSfileV1V2::ReadCBC(unsigned char &type, StringX &data)
                     m_fish, m_IV, m_terminal);
 
   if (buffer_len > 0) {
-    wchar_t *wc = new wchar_t[buffer_len+1];
 
-    size_t wcLen = pws_os::mbstowcs(wc, buffer_len + 1,
+    //edge cases can make wc bigger than buffer, so calculate the size needed
+    size_t wcsize = pws_os::mbstowcs(NULL, 0,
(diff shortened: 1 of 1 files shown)


ronys:
Thank. Will check on Windows before merging. Does this resolve #367?

bblough:
It resolves the crash due to the invalid next() on free().  So it for sure solves part of it.   I'm not sure of there are other issues related to saving changes to V2 databases (or if that's even something that's supported).

ronys:
Surprisingly, this required a bit of work to function under Windows.

bblough:
Sorry about that - that's my mistake.  

Apparently I was focusing on

    cchWideChar [in]
    If this value is 0, the function returns the required buffer size, in characters, including any terminating null character, and makes no use of the lpWideCharStr buffer.

and glossed over

    Note that, if cbMultiByte is 0, the function fails.


I should have passed the source length into pws_os::mbstowcs, rather than 0 (keeping dstlen at 0).  Had I done that, then I don't think the extra logic added in 634a42522d416a85748 would be necessary.

I've added commit 0dc7e6a8baf388c3bc827ebdcd44ea000b7cc11b to add the srclen.  After merging it, you could probably revert the other logic in the windows impl, but of course, it's up to you.

ronys:
I suspected as much, but didn't want to mess up with core t save having to test the Linux build, plus the fix I made to the windows version was definitely benign.

PR #370, opened 2017-10-25
Fix UpdatePasswordCommand test to use correct units

bblough:
According to the V3 spec and the comments in CItemData::UpdatePassword,
the password expiration interval should be specified in days.  However
the test for UpdatePasswordCommand was setting it in seconds since epoch.
This lead to sign flipping on 32-bit systems which resulted in test failure.


ronys:
Nice catch - thanks!

PR #373, opened 2017-11-01
Mac qrencode

sauravg:
Implementation of "Show password as QR code" on OSX


sauravg:
The Mac implementation is in pwsqrencode-mac.mm (not .cpp) because it uses a mix of C++ and ObjC code, which can only be done in a .mm file.

So at least for Mac, there has to be a separate implementation file. I don’t see a way around that.

ronys:
OK, so we can have pwsqrencode.cpp and pwsqrencode.mm, with comments in both explaining the above. No need for -{mac,linux}.

sauravg:
Renamed the files

ronys:
I'm assuming you want this merged, which I can't do when the request is closed.

PR #377, opened 2017-11-19
Unbreak build on FreeBSD

Sp1l:
The function wxLinuxDistributionInfo is only available on Linux. Add check for LINUX symbol as per [documentation](http://docs.wxwidgets.org/3.0/group__group__funcmacro__networkuseros.html#ga06f6fb212c396bd20865ee4e2f69aa1f)
`This function is Linux-specific and is only available when the LINUX symbol is defined. `


ronys:
Thanks!

PR #382, opened 2017-12-16
cmake build fix for IPO support

SeeK-ntr0py:
Small re-order.  Build target referenced before being defined when IPO support is available. 
Necessary fix to compile under [Arch] Linux.


ronys:
Thanks!

PR #384, opened 2017-12-18
Make wxDragbar not expandable in vertical direction.

rafaelx:
The dragbar took with its single row of icons half of the space of the main frame.

wxdragbar_wofix

This change keeps the dragbar height at its necessary size.

wxdragbar_fix


ronys:
On what distro has this been tested? I vaguely recall this or a similar fix in the past that broke the UI on Fedora, Ubuntu or Debian.

rafaelx:
It's been tested on Kubuntu. Here is what lsb_release provides.

Distributor ID: Ubuntu                                                                                                                  
Description:    Ubuntu 17.04                                                                                                            
Release:        17.04
Codename:       zesty

The wx version as provided by wx-config.
3.0.2.0

PR #388, opened 2017-12-19
Make text entry fields of properties sheet expandable

rafaelx:
This change makes the text entry fields of the properties sheet expandable so that they get more available space. This makes more content of the text fields visible.

Layout without this change:
addeditpropsheet_wofix

Layout after applied change:
addeditpropsheet_wfix


ronys:
Looks good on Debian as well.

Thanks!

Issue #148, opened 2016-08-07
Dialogue box for new password entry not correctly sized


peter-hb:
When creating a new entry the dialogue box for the details is sized so as to hide the "Generate" option.
![pwsafe 1](https://cloud.githubusercontent.com/assets/20797719/17463483/68cffa2e-5cbe-11e6-9293-26d1b84afa6f.png)
![pwsafe 2](https://cloud.githubusercontent.com/assets/20797719/17463486/7d1cd45c-5cbe-11e6-8d50-95f3396d4b0c.jpeg)

peter-hb:
Additionally to the above, the Username field is not populated with the default username.

rafaelx:
Regarding the first issue using version 1.03BETA on Kubuntu 17.04 with wxGtk 3.0.2 the "Generate" button is not hidden.
![pwsafe_103beta_2](https://user-images.githubusercontent.com/4042917/34272625-3d6f886e-e691-11e7-97ce-e1c9d4d916c5.png)

I can confirm the second issue, that the configured default name doesn't appear in the corresponding text field.

rafaelx:
This issue was resolved by the pull request #388, #390 and #393.

PR #393, opened 2017-12-24
Show default user name from preferences

rafaelx:
If a default username is set in preferences it gets
shown in wx based dialog for adding a new user.

This addresses the second part of issue #148.


ronys:
Thanks again!

PR #394, opened 2017-12-26
Improve layout of wx based SafeCombinationEntry dialog

rafaelx:
Combobox for file selection is now expandable to give users the opportunity
to enlarge the dialog to make the full path and selected file visible.
Controls are more align to each other.


ronys:
Much better - Thanks!

PR #396, opened 2017-12-28
Fix help system of wx based variant of PWSafe

rafaelx:
Calling the Help from a modal dialog broke the applications event handling.


ronys:
Thanks - still working on a Windows issue, so haven't tested on my Linuxen yet.

PR #399, opened 2018-01-06
Enable translation for comboboxes and improve layout of options dialog

rafaelx:
The translation enabling of string literals that are part of an array is based on the wx API documentation (see [here](http://docs.wxwidgets.org/3.0/group__group__funcmacro__string.html#gab8862de0f74049708bfea75ae9e14c45) and [here](http://docs.wxwidgets.org/3.0/group__group__funcmacro__string.html#ga8a02b8875a521df57263a9e6f090f2d0)).

Textfields, that are intended to show a user selected path, are now stretchable, hence got more width to show more of the path.


rafaelx:
It might be necessary to not only consider the macro _("...") for strings that should be added into the translation catalog, but with this pull request also wxTRANSLATE("...").

PR #401, opened 2018-01-14
Fix #398

### LINKED ISSUE #398: [Linux][wxGtk] Assertion after double click of header separator in list view
listview_tableheader

listview_assertion

This issue is related to the wxGtk framework. See [ticket](http://trac.wxwidgets.org/ticket/13685) for progress and current status.

rafaelx:


src/ui/wxWidgets/PWSgrid.cpp
```diff
@@ -78,6 +79,18 @@ PWSGrid::PWSGrid(wxWindow* parent, PWScore &core,
 {
   Init();
   Create(parent, id, pos, size, style);
+  
+  // Handler for double click events on column header separator
+  auto *header = wxGrid::GetGridColHeader();
+  
+  if (header) {
+    header->Bind(
```
dskrvk: Is it necessary to unbind the handler (e.g. in the dtor) to free up any system resources?

src/ui/wxWidgets/PWSgrid.cpp
```diff
@@ -78,6 +79,18 @@ PWSGrid::PWSGrid(wxWindow* parent, PWScore &core,
...
+      wxEVT_HEADER_SEPARATOR_DCLICK, 
+      [=](wxHeaderCtrlEvent& event) {
```
dskrvk: Looks like you're not using any variables from the outer scope, so don't need to capture anything (`[]`).

src/ui/wxWidgets/PWSgrid.cpp
rafaelx: To be honest, I'm not sure. If it is necessary, than how can this be done in case of lambda's?
In case of doubt I can choose the classical implementation for event handler's.

src/ui/wxWidgets/PWSgrid.cpp
rafaelx: It's necessary to capture 'this' to be able to call the base class method 'AutoSizeColumn' within the lambda/event handler.

src/ui/wxWidgets/PWSgrid.cpp
dskrvk: Checked the [docs](http://docs.wxwidgets.org/trunk/overview_events.html#overview_events_bind) - there's no mention of the requirement to unbind.

src/ui/wxWidgets/PWSgrid.cpp
dskrvk: I see

src/ui/wxWidgets/PWSgrid.cpp
rafaelx: Checked for memory leaks with and without this patch, but haven't found anything suspicious pointing to the application.

PR #410, opened 2018-02-01
Fixes for Big-Endian Architectures

bblough:
While passwordsafe compiles on big-endian architectures, many of the tests fail, and the program is unusable.

I've found the causes of the test failures and attempted to address them.  More details about the specific changes are in the individual commit messages.  I'd of course be happy to answer any questions, or to receive feedback/suggestions.

These changes were tested on s390x (64-bit) and mips (32-bit) running Linux.  With these changes, all tests pass.

I also verified that these changes didn't cause breakage on my little-endian (Linux on amd64) system.

I don't have a windows system to test against, but expect it to be fine, as the changes are either ifdef wrapped for big-endian systems or are changes that *should* cause issues in Linux and Windows equally if they were to occur.

### DIFF
--- a/src/core/Util.h
+++ b/src/core/Util.h
@@ -102,7 +102,9 @@ inline int32 getInt32(const unsigned char buf[4])
 #if defined(_DEBUG)
   // Following code works for big or little endian architectures but we'll warn anyway
   // if CPU is really little endian
-  if ( *reinterpret_cast<const int32 *>(buf) == (buf[0] | (buf[1] << 8) | (buf[2] << 16) | (buf[3] << 24)) )
+  if (
+      *reinterpret_cast<const int32 *>(buf) != 0 &&
+      *reinterpret_cast<const int32 *>(buf) == (buf[0] | (buf[1] << 8) | (buf[2] << 16) | (buf[3] << 24)) )
   {
     pws_os::Trace0(_T("Warning: PWS_BIG_ENDIAN defined but architecture is little endian\n"));
   }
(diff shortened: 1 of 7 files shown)


src/core/Util.h
```diff
@@ -158,7 +160,9 @@ inline void putInt32(unsigned char buf[4], const int32 val )
 #if defined(_DEBUG)
   // Above code works for big or little endian architectures but we'll warn anyway
   // if CPU is really little endian
-  if ( *(int32*) buf == (buf[0] | (buf[1] << 8) | (buf[2] << 16) | (buf[3] << 24)) )
+  if (
+      *reinterpret_cast<const int32 *>(buf) != 0 &&
+      *(int32*) buf == (buf[0] | (buf[1] << 8) | (buf[2] << 16) | (buf[3] << 24)) )
```
dskrvk: Could you replace the C-style cast here with `reinterpret_cast`?

bblough: Done.

ronys:
Thanks for taking the initiative on this!

Did you also check cross-compatibility? That is, that a pwsafe file written on a big-endian machine is readable on a little-endian one, and vice-versa? I would love to do this, but I don't have access to big-endian machines.

bblough:
I don't seem to have X11 access on these architectures, so I haven't tested via the GUI at all.

For database testing, I took the databases created by running tests on one architecture, and used them to run the tests on the other architecture, and vice versa.  (avoiding deleting/overwriting as necessary)

It would be nice to be able to manually test everything (I'm still trying to work out a way to do that), but for the time being, the above is all I've been able to do.

bblough:
Actually, it just occurred to me that I might have missed something, I need to go back and check.  Please hold until I verify.

bblough:
OK, nevermind.  I thought I had forgotten to do certain tests, but I went back through my notes and it appears I did them.

ronys:
Tested on Windows 32 and 64 bit - looks good - thanks!

PR #419, opened 2018-02-18
Change occurences of NULL macro to typesafe pointer literal nullptr

rafaelx:


ronys:
Restricting this change to just the ui/wxWidgets directory is a bit odd. Merging into master to avoid accumulating conflicts, but doing the same for os/unix and core in separate commit would be nice, I think. I'll handle the Windows code when that's done.

PR #422, opened 2018-02-19
Add help folder from build directory to search path for debug build

rafaelx:


src/os/unix/dir.cpp
```diff
@@ -184,7 +184,11 @@ stringT pws_os::gethelpdir(void)
 {
 #ifdef __FreeBSD__
   return _S("/usr/local/share/doc/passwordsafe/help/");
+#else
+#if defined(_DEBUG) || defined(DEBUG)
```
ronys: Why not switch this around to support debugging under FreeBSD?

PR #425, opened 2018-02-24
Change unscoped enum types to scoped enum types for more type safety

rafaelx:


dskrvk:
Looks good. Would this cause any serialized representations (e.g. options) to change and become incompatible?

rafaelx:
The enums that I have changed so far are not related to functions of the wx API, the core API or anything in the os directory. To benefit more from type safety of C++ enums across compilation units the enums of the core should be reconsidered. There are many locations within the code where integers and enums are used equivalent. To not break any functionality I haven't touched the core and was only focused on the wx implementation.

PR #426, opened 2018-02-24
Prevent adding an additional slash to users preferences directory

rafaelx:
pws_os::getenv already ensures that directory path ends with a slash.
Adding a slash in 'createuserprefsdir' leads to have two slashes
between the home and preferences directory.

### DIFF
--- a/src/os/unix/dir.cpp
+++ b/src/os/unix/dir.cpp
@@ -129,7 +129,7 @@ static stringT createuserprefsdir(void)
 {
   stringT cfgdir = pws_os::getenv("HOME", true);
   if (!cfgdir.empty()) {
-    cfgdir += _S("/.pwsafe");
+    cfgdir += _S(".pwsafe");
     struct stat statbuf;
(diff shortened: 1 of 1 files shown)


dskrvk:
Currently the [`getenv`](https://github.com/pwsafe/pwsafe/blob/e2b97ce2c1bac33af0ffd243a200ee0e88e6aa90/src/os/mac/env.cpp#L49) method always adds a trailing slash. Maybe we should add a test verifying that if there's not one already?

rafaelx:
According to line [51](https://github.com/pwsafe/pwsafe/blob/e2b97ce2c1bac33af0ffd243a200ee0e88e6aa90/src/os/mac/env.cpp#L51) and 52 of env.cpp function pws_os::getenv already checks, whether the returned home directory by std::getenv has a trailing slash or not. If it is missing, it gets appended, so no further checks or additional slashes are needed within createuserprefsdir. - That's at least my current understanding of the code.

PR #427, opened 2018-03-04
Some minor improvements to the policies dialog

rafaelx:
1. Exisitng rows of grid showing policy details are reused now, instead of always adding new rows.
2. Grid columns are using more of the available space and resizing correctly when dialog's size changes.


ronys:
Still not perfect, but definitely an improvement - thanks!

rafaelx:
@ronys What separates the dialog from the unreachable highest level of total perfection?  8^)

ronys:
Funny - running the program a second time gave a different (and much better!) dialog.

ronys:
Actually, this is repeatable. Try opening the policy dialog with database closed (should be disabled, BTW), as opposed to with a couple of policies defined.

PR #428, opened 2018-03-04
Change unscoped enum type to scoped enum type for more type safety for core/PWPolicy

rafaelx:


ronys:
Looks good.
In general, I think we can change all "internal" enums (those that are not in the official file format) to scoped.

rafaelx:
In case of this enumeration it seems that the intention was to have the possibility to combine logically the enumeration elements to indicate multiple changes on a single policy. That's why they have been named flags and got these numbers, but I haven't found in the code any line where this was done. If this is not needed I would propose to rename 'flags' to 'mode'. Mode would represent the type of a particular change that was applied to a policy. The assigned numbers to the individual enum elements could then also be removed.

PR #429, opened 2018-03-06
Implement deletion of policies

rafaelx:


src/ui/wxWidgets/ManagePwdPolicies.cpp
```diff
@@ -571,10 +572,41 @@ void CManagePasswordPolicies::OnEditPpClick( wxCommandEvent& )
 
 void CManagePasswordPolicies::OnDeleteClick( wxCommandEvent& event )
 {
-////@begin wxEVT_COMMAND_BUTTON_CLICKED event handler for ID_DELETE in CManagePasswordPolicies.
-  // Before editing this code, remove the block markers.
-  event.Skip();
-////@end wxEVT_COMMAND_BUTTON_CLICKED event handler for ID_DELETE in CManagePasswordPolicies.
+  UNREFERENCED_PARAMETER(event);
```
dskrvk: Is this to shut up the compiler? If so, why not simply omit the argument name in the method prototype?

PR #434, opened 2018-04-01
[wx] Use IsTaskBarIconAvailable override only on wxWidgets before 3.0.4/3.1.1

pm-kan:
Current check for `ldi.Id.IsEmpty()` disable tray icon on any distro that have no `/etc/lsb_release` (for example, Slackware). 
wxWidgets 3.0.4/3.1.1 added check for non-X11 DM (https://github.com/wxWidgets/wxWidgets/commit/c0d77a721, https://github.com/wxWidgets/wxWidgets/commit/0d62f728), so it should be safe to use only wx's check when we use new (runtime) versions on any distro.

### DIFF
--- a/src/ui/wxWidgets/wxutils.cpp
+++ b/src/ui/wxWidgets/wxutils.cpp
@@ -30,6 +30,7 @@
 #endif
 
 #include <wx/taskbar.h>
+#include <wx/versioninfo.h>
 
 /*
  * Reads a file into a PWScore object, and displays an appropriate msgbox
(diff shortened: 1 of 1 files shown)


src/ui/wxWidgets/wxutils.cpp
```diff
@@ -204,9 +205,15 @@ int pless(int* first, int* second) { return *first - *second; }
...
 {
 #if defined(__WXGTK__)
-  const wxLinuxDistributionInfo ldi = wxGetLinuxDistributionInfo();
-  if (ldi.Id.IsEmpty() || ldi.Id == wxT("Ubuntu") || ldi.Id == wxT("Fedora"))
-    return false;
+  const wxVersionInfo verInfo = wxGetLibraryVersionInfo();
+  int major = verInfo.GetMajor();
+  int minor = verInfo.GetMinor();
+  int micro = verInfo.GetMicro();
+  if (major < 3 || (minor == 0 && micro < 4) || (minor == 1 && micro < 1)) {
```
dskrvk: This check will also match e.g. "4.0.4". Is it possible to instead get the version as string and compare lexicographically?

pm-kan: In general, string version comparison will work incorrectly if any of parts will be more or equal to `10`, it's better to rewrite condition as 
`(major < 3 || (major == 3 && ((minor == 0 && micro < 4) || (minor == 1 && micro < 1))) )` 
(4.x whenever it'll be, shouldn't, have this bug) or more simple (as we don't use 2.x for some time) 
`(major == 3 && ((minor == 0 && micro < 4) || (minor == 1 && micro < 1)))`

dskrvk:
I’m not sure if the program would even compile with 2.x, but I’d use your
longer check just in case.

src/ui/wxWidgets/wxutils.cpp
dskrvk: I’m not sure if the program would even compile with 2.x, but I’d use your longer check just in case.

Issue #411, opened 2018-02-01
One or more prefs have changed, do you wish to cancel, even when no changes made.


B00ze64:
Using v3.45; this bug was not present in v3.43 (not sure about 3.44).

How to reproduce: Open database, open preferences, click on the Security TAB, then hit Cancel. PasswordSafe will ask you if you are sure you wish to cancel because some settings have now changed, even tho you haven't changed anything.

Thank you.

PR #436, opened 2018-04-01
Update m_hashIter on slider set and before check for changes

pm-kan:
Without update of `m_HashIter` on `PP_DATA_CHANGED` query, Preferences dialog incorrectly report about unsaved changes after canceling.
Also is seems to be right to update `m_HashIter` on any `UpdateHashIter` call, and use `MinHIslider` instead of hardcoded `0`.


pm-kan:
maybe related to #411

Issue #430, opened 2018-03-08
Renaming Password Policy leaves old on in db


ronys:
Create a password Policy called 'foo'. Rename to ' bar'. Close and open databse, you'll have both foo and bar (on Wx version)

PR #448, opened 2018-05-08
Implement management of password policies

rafaelx:


rafaelx:
Resolves #430

rafaelx:
### Command class
The Command class specifies the return type int for 'Execute()' and 'Redo()'.
The existing command implementations return all 0 (zero), so I've also used 0 (zero) as return value for the newly introduced commands, but it seems that the return code is nowhere evaluated.

### Proposals
1. Change return type from int to bool, if different return codes/values are not needed
2. Return value 1 or true on success, otherwise 0 or false
3. Change return type to void, if return value is not needed

src/ui/wxWidgets/PasswordPolicy.cpp
```diff
@@ -379,6 +380,14 @@ bool CPasswordPolicy::Verify()
     mess = _("Policy name is already in use");
     id = ID_POLICYNAME;
     retval = false;
+  } else if ((m_polname != m_oldpolname) &&
+             (
+               (m_polname == _(stringx2std(PolicyManager::DEFAULT_POLICYNAME))) ||
+               (m_polname == stringx2std(PolicyManager::DEFAULT_POLICYNAME))
```
dskrvk: Maybe add a method `PolicyManager::IsDefault(name)`? Also, why do you need both comparisons here?

ronys:
Changed to build under Windows, as well as cleaned up a little internally.

TBD - rework Windows code to take advantage of this.

Issue #446, opened 2018-04-30
The version 1.05.0-BETA for Ubuntu does not save the "Lock password database after X minutes idle"


valentinbrasovfanduel:
On Ubuntu 16.04 LTS with all software up-to-date, it does not save the minutes. Very easy to reproduce: with the option enabled, change the number of minutes to something else than the default 5. Then press OK. Then load again the settings and the displayed number is reset to 5. 

(And talking about locking the database, after it locks the database, I did not see any option to unlock it other than either restarting the whole application, or choosing to close the database, then selecting Open from the menu and having to navigate is the folder where the database is when that is not the user's home folder is (because the app does not even remember the location and does not show a recent file list in the menu either). Both these ways are painful. Can we just have an unlock button in the GUI for unlocking a locked database?)

rafaelx:
Regarding the second issue. An icon appears usually in the system tray which offers the possibility to unlock the database. Unfortunately, the Ubuntu's are affected by a [bug](https://sourceforge.net/p/passwordsafe/bugs/1443/) in the wx framework so that the system tray icon doesn't appear.
I would propose to add new menu items to the menu bar of the applications main frame to give users, which are affected by this bug, the possibility to unlock the database.

valentinbrasovfanduel:
A menu item in the application's menu for unlocking and a corresponding button on the application's toolbar would work well. Thank you.

valentinbrasovfanduel:
Ahh, then it must be because I opened the DB as read-only. Then perhaps it would be better when the DB is opened read-only, to disable the settings that cannot be saved and show a tooltip when hoovering with the mouse over them, tooltip that could say that these settings cannot be changed because they are stored in DB and the DB was opened read-only.

And speaking of read-only opened DB, the reason I am opening it read-only is because I am afraid that this product being still in beta it might mess up my DB if I try to save it from this application. So because of this, I prefer to change the DB only from the app on my Android phone (I am using the DB on both my Android phone and my Ubuntu laptop, synchronised through Google Drive). How safe/stable is saving the DB from this application? Any chance that this application may have some bugs (being still in beta) that could corrupt the DB?

PR #451, opened 2018-05-13
Add menu items for lock/unlock of DB

rafaelx:
- Added new menu items (Lock Safe/Unlock Safe) under menu item 'Edit'
  if taskbar icon is not available (see issue [#1443](https://sourceforge.net/p/passwordsafe/bugs/1443/) at SourceForge).

- Menu items 'Lock Safe', 'Unlock Safe' and 'Password Policies' are
  enabled and disabled depending on DB state.

- Related to reported issue [#446](https://github.com/pwsafe/pwsafe/issues/446) on GitHub.


dskrvk:
What's the harm in having these menu items even if the taskbar icon is available?

rafaelx:
The idea was to just provide these menu items for all those users that are affected by the missing application icon in the system tray. Of course, the newly introduced menu items could be always available, but I wasn't sure how much the Linux version may differ from the Windows version, where such menu items are not existing.

PR #454, opened 2018-05-16
Fix Makefile-based build

sauravg:
Basically, two changes:
1. Don't use <wx/intl.h> in core. Anyway, it doesn't look like we need ascii <=> unicode conversions.
2. Include PolicyManager.* in header and Makefile

### DIFF
--- a/Xcode/pwsafe-xcode6.xcodeproj/project.pbxproj
+++ b/Xcode/pwsafe-xcode6.xcodeproj/project.pbxproj
@@ -31,6 +31,8 @@
 		A2FE25911C5ACF7500210C36 /* PWSfileHeader.cpp in Sources */ = {isa = PBXBuildFile; fileRef = A2FE25891C5ACF7500210C36 /* PWSfileHeader.cpp */; };
 		A2FE25921C5ACF7500210C36 /* PWSfileV4.cpp in Sources */ = {isa = PBXBuildFile; fileRef = A2FE258B1C5ACF7500210C36 /* PWSfileV4.cpp */; };
 		A2FE25951C5ACFBD00210C36 /* PWStime.cpp in Sources */ = {isa = PBXBuildFile; fileRef = A2FE25931C5ACFBD00210C36 /* PWStime.cpp */; };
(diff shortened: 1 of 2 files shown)


src/core/StringX.cpp
```diff
@@ -156,7 +155,7 @@ template<class T> void LoadAString(T &s, int id)
   cs.LoadString(id);
   s = cs;
 #else
-  s = _(core_st[id]);
+  s = core_st[id];
```
ronys: This change breaks the translation of the text at runtime.

src/ui/wxWidgets/PasswordPolicy.cpp
```diff
@@ -24,7 +24,7 @@
 ////@end includes
 
 #include "PasswordPolicy.h"
-#include "PolicyManager.h"
+#include "../../core/PolicyManager.h"
```
ronys: Should be "core/PolicyManager.h", to be consistent with the following line.

src/core/StringX.cpp
sauravg: I'll look into this. Just wanted to keep the core wx-free.

src/ui/wxWidgets/PasswordPolicy.cpp
sauravg: I don't understand this one. Is it just for convention, or is the patch actually breaking anything?

I always thought an explicit path makes it better: compilation is faster since there's no searching for the header, you know while looking at the code where to find the header, can open the header easily from the editor (gf in vim), etc. etc.

waal70:
This is the third time I built the app on my MacBook. This time 'round, I found the app would not compile without sauravg's changes from this commit (Xcode 9.3b4).
Additionally, PolicyManager.cpp and .h needed to be explicitly added to the "core" PBXGroup in the xcode project-file.
In other words, only after this commit AND the two files added to the project, the app would successfully build on MacOS.

src/ui/wxWidgets/PasswordPolicy.cpp
ronys: Convention and consistency.
With today's IDEs, there's no real need to specify the full relative path when -I directives have it. Also, this breaks when the including file's relative path changes (rare, but happens).

PR #460, opened 2018-06-11
Add check for virtual keyboard (xvkbd)

rafaelx:
If xvkbd (virtual keyboard) doesn't exist on users PC wxExecute still returns with a valid PID, which might be a [bug](http://wxwidgets.10942.n7.nabble.com/Asynchronous-wxExecute-failure-reporting-td78084.html) in the Wx framework. From the users perspective nothing points to the reason why the virtual keyboard doesn't appear, which looks like a problem of Password Safe. There are already two tickets on SourceForge referring to a not appearing virtual keyboard.

- [#930](https://sourceforge.net/p/passwordsafe/bugs/930/) Virtual keyboard doesn't work on Debian Linux
- [#944](https://sourceforge.net/p/passwordsafe/bugs/944/) Virtual keyboard not appearing

### DIFF
--- a/src/os/file.h
+++ b/src/os/file.h
@@ -37,6 +37,7 @@ namespace pws_os {
       time_t &ctime, time_t &mtime, time_t &atime);
   extern bool SetFileTimes(const stringT &filename,
       time_t ctime, time_t mtime, time_t atime);
+  extern bool ProgramExists(const stringT &filename);
   extern const TCHAR PathSeparator; // slash for Unix, backslash for Windows
 
   // Most stdio.h routines return -1 for an error (not INVALID_HANDLE_VALUE)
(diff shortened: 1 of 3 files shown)


bblough:
Another consideration is that I don't believe that the xkvdb package is in the Depends or Recommends section of the deb control file.  Without a declared dependency, apt or dpkg don't know that xvkbd should be installed, which would lead to a larger number of users not having xvkdb installed. This dependency has been added to the official Debian package (as Recommended, but not required), but I forgot (sorry) to submit the changes upstream.  The control file used to be a template that could be edited, but now, I believe it's generated by cmake rules.

ronys:
Will add set (CPACK_DEBIAN_PACKAGE_RECOMMENDS "xvkbd (>= 3.3-1+b1)") to cmake file per @bblough 's comment

PR #467, opened 2018-07-03
Don't hardcode pkg-config in Makefile

bblough:
The os/unix makefile hardcodes the name of the pkg-config utility, which
prevents cross-compiling on some systems.

Forwarded from https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=901092


ronys:
Thanks!

PR #468, opened 2018-07-03
Cli fixes

sauravg:
1. Better indent multiline fields, like Notes & password history, especially in side-by-side diffs
2. Show human-readable values for password policy, history & double-click actions
3. While comparing, print only relevant fields, like those that were actually compared.


ronys:
Nice work! Very readable code :-)

PR #473, opened 2018-08-05
[win] Update German translation

rafaelx:


rafaelx:
Noticed the following very similar strings that might be merged to one.

#. Resource IDs: (4026)
msgid "Double-Click on entry to View/Edit it"

#. Resource IDs: (5910)
msgid "Double-click an entry to Edit/View it"

PR #477, opened 2018-08-13
[wx] Disable controls of options dialog in R-O mode

rafaelx:


ronys:
Impressive... Thanks!

PR #479, opened 2018-08-15
Fix non-cmake-based build

bblough:
The inclusion of WX calls in core required changes to the core and test
makefiles in order to successfully build without using cmake


ronys:
Thanks!

PR #481, opened 2018-08-18
[wx] Show more properties at properties dialog

rafaelx:
- Number of attachments
- Date when master password was last changed
- Database name
- Database description


ronys:
Thanks!

PR #484, opened 2018-08-26
[wip] Mac build fixes

vovkasm:
I was trying to build a fresh version for macOS (tagged as `1.06BETA`) and as always I followed instructions in [README.MAC.DEVELOPERS.txt](https://github.com/pwsafe/pwsafe/blob/1660b5433a5d6aaa2939bc5d66dfd28986dc997a/README.MAC.DEVELOPERS.txt)
Unfortunately it does not work. I can quickly fix all issues, and document them within this pull request in hope that it will be useful for future devs. Also I can work on this and polish this patchset to a mergeable state after communication about some changes.

Environment:
- macOS 10.13.6 High Sierra
- Xcode Version 9.4.1 (9F2000)

Preparations:
- wxWidgets 3.1.1 built with slightly modified configure script... mainly I deleted most old stuff, such as `--disable-compat24` flag that was removed from configure script.

Application:

After generation of pwsafe-debug.xcconfig and pwsafe-release.xcconfig, application was not built, because core target now uses wxWidgets: https://github.com/pwsafe/pwsafe/commit/4cc8cf272025070ccbe93fcc7493c7d75b22a6d8#r30291181

So (maybe this is not desirable actually) instead of fixing code, I simply try to link core with wx in the most straightforward way: I based `core` target configuration on already generated `pwsafe-*.xcconfig` files. Also it leads to `pwsafe-cli` target also being dependent on wx, which is bad, but I did the same.

It still was not able to build, because wx requires some preprocessor definitions, so I improved the script for generating xcconfig files to include `GCC_PREPROCESSOR_DEFINITIONS` settings, see https://github.com/vovkasm/pwsafe/commit/5523650e2ebeb886f71696c6c8717949d52dd514#diff-446adca1e1b66f072d52f6338af3a757. I think this improvement is useful on its own, because it allows to remove some custom definitions in xcode project file.

And still it was not able to build ))
Last change was to add missing files to `core` target, here: https://github.com/vovkasm/pwsafe/commit/c09c8969bd43740251412fe29c96e99723edb60d

After this pwsafe.app was built and runs without issues at first glance.

Please don't merge as is... because I'm almost sure that core should not depend on wxWidgets, even if it does, it should not depend on GUI part of wxWidgets.

### DIFF
--- a/Xcode/generate-configs
+++ b/Xcode/generate-configs
@@ -34,8 +34,16 @@ foreach(split /\s+/, `$wxconfig --libs --unicode=yes --debug=$dbgflag`) {
 }
 print "OTHER_LDFLAGS = -L/usr/local/lib ", $libflags, "\n";
 
-my $cxxflags;
+my @HEADER_SEARCH_PATHS;
+my @GCC_PREPROCESSOR_DEFINITIONS;
 foreach(split /\s+/, `$wxconfig --cxxflags --unicode=yes --debug=$dbgflag`) {
-	$cxxflags .= $1." " if /\A-I(.+)/;
+    if (/\A-I(.+)/) {
+        push @HEADER_SEARCH_PATHS, $1;
+    }
+    if (/\A-D(.+)/) {
+        push @GCC_PREPROCESSOR_DEFINITIONS, $1;
+    }
 }
-print "HEADER_SEARCH_PATHS = ", $cxxflags, "\n";
+print "HEADER_SEARCH_PATHS = ".join(' ', @HEADER_SEARCH_PATHS)."\n";
+print "GCC_PREPROCESSOR_DEFINITIONS = ".join(' ', @GCC_PREPROCESSOR_DEFINITIONS)."\n";
+
--- a/Xcode/pwsafe-xcode6.xcodeproj/project.pbxproj
+++ b/Xcode/pwsafe-xcode6.xcodeproj/project.pbxproj
@@ -22,6 +22,7 @@
 
 /* Begin PBXBuildFile section */
 		6FF5D4A11DA14EE80032F5B6 /* passwordsubset.cpp in Sources */ = {isa = PBXBuildFile; fileRef = 6FF5D49D1DA14AB20032F5B6 /* passwordsubset.cpp */; };
+		A29413002132D2F40054535C /* Cocoa.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = E6B1448712B26D1500415AAE /* Cocoa.framework */; };
 		A2D181461C8FF86C0018AE03 /* media.cpp in Sources */ = {isa = PBXBuildFile; fileRef = A2D181451C8FF86C0018AE03 /* media.cpp */; };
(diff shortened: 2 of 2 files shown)


ronys:
Appreciate the effort you put in making this work and documenting your changes.

Care to try moving StringX.{cpp,h} into os? If you get it to work under Mac, I'll tweak it for Windows and Linux before merigng into master.

vovkasm:
Ok will do, but before I will try to eliminate the need to have xcode project files and duplicate build rules for macos... I hope that modern cmake versions should allow us to build application bundles with minimal effort. Thought, it may take few days until I came back...

ronys:
Cool, no rush.

vovkasm:
Progress report :-)

Still WIP, current progress can be seen here: https://github.com/pwsafe/pwsafe/compare/master...vovkasm:feature/cmake-macos
(I will rebase this branch periodically)

Got something, but have some questions.
1. I did use features of CMake 3.x (not sure exact min version currently). So question: can we rely on fresh cmake versions or Debian and other Linux distros still needs backward compatibility to cmake 2.8?
2. Did we need ability to install command line client on macOS? Or this can be made in later iterations (PRs)?
3. What versions and distributions of linux should I test? What configurations should I test on windows (don't understand from documentation what exactly supported by CMakeLists.txt)?

What done:
1. pwsafe.app bundle can be built on macOS with project generated from cmake (currently tested only with Ninja and Xcode generators, "Unix Makefiles" probably should work also)
2. install target for macOS place all external deps into pwsafe.app bundle, so when installed, it does not rely on globally installed wxWidgets or xerces libs.
3. cpack can generate dmg on macOS

Still my TODO:
1. I broke help compilation (chm) on windows. Need to fix.
2. Probably I broke something other... need to test on linux and windows.
3. Tune dmg file, add background from `install/macosx`
4. Tune config variables in CMakeLists.txt - decouple bundle and standard install, so on macOS we can select between unix-style install layout and application bundle.

ronys:
Nice work! To answer your questions:

1. Debian is the slowest to update their packages, and I see that even they've moved to cmake 3.7 for the stable release (https://packages.debian.org/stretch/cmake), so that's a reasonable minimum.
2. Command line client is a nice-to-have at this stage, not mandatory for macOS
3. Ideally, Linux should be tested on: Debian - latest stable, Ubuntu: Latest LTS and latest stable, Fedora: latest stable, all for 32 and 64 bit architectures. Debian 64 bit is my personal Linux default, so at least on that...
4. Windows: MFC 32 and 64 bits.

seamusdemora:
Without trying to sound the total dufus... I stumbled on this thread just now, and don't want to jump to conclusions, but does all of this mean that a "real" pwsafe may be in the near future for mac os?

ronys:
Not sure what you mean by "real", but when this is completed and merged into the main repository, it will be possible to build a version of PasswordSafe using this code base on Mac.
This doesn't mean that there will be formal packages available as for Windows and various Linux distros, mainly because I don't have a Mac to build this on (yet...).

seamusdemora:
Oh - sorry for the confusion... by "real" I just meant to contrast it with the Java version that's currently available. I'm happy to have the Java version, but it's never had all the functions the Win32 version did, and it seems it's not been maintained for a while.

PR #485, opened 2018-08-26
[wx] Consider symbolic links

### LINKED ISSUE #1136: Merge dialog - Report dialog - change Finish button with Close button
File | Merge - Report dialog.
image
I love description of first two buttons, but the last one **Finish** is not the best. Finish is used when there are several wizard dialogs with Next buttons and then finally user gets to Finish that confirms some drastic change will happen or user has gotten to the end of wizard. I think in this case it would be better to use **Close** button, because nothing special is performed, just dialog is closed down.

rafaelx:
This resolves #1136 and #1153.


ronys:
Thanks!

PR #486, opened 2018-09-02
[wx] Add new DB property to XSD and fix history import from XML

rafaelx:


rafaelx:
Resolves [#1297](https://sourceforge.net/p/passwordsafe/bugs/1297).

ronys:
Thanks!

PR #491, opened 2018-09-23
Few updates in Russian texts

pm-kan:
Also it seems, that aee280706 should be reverted. Two `<param name="Name" ...>` elements aren't error: the first one is "Index Group" name, and the second is "Item name". 
https://www.nongnu.org/chmspec/latest/Sitemap.html#HHK


ronys:
Thanks - couldn't find the description of the duplicates in the link you referenced, but I'll take your word for it.

pm-kan:
> couldn't find the description of the duplicates in the link you referenced
It's here: "The order of the item properties is as follows: Merge | `([Name] ([Name] [Type...] [Local [URL] | See Also])`... [FrameName] [WindowName] [Comment])"

For example, multi-item object in Russian help
```
<object type="text/sitemap">
<param name="Name" value="Вкладка «Вид»">
<param name="Name" value="Меню «Сервис»">
<param name="Local" value="html/manage_menu.html">
<param name="Name" value="Вкладка «Вид»">
<param name="Local" value="html/display_tab.html">
</object>
```

PR #493, opened 2018-09-23
[wx] Let the wx framework crop text at properties dialog

rafaelx:


rafaelx:
The advantage of this solution compared to the previous one is that whenever more space gets available on the UI due to resizing of the dialog the more text gets visible. Everything is managed by wx without the need of custom code.

ronys:
Indeed, nice find. MFC has something similar, so I'm not totally surprised...

PR #496, opened 2018-10-25
[wx] Apply configured password font in Edit/View dialog

rafaelx:
Fixes reported issue [1478](https://sourceforge.net/p/passwordsafe/bugs/1478/) on SF.


rafaelx:
The reason for applying the font also to the confirmation text field is to get the asterisks shown with the same font size as in the password text field, as it is the case of the Windows version.

bblough:
Thanks for this.  I noticed this a while back and it really bugged me, but I haven't had time to do anything about it.

ronys:
Thanks!

Issue #75, opened 2016-02-21
Autolock don't closes "edit" window


mioux:
Hello,

In Linux version (pwsafe-0.95BETA-3.x86_64 on Fedora), if you don't close an "edit" window, you wait for autolock, and then, click on the task bar, pwsafe don't have closed the edit window, and you can access current opened "edit" window.
![pwsafe-bug](https://cloud.githubusercontent.com/assets/6904170/13201827/a364967e-d882-11e5-973a-60a9cdf2ef10.png)

I had 2 "unlock" window too, as I have restored passwordsafe without unlocking the safe, waited for autolock, and re-restored passwordsafe.

I don't have tested on Windows version.
Version 0.96 fails on start, no RPM for 0.97.

rafaelx:
Are you still able to reproduce this with the latest version?

ronys:
If a user has modified an entry but hasn't closed the Edit window, it would be right to suspend autolock, since locking writes the database, and it's unclear if the user intended to save the change or not. That, at least, is how the Windows version behaves.

rafaelx:
I've tested this under Windows 7 with version 3.47.02 and observed that the autolock kicks in and hides all application specific windows even while the Edit dialog is still open and the selected entry is modified. 
Restoring the application shows that the modifications were stored.

dskrvk:
Another solution would be to alert the user (e.g. by bringing the window to top) that they have unsaved changes that are preventing the database from locking. This could help avoid the issue of indefinitely postponing the lock.

rafaelx:
This error report addresses two issues.

**Issue 1**: _Edit/View dialog remains open when DB was automatically locked after timeout_
I haven't observed this on Ubuntu (Gnome) and Linux Mint Debian Edition 2 (Cinnamon based on Gnome), but on Kubuntu (KDE). On Kubuntu, the Edit/View dialog even loses partly its modal characteristic. After unlocking it is suddenly possible to interact with the applications main window and initiate some functionality.

**Issue 2**: _Two passphrase dialogs appear to unlock DB_
I was not able to reproduce this on Ubuntu, LMDE2 and Kubuntu.

mioux:
I can't reproduce ithese on ubuntu 18.04 with the latest version

Issue #59, opened 2016-01-12
Uncommitted new entries can be lost after database auto-lock when auto-read-only is enabled


kappa7194:
How to reproduce
1. Enable the “Open database read-only by default” option (“System” tab);
2. Enable the “Lock password database after” option (“Security” tab);
3. Open a database in read-write;
4. Open the “Add entry” dialog and fill it without committing (do not click “OK”);
5. Wait for the auto-lock timeout to expire and PasswordSafe to minimize;
6. Reopen PasswordSafe and unlock it — don't touch the “Open as read-only” checkbox;
7. Click the “OK” button to save the entry you were adding;
8. Notice that the entry has not been added to the database;
9. Weep.

The problem lies in steps 6 and 7: if you don't notice — or remember to unckeck — the “Open as read-only” checkbox, when you unlock the database you're shown the entry you were adding before the database auto-locked and you're allowed to “save” the entry by clicking the “OK” button, however since the database is now read-only the saving of the entry fails silently and its information is lost.

This probably applies to entries editing too (or any other kind of change which can be in an “uncommited” state).

I think this problem could be prevented in two different ways (I don't know if either is feasible though):
- don't allow the database to be unlocked as read-only if there are changes pending changes;
- don't allow the user to “OK” the “Add entry” dialog if the database has been unlocked in read-only ode;
- ask the user to switch the database in read-write when he tries to “OK” the “Add entry” dialog;

I think the third solution would be the better from a UX point and — possibly — the easiest one code-wise.

vossenjp:
The problem is worse than this.  If you open a DB in Read-only, forget, and make **any** change it is silently discarded!  At least it just was in 0.97 when I tested (not using 0.98 as noted in #90).

Steps to reproduce:
1. Close PWS
2. Open DB in read-only
3. Create a test entry and click "OK"
4. Test entry silently discarded!
5. Edit an existing entry and make a change
6. Create a test entry and click "OK"
7. Test change silently discarded!

kappa7194:
>The problem is worse than this.

I don't know if the issues are related, my issue is about the Windows GUI (I don't know if and how much code shares with the Linux one).

ronys:
Windows version fixed in commit 2180aa3 by disabling the ability to turn on read-only when unlocking a database with changes pending.

dskrvk:
This is what happens for me (v.0.99 wxWidgets, Ubuntu 15.10):

1. Open database in r/w mode.
[...]

PR #499, opened 2018-11-04
[wx] Make Edit/View dialog not editable if DB is read-only

rafaelx:
**AddEditPropSheet Dialog**
- If DB is in read-only mode corresponding controls of Edit/View dialog are not editable.
- If application gets iconized and Edit/View dialog wasn't iconized by window manager than it will be iconized as well.

**Menu Item**
- If DB is in read-only mode menu item from menu bar and context menu is shown as "View Entry..." otherwise as "Edit Entry..." as it is the case for the Windows version.

**Related Issues**
- [#75](https://github.com/pwsafe/pwsafe/issues/75)
- [#59](https://github.com/pwsafe/pwsafe/issues/59)


src/ui/wxWidgets/addeditpropsheet.cpp
```diff
@@ -1584,6 +1614,109 @@ void AddEditPropSheet::OnUseHexCBClick( wxCommandEvent& /* evt */ )
...
+      break;
+    }
+    case ID_COMBOBOX:
+    case ID_COMBOBOX2:
+    case ID_CHECKBOX1:
+    case ID_SPINCTRL:
+    case ID_GRID:
+    case ID_BUTTON1:
+    case ID_BUTTON4:
+      event.Enable(!dbIsReadOnly);
```
dskrvk: This looks cryptic. Maybe these controls didn't have any logic attached to them before, so they had these generic names, but now they have different behavior, so maybe it's time to give them meaningful names?

dskrvk: Awesome!

ronys:
Good work - thanks!

PR #500, opened 2018-11-05
[wx] Close application properly on receipt of SIGHUP

rafaelx:
Resolves reported issue [#1242](https://sourceforge.net/p/passwordsafe/bugs/1242) on SourceForge.

### DIFF
--- a/src/os/unix/cleanup.cpp
+++ b/src/os/unix/cleanup.cpp
@@ -11,7 +11,7 @@
 #include <iostream>
 
 namespace {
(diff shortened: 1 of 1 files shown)


ronys:
Hm,

With daemon processes, the convention is that SIGHUP causes the configuration file to be re-read, the idea being that this be sent to the daemon process after modifications to the configuration, rather than stopping and restarting the process.

I guess this would be an overkill for PasswordSafe, since it is not meant to be run as a daemon, so handling SIGHUP as the rest is fine.

PR #501, opened 2018-11-11
[wx] Add exit button to CSafeCombinationPrompt dialog

rafaelx:
Resolves reported issue [1159](https://sourceforge.net/p/passwordsafe/bugs/1159/) on SourceForge.


src/ui/wxWidgets/passwordsafeframe.cpp
```diff
@@ -2550,7 +2552,11 @@ void PasswordSafeFrame::UnlockSafe(bool restoreUI, bool iconizeOnFailure)
   }
   StringX password;
   if (m_sysTray->IsLocked()) {
-    if (VerifySafeCombination(password)) {
```
dskrvk: This was the only usage of this method - maybe remove it?

rafaelx:
- Removal of unused method 'PasswordSafeFrame::VerifySafeCombination'
- Removal of trailing whitespace characters
- Sort of includes
- Change of 'if-else if' chain to 'switch-case'

PR #502, opened 2018-11-21
[wx] Close application on ESC key event

rafaelx:
Resolves reported issue [1033](https://sourceforge.net/p/passwordsafe/bugs/1033/) on SourceForge.


ronys:
Indeed - thanks!

PR #504, opened 2018-11-24
[wx] Unification of password rules dialogs

rafaelx:
- Width of all spinners is unified to 60 or 80 pixels depending on spinners max. value
- Value range of all spinners is set to values from preferences
- Takeover of AddEditPropSheet::OnAtLeastChars to CPasswordPolicy for an uniform behavior

### DIFF
--- a/src/ui/wxWidgets/PasswordPolicy.cpp
+++ b/src/ui/wxWidgets/PasswordPolicy.cpp
@@ -415,7 +415,7 @@ void CPasswordPolicy::CreateControls()
     itemStdDialogButtonSizer39->Show(false);
     itemStdDialogButtonSizer40->Show(true);
 
-     EnableSizerChildren(m_itemStaticBoxSizer6, !m_pwpUseNamedPolicyCtrl->IsChecked());
+    EnableSizerChildren(m_itemStaticBoxSizer6, !m_pwpUseNamedPolicyCtrl->IsChecked());
   }
   else {
     itemBoxSizer3->Show(true);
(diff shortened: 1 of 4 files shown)


src/ui/wxWidgets/PasswordPolicy.cpp
```diff
@@ -818,3 +853,27 @@ void CPasswordPolicy::OnCopyPassword( wxCommandEvent& WXUNUSED(event) )
...
+void CPasswordPolicy::OnAtLeastChars( wxSpinEvent& WXUNUSED(event) )
+{
+  wxSpinCtrl* spinCtrls[] = {m_pwpUCSpin, m_pwpLCSpin, m_pwpDigSpin, m_pwpSymSpin};
+  int total = 0;
+  for (size_t idx = 0; idx < WXSIZEOF(spinCtrls); ++idx) {
+    if (spinCtrls[idx]->IsEnabled())
+      total += spinCtrls[idx]->GetValue();
+  }
+
+  if (total > m_pwpLenCtrl->GetValue())
```
dskrvk: Why not `>=`?

src/ui/wxWidgets/PasswordPolicy.cpp
```diff
@@ -818,3 +853,27 @@ void CPasswordPolicy::OnCopyPassword( wxCommandEvent& WXUNUSED(event) )
...
+ * the password length has to be bigger than the sum of all
+ * "at least" lengths.  This is not comprehensive & foolproof
+ * since there are far too many ways to make the password length
+ * smaller than the sum of "at least" lengths, to even think of.
+ *
+ * In OnOk(), we just ensure the password length is greater than
+ * the sum of all enabled "at least" lengths.  We have to do this in the
+ * UI, or else password generation crashes
+ */
+void CPasswordPolicy::OnAtLeastChars( wxSpinEvent& WXUNUSED(event) )
```
dskrvk: I would change the name to `OnAtLeastPasswordChars` since this handler is specific to that group of controls and is not meant to react to any others.

src/ui/wxWidgets/PasswordPolicy.cpp
rafaelx: I just reused the already existing code from _AddEditPropSheet_ without paying to much attention what it actually does, because I found the behavior at the UI convincing.
However, wouldn't the additional check for equality mean to update the password length control with a value it already shows?

src/ui/wxWidgets/PasswordPolicy.cpp
rafaelx: Mmmmmm...ok

src/ui/wxWidgets/PasswordPolicy.cpp
rafaelx: Indeed, the description is confusing and should be changed, because it doesn't reflect properly the implementation.

At line [502](https://github.com/pwsafe/pwsafe/blob/307d2fe121e78340ffb3d290626f350f328379a6/src/ui/wxWidgets/PasswordPolicy.cpp#L502) you can find the corresponding check regarding the given length information. It fails, if the sum of uppercase min., lowercase min., numbers min. and symbols min. is greater than the password length.

      if (!(m_pwUseHex || m_pwUseEasyVision || m_pwMakePronounceable) &&
          (m_pwDigitMinLength + m_pwLowerMinLength +
           m_pwSymbolMinLength + m_pwUpperMinLength) > m_pwdefaultlength) {
        mess = _("Password length is less than sum of 'at least' constraints");
        id = ID_PWLENSB;
        retval = false;
      }

Hence, the sum is only allowed to be equal or less than the password length. If this is not fulfilled the settings are not accepted and the user gets informed via a message (line [503](https://github.com/pwsafe/pwsafe/blob/307d2fe121e78340ffb3d290626f350f328379a6/src/ui/wxWidgets/PasswordPolicy.cpp#L503)).

Back to the line in question, 877, our edge case. If the calculated sum of characters corresponds to the min. required password length the password length doesn't need to be adjusted.

dskrvk: Ohh, I misunderstood what this was doing. It adjusts the overall length based on the sum - makes sense now. TBH, neither the method name nor the comment gives the idea that the goal here is to keep the various length controls in sync.

Perhaps add this to the doc: "Ensures that the sum of each character class' minimum counts doesn't exceed the overall password length, increasing it as necessary."

ronys:
Thanks to both Dmitry and Rafael - good constructive discussion.

PR #506, opened 2018-12-08
XSLT template for KeepassX v0.4.3 XML

pm-kan:
was requested at SF BR 1173

as usual
- only supported autotype template placeholders are translated to pwsafe's autotype
(unknown copied to autotype string as is), also original autotype string stored at Notes,
- binary attachments are ignored (but it's description copied to Notes)


ronys:
Thanks again - needs to be added into the package building inventories.

PR #507, opened 2018-12-11
[wx] Fix Version Check

rafaelx:


rafaelx:
### Description
Currently, the version check doesn't work, because the Wx framework seems not to support SSL/TLS.
As alternative the Curl library was chosen which supports various SSL/TLS backends.
The new implementation that fetches "latest.xml" from "https://pwsafe.org" is based on the Open SSL backend, which is licensed under the 4-clause BSD license.
If necessary, due to any reasons, the backend can be changed to GNU TLS for instance, without any impact on the API. Probably, only some dependencies will change.

See also the comparison of SSL libraries for more information.
https://curl.haxx.se/docs/ssl-compared.html

### Introduced Dependencies

- libcurl4 & libssl1.1 ; OpenSSL flavor (alternative: libcurl3-gnutls)
- libcurl4-openssl-dev & libssl-dev (alternative: libcurl4-gnutls-dev)

### Implementation
The UI shouldn't be blocked during the data transfer to give the user still the ability to cancel the version check. For this reason the file download is handled by a secondary thread.

### To be considered

- [ ] Was the cmake control file correctly updated regarding the newly introduced dependencies?
- [ ] Update of IDE (CodeBlocks, CodeLite) specific configuration/project files.
- [ ] Update of Makefiles.
- [ ] Update of README.*.DEVELOPERS.md files regarding the newly introduced dependencies.
- [ ] New strings has been introduced which need a translation (are they correctly phrased?).
- [ ] Remove some of the output to stdout, which still exists for testers/reviewers of this pull request.

ronys:
Nice work - thanks!

PR #514, opened 2018-12-29
Make OK button default in safe unlock dialog

dskrvk:
Previously if I typed in a safe combination and pressed Enter nothing
would happen because the OK button didn't have focus by default.


ronys:
Thanks!

PR #517, opened 2019-01-07
Adding a set of missing Hungarian translations

akossomfai:
Adding a set of missing Hungarian translations. Some items were relocated by poedit within the file so the change looks quite bigger than it really is.


ronys:
Thanks!

PR #522, opened 2019-02-01
[wx] Show recently used DBs on File menu

rafaelx:
Resolves reported issue [1484](https://sourceforge.net/p/passwordsafe/bugs/1484).

### DIFF
--- a/src/ui/wxWidgets/passwordsafeframe.cpp
+++ b/src/ui/wxWidgets/passwordsafeframe.cpp
@@ -437,7 +437,25 @@ void PasswordSafeFrame::CreateMenubar()
   itemMenu3->Append(wxID_OPEN, _("&Open..."), wxEmptyString, wxITEM_NORMAL);
   itemMenu3->Append(wxID_CLOSE, _("&Close"), wxEmptyString, wxITEM_NORMAL);
   itemMenu3->Append(ID_LOCK_SAFE, _("&Lock Safe"), wxEmptyString, wxITEM_NORMAL);
-  itemMenu3->AppendSeparator();
+
+  if (wxGetApp().recentDatabases().GetCount() > 0) {
+
+    // Most recently used DBs listed directly on File menu
+    if (PWSprefs::GetInstance()->GetPref(PWSprefs::MRUOnFileMenu)) {
+      wxGetApp().recentDatabases().AddFilesToMenu(itemMenu3);
+    }
+    // Most recently used DBs listed as submenu of File menu
+    else {
+      wxMenu* recentSafesMenu = new wxMenu;
+      wxGetApp().recentDatabases().AddFilesToMenu(recentSafesMenu);
+      itemMenu3->AppendSeparator();
+      itemMenu3->Append(ID_RECENTSAFES, _("&Recent Safes..."), recentSafesMenu);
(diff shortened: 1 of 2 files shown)


src/ui/wxWidgets/passwordsafeframe.cpp
```diff
@@ -1167,6 +1185,7 @@ void PasswordSafeFrame::OnOpenClick( wxCommandEvent& /* evt */ )
 
   if (rc == PWScore::SUCCESS)
     m_core.ResumeOnDBNotification();
+    CreateMenubar(); // Recreate the menu with updated list of most recently used DBs
```
rafaelx: 1. Curly braces are missing, so that CreateMenubar() is only called on SUCCESS.
2. Is this the desired behavior that a new opened DB appears immediately in the list of most recently used DBs?

src/ui/wxWidgets/passwordsafeframe.cpp
```diff
@@ -2779,20 +2799,25 @@ void PasswordSafeFrame::OnOpenRecentDB(wxCommandEvent& evt)
   {
     case PWScore::SUCCESS:
       m_core.ResumeOnDBNotification();
+      CreateMenubar();  // Recreate the menu with updated list of most recently used DBs
```
rafaelx: Is this the desired behavior that a new opened DB appears immediately in the list of most recently used DBs?

src/ui/wxWidgets/passwordsafeframe.cpp
ronys: 1 - certainly a bug
2 - this is arguable. I think it would be more useful *not* to add the currently opened database to the list as long as it is opened, as this would push off the list an older database that might be of interest, without really adding any functionality.

rafaelx: 1 - Fixed
2 - Still meditate on it  ;)

rafaelx: **Issue 2**
Checked a few applications, like LibreOffice, Kate, CodeBlocks and some others. Right after opening a new file it appears immediately in the recent open files list of these applications.

Playing with the code and various possibilities I'm not sure which strategy to choose for handling the history of previously opened files to fulfill your idea, ronys.

Whenever a new file was opened successfully it will be added to the list of recent open files. This in turn changes each time the internally managed file history and therefor the index of menu items. At this point the menu structure needs to be resynchronized, resp. recreated, otherwise the menu items are pointing to wrong files. If the menu is recreated the currently open file/DB appears in the menu structure and an older entry disappers if the configured maximum of files is reached.

The only idea that comes into my mind right now to overcome this is to add only a file into the history list when it was closed, so that the history list only reflects closed files. Maybe something like this, which complicates everything a little more, especially when the file doesn't get closed explicitly, but implicitly by opening a new one.

Any ideas are very appreciated.

ronys: Thanks for doing the research on this. Looks like adding the file immediately is the path of least resistance. Since this is the popular behavior, I would conclude that the other approach isn't worth the additional effort/complexity. Vox populi vox dei...

PR #523, opened 2019-02-02
Fix enabled/disabled state for static text of System tray options

akossomfai:
DImming did not work for the "Remember last" option under Options/System/System-tray/Put-icon-in-system-tray. Fixing it.


rafaelx:
"Remember last" will not be disabled as it is the case for the Windows version, which acts as the reference for the Linux version, but I agree that disabling also "Remember last" provides a more consistent visual appearance.

ronys:
Thanks!

PR #524, opened 2019-02-08
[wx] Remove orphan plk file

rafaelx:


src/core/PWScore.cpp
```diff
@@ -3255,6 +3255,12 @@ void PWScore::UpdateWizard(const stringT &s)
 
 bool PWScore::LockFile(const stringT &filename, stringT &locker)
 {
+  // If there is a matching plk file to the database (filename)
```
ronys: Actually, the Windows implementation of pws_os::LockFile() succeeds when the locking process no longer exists, so it's more accurate to move IsLocked/TryUnlock into the Unix implementation of pws_os::LockFile() - this way both implementations have the same semantics.

PR #525, opened 2019-02-09
[wx] Expand implementation of RunCommand

rafaelx:
Resolves [1486](https://sourceforge.net/p/passwordsafe/bugs/1486/)


ronys:
Looks like a straightforward port of the corresponding Windows DboxMain::OnRunCommand() function.
Almost worth refactoring to core to avoid the duplication...
Thanks.

PR #528, opened 2019-02-10
[wx] Implement UpdateLastClipboardAction

rafaelx:


ronys:
Nice port, but the effect doesn't look very good. In the attached, the top status bar is from the wx pwsafe, and the lower one is the Windows pwsafe.
I think that reducing the font size of the wx status bar by a couple of points would allow as much text to be displayed as the Windows version, making this feature effective. As it is, the texts are all truncated, which is confusing for the user.

ronys:
statbars

rafaelx:
Reduced the width of one of the fields (CPWStatusBar::Field::READONLY) and made the font smaller.
The screenshot shows borders for an easier review of how the changes are affecting the status bar.
Removed borders in latest commit.

statusbar

Issue #551, opened 2019-05-04
MINT 19.1 Pwsafe requires libcurl3 but 4 is installed


IanBlakeley:
Linux MINT 19.1

After installing virtualbox 6 libcurl3 gets replaced by libcurl4, this breaks pwsafe. If I reinstall libcurl3 `apt install libcurl3 libcurl-openssl1.0-dev` then libcurl4 is removed along with virtualbox cannot have both versions of libcurl installed at the sametime.

dskrvk:
I faced the same issue on Ubuntu 18.04 and fixed it by changing the
dependency to libcurl4 and recompiling. I can submit a pull request.

ronys:
Fixed in commit fdf4298ae.

Will be in 1.08.1

ronys:
1.08.1 released with .deb for Mint 19.1

wickedsp1d3r:
Could you make this change also to Debian builds? Debian 10 (Buster) just released and this issue exists.

ronys:
Done. 1.08.2 just released with 64 bit Debian 10 support.

PR #552, opened 2019-05-04
Update runtime version of libcurl to match compile version

dskrvk:
This fixes a dependency conflict error when installing on Ubuntu. The
compile-time version is libcurl4*, so runtime one should match.

Related to https://github.com/pwsafe/pwsafe/issues/551


dskrvk:
If we want to support all of these OSs we need a more dynamic way of choosing the `libcurl` version. I'm not sure how the current Linux version can be installed anywhere at all given the conflict between the compile- and runtime versions. I don't know much about cmakefiles, so this was just a quick fix to get it to install on my system.

rafaelx:
The CMake file currently only considers the distribution ID (Ubuntu, Debian, ...) (via 'lsb_release -is'). With help of the codename ('lsb_release -cs') could be distinguished finer between the distribution versions (Ubuntu: bionic, cosmic, disco, ...; Debian: stretch, buster, ...; Linux Mint: tessa, ...).

ronys:
Fixed the CMake file to make the right distinction between versions of Ubuntu.

PR #556, opened 2019-05-04
Option to disable online version check

pm-kan:
When active, it'll remove dependency on curl + ssl (if xerces build without curl too) and, link will be changed to open https://pwsafe.org/news.shtml in browser.
Also C::B project updated.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -91,6 +91,7 @@ endif (pwsafe_REVISION)
 # Configurable options:
 option (NO_YUBI "Set ON to disable YubiKey support" OFF)
 option (NO_GTEST "Set ON to disable gtest unit testing" OFF)
+option (NO_VERCHECK "Set ON to disable online version check" OFF)
 
 if (WIN32)
   option (WX_WINDOWS "Build wxWidget under Windows" OFF)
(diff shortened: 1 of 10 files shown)


dskrvk (approved):

ronys:
What's the rationale behind NO_VERCHECK? Looks like a hack to workaround the libcurl3/4 which was resolved for Ubuntu.
Or am I missing something?

dskrvk:
Not the author, but I think the benefit is completely cutting out network I/O from the program.

pm-kan:
I'm mainly using Slackware so can't say if there are any unsolved issues with curl on Ubuntu. This option was added to reduce number of dependencies and network-related code when online version check isn't used (for example, when package built locally or by distro maintainer). But I forgot that xerces-c also may be built with curl-support and, in this case, number of dependencies will remain the same.

I have no objections if this pull request will be closed without merge if this option is useless in most cases.

bblough:
Just to add another point of view - 

Some downstreams (Debian in particular) prefer to handle updates through the distro package manager.  So the version check, while certainly good/applicable for the official packages provided here, is not really useful or desirable for some downstreams.

So I do see benefit to having this configurable as a build option.  Though, it isn't strictly necessary, as the functionality can always be patched out in the distro packages if desired.

madmartin:
Hello,
I would also appreciate the possibility to turn off automatic update checking - because 
* users installing this package through the distribution should **never** be guided to install an update beside the distribution software management.
* some people dont like security sensitive tools to open network connections at all

Just add an option to DISABLE update checking and leave the default enabled, then everybody should be happy.

ronys:
> Hello,
> I would also appreciate the possibility to turn off automatic update checking - because
> 
> * users installing this package through the distribution should **never** be guided to install an update beside the distribution software management.
> * some people dont like security sensitive tools to open network connections at all
> 
> Just add an option to DISABLE update checking and leave the default enabled, then everybody should be happy.

Just to clarify - the update checking is *not* automatic, but rather invoked by the Help->About "Click here to check" action.
Furthermore, PasswordSafe will not open a network connection with an open database - you're prompted to close the database before the check is performed.

dskrvk:
Related: https://github.com/pwsafe/pwsafe/issues/1849

PR #558, opened 2019-05-12
[wx] Add dependencies for LinuxMint and generate correct package name

rafaelx:


CMakeLists.txt
```diff
@@ -399,74 +399,154 @@ install (TARGETS pwsafe RUNTIME DESTINATION "bin")
...
 
+#
+# Package file name is project name.
+#
+if (NOT CPACK_PACKAGE_NAME)
+  string(TOLOWER "${PROJECT_NAME}" CPACK_PACKAGE_NAME)
+endif ()
+
+#
+# Determine the architecture (amd64, i386) with command 'dpkg --print-architecture'
```
ronys: 'dpkg' isn't part of the RPM-based distros. A more portable way to do this is via the 'arch' command or 'uname -m'.

Do you have the ability to test this pull request on Fedora or CentOS?

rafaelx: My goal was to get 'amd64' as in the case of the already published packages. This was only provided by 'dpkg'. 'arch' and 'uname -m' provide 'x86_64' instead of 'amd64'. However, 'x86_64' could be easily replaced by 'amd64'. Please let me know which identifier is desired.
I would very appreciate if you could do the testing on Fedora or CentOS.

ronys: I'd like to map x86_64 as returned by 'uname -m' (more portable that arch) to arm64.

I'll do the work on RPM distros once you're done with this part.

ronys:
Merged into master, still need to get rpm names to work...

PR #572, opened 2019-08-01
Improvements at Policy Handling

rafaelx:
Motivated by [1493](https://sourceforge.net/p/passwordsafe/bugs/1493/).

### DIFF
--- a/src/ui/wxWidgets/addeditpropsheet.cpp
+++ b/src/ui/wxWidgets/addeditpropsheet.cpp
@@ -238,30 +238,30 @@ void AddEditPropSheet::Init()
 ////@end AddEditPropSheet member initialisation
 }
 
-class PolicyValidator : public MultiCheckboxValidator
-{
-public:
-  PolicyValidator(int rbID, int ids[], size_t num,
-      const wxString& msg, const wxString& title)
-    : MultiCheckboxValidator(ids, num, msg, title), m_rbID(rbID) {}
-  PolicyValidator(const PolicyValidator &other)
-    : MultiCheckboxValidator(other), m_rbID(other.m_rbID) {}
-  ~PolicyValidator() {}
+// class PolicyValidator : public MultiCheckboxValidator
+// {
+// public:
+//   PolicyValidator(int rbID, int ids[], size_t num,
+//       const wxString& msg, const wxString& title)
+//     : MultiCheckboxValidator(ids, num, msg, title), m_rbID(rbID) {}
+//   PolicyValidator(const PolicyValidator &other)
+//     : MultiCheckboxValidator(other), m_rbID(other.m_rbID) {}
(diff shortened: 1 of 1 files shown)


src/ui/wxWidgets/addeditpropsheet.cpp
```diff
@@ -1229,276 +1229,345 @@ static short GetSelectedDCA(const wxComboBox *pcbox,
...
+            }
+          }
+          // User has selected to use an item specific policy
+          else {
+            // Use the data of the item specific policy collected from the UI
+            m_item.SetPWPolicy(pwp);
+
+            // Remove database policy information from item
+            m_item.SetPolicyName(tostringx(wxEmptyString));
+          }
```
dskrvk: It's my own style opinion but this version is much more verbose than the previous.

src/ui/wxWidgets/addeditpropsheet.cpp
```diff
@@ -238,30 +238,30 @@ void AddEditPropSheet::Init()
...
+//       wxRadioButton* rb = wxDynamicCast(win, wxRadioButton);
+//       if (rb && rb->GetValue()) {
+//   return true;
+//       }
+//     }
+//     return MultiCheckboxValidator::Validate(parent);
+//   }
+// private:
+//   int m_rbID;
+// };
```
dskrvk: Why not just remove this code?

src/ui/wxWidgets/addeditpropsheet.cpp
rafaelx: However, due to the additional new code, the resulting verbosity offers the advantage of more stability.

src/ui/wxWidgets/addeditpropsheet.cpp
rafaelx: I thought this might reflect an idea that will be followed up by somebody at a later time.

src/ui/wxWidgets/addeditpropsheet.cpp
dskrvk: I don't feel strongly enough about this and am dropping this comment.

src/ui/wxWidgets/addeditpropsheet.cpp
dskrvk: Personally I think commented-out chunks are a code smell, but again this is a minor issue.

rafaelx: @ronys I would also tend to remove this code if nobody has it on its radar.

ronys: It's elegant, in its way, so I'm leaving it for now.
Note that if we remove it, we should also remove its base class in wxutil.cpp/h

PR #575, opened 2019-08-19
Minor fixes for policy settings

pm-kan:
- don't show assert, when only hexadecimals are selected
- new description wasn't used for pronounceable passwords


ronys:
Thanks!

PR #578, opened 2019-08-24
[wx] Implement more font settings

rafaelx:


rafaelx:
Resolves [1478](https://sourceforge.net/p/passwordsafe/bugs/1478/), except for virtual keyboard.

src/ui/wxWidgets/wxutils.h
```diff
@@ -59,9 +59,25 @@ inline wxString& operator<<(wxString& w, const stringT& s) {
```
ronys: Wouldn't it be more elegant to replace the following 3 with a single
ApplyPrefFont(wxWindow *win, const PWSprefs::StringPrefs fontPref) ?

src/ui/wxWidgets/wxutils.h
```diff
@@ -57,27 +57,23 @@ inline wxString& operator<<(wxString& w, const stringT& s) {
...
-    win->SetFont(font);
-  }
-}
-
-inline void ApplyNotesFont(wxWindow* win)
-{
-  wxFont font(towxstring(PWSprefs::GetInstance()->GetPref(PWSprefs::NotesFont)));
-  if (font.IsOk()) {
-    win->SetFont(font);
+  switch (fontPreference)
```
ronys: Hm. Since we're calling this directly with hardcoded enum values, would it be possible to do compile-time, rather than runtime checks on the values?
Alternately, an assert() on the default would be easier to debug rather than just failing silently, no?

rafaelx: I can replace the return on the default with an assert(), but I'm not sure how a compile-time check could look like on the enum.

> enum StringPrefs {CurrentBackup, CurrentFile, LastView, DefaultUsername,
    **TreeFont**, BackupPrefixValue, BackupDir, AltBrowser, ListColumns,
    ColumnWidths, DefaultAutotypeString, AltBrowserCmdLineParms,
    MainToolBarButtons, **PasswordFont**, TreeListSampleText, PswdSampleText,
    LastUsedKeyboard, VKeyboardFontName, VKSampleText, AltNotesEditor,
    LanguageFile, DefaultSymbols, **NotesFont**, NotesSampleText, AutotypeTaskDelays,
    **AddEditFont**, AddEditSampleText, AltNotesEditorCmdLineParms,
    NumStringPrefs};`

ronys: You're right - the only way a compile time check would work would be if the function were templatized, and I don't think that's worth the effort. A runtime assert() on the default case would be fine.
(Anther possibility would be to define a new enum type, "FontPrefs", and pass a parameter of that type - again, seems overkill to me...)

PR #579, opened 2019-09-04
Fixes #48  - Linux/Unix implementation of media.h using libmagic

### LINKED ISSUE #48: libmagic on Linux
/usr/bin/file uses libmagic, I believe it would be a better idea to use `magic_file()` (see `man 3 libmagic`) instead of executing `/usr/bin/file`.

cmaritan:
Fixes issue 
> libmagic on Linux  #48 

Also `CMakeLists.txt` have been updated for correct building.
I think that by using libmagic we have compatibility with every Linux/Unix system.
coretest are all ok after change.


ronys:
Looks like magic.h isn't part of the default distribution, which broke the CI. I'm guessing that libmagic-dev[el] should be added to .travis.yml (and Misc/setup-*-dev-env.sh).

Also, there's a Makefile in src/os/unix that should be updated (for those who won't/can't use cmake).

Finally, a unit test in src/test would be nice, if you're familiar with gtest.

cmaritan:
Ok gonna fix all asap, then document on gtest.
Thank you very Rony much for your feedback.
Bye.

rafaelx:
I would propose to add a check for the required packages/libraries to CMakeLists.txt.
Additionally, the new dependency (libmagic1 (>= 1:5.25)) could also be added to the packaging information to allow a smoothless installation.

PR #580, opened 2019-09-11
Update CMakeLists regarding 'Magic' library

rafaelx:
Related to #579

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -29,6 +29,8 @@ if (NOT WIN32)
 include(FindUnixCommands)
 endif (NOT WIN32)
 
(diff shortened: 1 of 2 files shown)


rafaelx:
RPM specific metadata might needs to be updated, too.

cmaritan:
> RPM specific metadata might needs to be updated, too.

Hello.
Adding "file-libs" to RPM metadata on CMakeLists seems to be enough.
BTW, not a real problem I think since dnf package depends on it and so file-libs is always installed!

I modified MakeFile in install/rpm with libmagic dependencies but found some other problems like:
`install: target '/home/claudio/rpmbuild/BUILDROOT/passwordsafe-1.08.2BETA-3.x86_64/usr/share/doc/passwordsafe' is not a directory
error: Bad exit status from /var/tmp/rpm-tmp.i9h1NE (%install)`

PR #582, opened 2019-09-16
libmagic - update libs README.LINUX.DEVELOPERS.md and CMakeList RPM metadata

cmaritan:
Removed libqrencode-devel on README.LINUX.DEVELOPERS.md because correct one, already included, is qrencode-devel.
Update CMakeLists for a typo on RPM metadata -> file-libs inclusion.


ronys:
Thanks!

PR #584, opened 2019-10-03
Fix "Encrytion" typo in msgid strings

bblough:
I noticed this typo while browsing through the code.  Since it's an internal string that isn't displayed anywhere, I don't suppose it matters all that much.  Regardless, here's a patch.


ronys:
Actually, it appears in the English version, and is used as the key for lookup in translations.
Thanks!

PR #592, opened 2019-10-13
Renaming of wx source code files

rafaelx:
- Naming convention (Pascal Case) as for Windows source code files
- Removed prefix 'Pws'
- Added suffix 'Dlg' if a dialog is implemented
- Added suffix 'Ctrl' if a new control is introduced
- Updated Makefile, CodeBlocks, CodeLite and Xcode project files


ronys:
Nice work!

PR #603, opened 2019-11-07
[wx] Activate system language if no language setting is specified in user preferences

rafaelx:


rafaelx:
If there is no language setting in the preferences file, in my case a wrong language (Konkani) was selected. Additionally, the wrong menu item was selected in the language menu.

PR #604, opened 2019-11-11
Fix Xcode build

sauravg:
* Add new & moves files to appropriate groups/targets
* Fix Include path for 'os'
* Include headers required for symbols used in sources

Only built one scheme on Xcode 8.2.1


ronys:
Thanks!

PR #607, opened 2019-12-06
[wx] Improve context menu handling at tree view

rafaelx:
Resolves reported issue [1217](https://sourceforge.net/p/passwordsafe/bugs/1217/).

### DIFF
--- a/src/ui/wxWidgets/TreeCtrl.cpp
+++ b/src/ui/wxWidgets/TreeCtrl.cpp
@@ -826,7 +826,7 @@ void TreeCtrl::OnMouseRightClick(wxMouseEvent& event)
 
   HitTest(mouseClickPosition, positionInfo);
 
-  if ((positionInfo | wxTREE_HITTEST_NOWHERE) == wxTREE_HITTEST_NOWHERE) {
+  if ((positionInfo & wxTREE_HITTEST_NOWHERE) == wxTREE_HITTEST_NOWHERE) {
     auto *parentWindow = dynamic_cast<PasswordSafeFrame*>(GetParent());
     wxASSERT(parentWindow != nullptr);
     /*
--- a/src/ui/wxWidgets/TreeCtrl.cpp
+++ b/src/ui/wxWidgets/TreeCtrl.cpp
@@ -21,22 +21,24 @@
 #include <wx/wx.h>
 #endif
 
+#ifdef __WXMSW__
+#include <wx/msw/msvcrt.h>
+#endif
+
 ////@begin includes
 #include <wx/imaglist.h>
 #include <wx/tokenzr.h>
 ////@end includes
 
-#include "TreeCtrl.h"
-#include "PasswordSafeFrame.h"
 #include "core/PWSprefs.h"
 #include "core/Command.h"
 
+#include "PasswordSafeFrame.h"
+#include "PWSafeApp.h"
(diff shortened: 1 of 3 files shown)


src/ui/wxWidgets/TreeCtrl.cpp
```diff
@@ -99,6 +101,24 @@ BEGIN_EVENT_TABLE( TreeCtrl, wxTreeCtrl )
...
+    In Linux environments context menus appear on Right-Down mouse click.
+    Which mouse click type (Right-Down/Right-Up) is the right one on a platform
+    is considered by 'EVT_CONTEXT_MENU(TreeCtrl::OnContextMenu)' which doesn't
+    work for wxTreeCtrl prior wx version 3.1.1. See also the following URL.
+    https://github.com/wxWidgets/wxWidgets/commit/caea08e6b2b9e4843e84c61abc879880a08634b0
+  */
+#if wxCHECK_VERSION(3, 1, 1)
+  EVT_CONTEXT_MENU(TreeCtrl::OnContextMenu)
+#else
+#ifdef __WINDOWS__
```
dskrvk: Isn't the whole wxWidgets part Linux-specific?

src/ui/wxWidgets/TreeCtrl.cpp
```diff
@@ -785,6 +819,46 @@ void TreeCtrl::OnKeyDown(wxTreeEvent& evt)
...
+}
+
+void TreeCtrl::OnMouseLeftClick(wxMouseEvent& event)
+{
+  wxPoint mouseClickPosition = event.GetPosition();
+  int positionInfo;
+
+  HitTest(mouseClickPosition, positionInfo);
+
+  if ((positionInfo | wxTREE_HITTEST_NOWHERE) == wxTREE_HITTEST_NOWHERE) {
```
dskrvk: I don't know much about the wxWidgets API, but if `wxTREE_HITTEST_NOWHERE` is a mask with a single bit set to 1, shouldn't the operator be `&`?

src/ui/wxWidgets/TreeCtrl.cpp
ronys: Actually, no.
- The wx code also runs on Mac
- It's also built and run under Windows as an exercise in portability.

src/ui/wxWidgets/TreeCtrl.cpp
ronys: Agree - certainly looks like this should be "&", not "|"

src/ui/wxWidgets/TreeCtrl.cpp
rafaelx: Would be great if someone could test the implementation with wxWidgets 3.1.1 and on platforms other than Linux.

src/ui/wxWidgets/TreeCtrl.cpp
rafaelx: Ow, good catch. - Fixed.

dskrvk: Fixed in OnMouseRightClick but not in OnMouseLeftClick.

rafaelx: Thanks for pointing this out. Fixed in #609.

PR #619, opened 2020-01-04
Fixing adding new groups and entires from treeview

akossomfai:
Starting from 06403ca63d93e7300cd2862cfe055af15ab5a186 is was not possible to add new groups or entires from the treeview in case of an empty database. The reason is that in this case there is neither root nor group selection so menu items will be dimmed. Fixing.


rafaelx:
When I was testing the changes of 06403ca I started always with a not empty database. So even after deleting of all items I was never faced with the situation of having no root element in the tree. :-/

This fix is very important for users that want to start with a new database.
A workaround for those users is to create the first entry in the list view.

ronys:
Thanks!

PR #621, opened 2020-01-06
Various fixes for compiler warnings

pm-kan:
Most of them are "cosmetic" and don't change logic, but cleanup compiler output and simplify catching warnings for new changes. But there are some changes too. Please check if I correctly understand desired behavior in src/ui/wxWidgets/MenuFileHandlers.cpp


src/core/Match.cpp
```diff
@@ -17,6 +17,12 @@
 
 #include <time.h>
 
+#ifdef _MSC_VER
+// suppress unknown attribute warning
+#pragma warning(disable : 5051)
```
ronys: Hm. On VS2015, this is warning 5030. Where did 5051 come from?
Also, this pragma should go into something more low-level, we don't want to reintroduce it whenever we use [[fallthrough]]...

src/core/PWPolicy.cpp
```diff
@@ -18,6 +18,12 @@
 #include "os/debug.h"
 
 #include <iomanip>
+
+#ifdef _MSC_VER
```
ronys: See previous comment...

src/core/PWScore.cpp
```diff
@@ -39,6 +39,12 @@
 #include <set>
 #include <iterator>
 
+#ifdef _MSC_VER
```
ronys: ditto...

src/core/PWSfile.cpp
```diff
@@ -22,6 +22,11 @@
 #include <errno.h>
 #include <limits>
 
+#ifdef _MSC_VER
```
ronys: ...

src/core/Util.cpp
```diff
@@ -38,6 +38,11 @@
 
 #include <errno.h>
 
+#ifdef _MSC_VER
```
ronys: ...

src/core/crypto/TwoFish.cpp
```diff
@@ -14,6 +14,11 @@
 #include "bitops.h"
 #include "../Util.h"
 
+#ifdef _MSC_VER
```
ronys: ...

src/ui/Windows/OptionsShortcuts.cpp
```diff
@@ -763,7 +763,7 @@ int COptionsShortcuts::CheckAppHotKey()
 
       CGeneralMsgBox gmb;
       CString cs_title(MAKEINTRESOURCE(IDS_SHORTCUT_WARNING));
-      gmb.MessageBox(cs_errmsg, cs_title, MB_OK | MB_ICONSTOP); gmb;
+      gmb.MessageBox(cs_errmsg, cs_title, MB_OK | MB_ICONSTOP);
```
ronys: !

src/ui/wxWidgets/MenuFileHandlers.cpp
```diff
@@ -334,6 +335,7 @@ int PasswordSafeFrame::Save(SaveType savetype /* = SaveType::INVALID*/)
               if (wxMessageBox(_("Unable to create intermediate backup.  Do you wish to save changes to your database without it?"),
                 _("Write Error"), wxYES_NO | wxICON_EXCLAMATION, this) == wxID_NO)
                 return PWScore::USER_CANCEL;
+              break; // continue save
```
ronys: Yes, this is correct. Nice catch!

src/ui/wxWidgets/MenuFileHandlers.cpp
```diff
@@ -345,10 +347,9 @@ int PasswordSafeFrame::Save(SaveType savetype /* = SaveType::INVALID*/)
                * SaveType::WTSLOGOFFEXIT
                * SaveType::FAILSAFESAVE
                */
-              break;
+               wxMessageBox(_("Unable to create intermediate backup."), _("Write Error"), wxOK|wxICON_ERROR, this);
```
ronys: Correct.

src/ui/wxWidgets/MenuFileHandlers.cpp
```diff
@@ -569,6 +570,8 @@ int PasswordSafeFrame::SaveIfChanged()
         // Make sure that file was successfully written
         if (rc != PWScore::SUCCESS)
           return PWScore::CANT_OPEN_FILE;
+        UpdateStatusBar();
```
ronys: Correct

src/core/PWSFilters.cpp
```diff
@@ -769,6 +775,7 @@ stringT PWSFilters::GetFilterDescription(const st_FilterRow &st_fldata)
         break;
       }
       // Note: purpose drop through to standard 'string' processing
+      [[fallthrough]];
```
rafaelx: Is this project already supporting C++17 features?

src/ui/wxWidgets/PasswordPolicyDlg.cpp
```diff
@@ -297,7 +298,7 @@ void PasswordPolicyDlg::CreateControls()
 
   m_pwpSymCtrl = new wxCheckBox( itemDialog1, ID_CHECKBOX6, _("Use symbols"), wxDefaultPosition, wxDefaultSize, 0 );
   m_pwpSymCtrl->SetValue(false);
-  m_pwpSymCtrl->Bind(wxEVT_MOTION, [&](wxMouseEvent& event) { m_pwpSymCtrl->SetToolTip(_("i.e., ., %, $, etc.")); });
+  m_pwpSymCtrl->Bind(wxEVT_MOTION, [&](wxMouseEvent& /*event*/) { m_pwpSymCtrl->SetToolTip(_("i.e., ., %, $, etc.")); });
```
rafaelx: What about WXUNUSED as it is used already in some other cases?

src/ui/wxWidgets/OptionsPropertySheetDlg.cpp
```diff
@@ -1247,7 +1247,7 @@ void OptionsPropertySheetDlg::OnBrowseLocationClick( wxCommandEvent& /* evt */ )
  * wxEVT_COMMAND_BUTTON_CLICKED event handler for ID_PWHISTAPPLY
  */
 
-void OptionsPropertySheetDlg::OnPWHistApply( wxCommandEvent& evt )
+void OptionsPropertySheetDlg::OnPWHistApply( wxCommandEvent& /*evt*/)
```
rafaelx: How about a uniform use of WXUNUSED?

src/ui/wxWidgets/AddEditPropSheetDlg.cpp
```diff
@@ -738,7 +738,7 @@ wxPanel* AddEditPropSheetDlg::CreatePasswordPolicyPanel()
   // Symbols Rules
   m_PasswordPolicyUseSymbolsCtrl = new wxCheckBox(panel, ID_CHECKBOX6, _("Use symbols"), wxDefaultPosition, wxDefaultSize, 0);
   m_PasswordPolicyUseSymbolsCtrl->SetValue(false);
-  m_PasswordPolicyUseSymbolsCtrl->Bind(wxEVT_MOTION, [&](wxMouseEvent &event) { m_PasswordPolicyUseSymbolsCtrl->SetToolTip(_("i.e., ., %, $, etc.")); });
+  m_PasswordPolicyUseSymbolsCtrl->Bind(wxEVT_MOTION, [&](wxMouseEvent &/*event*/) { m_PasswordPolicyUseSymbolsCtrl->SetToolTip(_("i.e., ., %, $, etc.")); });
```
rafaelx: Maybe WXUNUSED, instead?

src/core/Match.cpp
pm-kan: I've got warning C5051 on VS2019 Community Edition. Rechecked with Clang, and is also warn about usage of this attributes with `std=c++11` (but understand them), so I've commented these attributes and removed these pragmas. gcc understand "special comments" (https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Wimplicit-fallthrough) and `\\[[fallthrough]];` processed in `-Wimplicit-fallthrough=2` mode (clang don't use comments and will warn if `-Wimplicit-fallthrough` enabled). It's also possible to use special attributes for clang and gcc in c++11 mode, but I'm not sure, if we want to duplicate them for each compiler (another way is to add macro that will use correct attribute for each compiler).

src/core/PWSFilters.cpp
pm-kan: gcc 9 process them even in c++11 mode, but now I've commented them (see detail in the first comment)

src/ui/wxWidgets/PasswordPolicyDlg.cpp
pm-kan: `WXUNUSED` macro comment them too), but, for uniformity, changed commented names to macro (only in wx-related code).

pm-kan:
Added few more fixes.
Also found few things, that could be fixed later:
- In ` CItemData::Read` almost all return codes are enum values from `PWSfile`, but at https://github.com/pwsafe/pwsafe/blob/master/src/core/ItemData.cpp#L179 it returns negative value. If I not missed something, signed check used only at https://github.com/pwsafe/pwsafe/blob/master/src/core/PWSfileV4.cpp#L377 and maybe changed to some enum value too (and function's return type could be reduced to enum).
- Compiler warns about type redefinition at https://github.com/pwsafe/pwsafe/blob/master/src/core/crypto/bitops.h#L68. So it maybe better to define ulong32/64 only at typedefs.h or event remove custom integer `typedef`s and directly use intN_t (as we use c++11, it should be available on all platforms, https://en.cppreference.com/w/cpp/types/integer).
- It seems that static casts in LOAD32* macro in bitops.h should be `ulong32` instead of `unsigned long`.

ronys:
Checked it locally on MSVS2015 in various configurations, added another fix to shutup a warning that popped up in some of the build configs.

PR #627, opened 2020-01-16
Added info about Gentoo to README.LINUX.md

hansfn:
There has been an ebuild (pacakge) for Gentoo since 2016. It's about time it gets listed ;-)


hansfn:
You could consider having a Gentoo ebuild in this project making the Gentoo support self contained. I notice that you have a `install/slackware` directory. Let me know if that is interesting.

ronys:
Sure, I'd be glad to add integral Gentoo ebuild support.

PR #629, opened 2020-01-25
[wx] Change conversion from wxString and rearrange some headers

rafaelx:


src/ui/wxWidgets/MergeDlg.cpp
```diff
@@ -12,24 +12,29 @@
 
 #include <wx/wxprec.h>
 
+#ifdef __BORLANDC__
```
ronys: Where did this come from? I don't think Borland C has been around for years (RIP).

src/ui/wxWidgets/OpenFilePickerValidator.cpp
```diff
@@ -12,18 +12,23 @@
 
 #include <wx/wxprec.h>
 
+#ifdef __BORLANDC__
```
ronys: Ditto

src/ui/wxWidgets/QRCodeDlg.cpp
```diff
@@ -10,22 +10,24 @@
...
 
-#ifndef WX_PRECOMP
-#include "wx/wx.h"
-#endif
+// For compilers that support precompilation, includes "wx/wx.h".
+#include <wx/wxprec.h>
 
-#include "./QRCodeDlg.h"
-#include "./wxUtilities.h"
+#ifdef __BORLANDC__
```
ronys: ...

src/ui/wxWidgets/SystemTray.cpp
```diff
@@ -12,40 +12,45 @@
 
 #include <wx/wxprec.h>
 
+#ifdef __BORLANDC__
```
ronys: ...

src/ui/wxWidgets/ViewReportDlg.cpp
```diff
@@ -12,21 +12,24 @@
 
 #include <wx/wxprec.h>
 
+#ifdef __BORLANDC__
```
ronys: ...

rafaelx:
Basically, I just wanted to change some of the string conversions that I have introduced in the passed. They should base on what wxUtilities.h provides for that purpose. After adding wxUtilities.h to one of the implementation files I got compilation errors which is the reason for adding <wx/eventfilter.h> and <wx/dataobj.h> to wxUtilities.h. I was wondering why this didn't gave compilation errors before and searched for all appearances where wxUtilities.h is included and that is how I stumbled over the header include sections of some source files. In most of the wx source files there is this Borland specific preprocessor directive. I added it to other source files just for consistency. I guess it was originally added by the UI designer that was used to create the wx UIs. - Will remove it from all source files.

ronys:
Looks good - thanks!

PR #630, opened 2020-01-25
Update README.LINUX.DEVELOPERS.md

rafaelx:


seamusdemora:
Nice!... Is there a document like this for macos developers?

ronys:
> Nice!... Is there a document like this for macos developers?

You might want to look at https://github.com/pwsafe/pwsafe/blob/master/README.MAC.DEVELOPERS.md

Issue #625, opened 2020-01-08
Function void TreeCtrl::OnContextMenu(wxContextMenuEvent& event) declared but not defined


quirxmode:
This breaks the build under linux with wxWidgets 3.1.1. Actually, to get to that error you have to get to the linking step, where it will say something like `<artificial>:(.text.startup+0x3b1): undefined reference to `TreeCtrl::OnContextMenu(wxContextMenuEvent&)'`
But to get there you already need to fix code: the function `TreeCtrl::OnMouseRightClick` is not declared but defined in TreeCtrl if wxWidgets 3.1.1 is available, leading to a compilation error. The ifdef

`#if wxCHECK_VERSION(3, 1, 1)
  EVT_CONTEXT_MENU(TreeCtrl::OnContextMenu)
#else
#ifdef __WINDOWS__
  EVT_RIGHT_UP(TreeCtrl::OnMouseRightClick)
#else
  EVT_RIGHT_DOWN(TreeCtrl::OnMouseRightClick)
#endif
#endif // wxCHECK_VERSION(3, 1, 1)
`

on lines ~src/ui/wxWidgets/TreeCtrl.cpp:111 imho also needs to be repeated for the implementation, where OnMouseRightClick is implemented on line ~822, and `TreeCtrl::OnContextMenu(wxContextMenuEvent&)'` needs to be implemented.

rafaelx:
Could you please give the following branch a try?

https://github.com/rafaelx/pwsafe/tree/Issue625

If it builds successfully it would be interesting whether the context menu is properly working if a tree item is clicked but also if clicking into an empty area of the tree.

quirxmode:
Compiles, works, doesn't crash.
You get the same context menu when you click on an entry or in the empty space to the right of the entry. In both cases this highlights the entry.
If you click in empty space below you do not get a context menu.
I did not test every entry in the context menu, the copy password function does appear to copy the password of the selected entry though.

rafaelx:
Fixed by #635

PR #635, opened 2020-02-02
[wx] Fix TreeCtrl::OnContextMenu

rafaelx:


rafaelx:
Fixes #625.

src/ui/wxWidgets/PasswordSafeFrame.cpp
```diff
@@ -1467,7 +1467,7 @@ void PasswordSafeFrame::OnContextMenu(const CItemData* item)
         but is not allowed to and cannot be modified.
         Only groups from the users database are editable.
       */
-      if (!m_tree->IsRootSelected()) {
+      if (m_tree->IsGroupSelected()) {
```
dskrvk: The comment above probably needs updating (or removing).

src/ui/wxWidgets/TreeCtrl.cpp
```diff
@@ -993,6 +1028,20 @@ void TreeCtrl::RestoreGroupDisplayState()
...
 
+bool TreeCtrl::Show(bool show)
+{
+  /*
+    At least one tree element is needed to make the context menu work,
+    otherwise it won't show. TreeCtrl::AddGroup checks if a root item
+    is needed or an existing one is available.
+  */
+  if (IsEmpty()) {
+    AddRoot(wxT("root"));
```
dskrvk: The `Show` method looks like a simple predicate, so it's a bit counter-intuitive that it may modify the database. Ideally any modifying operation should ensure it preserves the root so we won't have to do it here.

PR #641, opened 2020-03-29
[wx] Adjust clipboard formats to support Clipboard copy into Firefox with Wayland backend

pm-kan:
Quick fix to support clipboard paste in FF with Wayland backend (should fix sf-bug 1500). 
Almost the same trick as for DnD support, but with other mime-type.

### DIFF
--- a/src/ui/wxWidgets/Clipboard.cpp
+++ b/src/ui/wxWidgets/Clipboard.cpp
@@ -10,6 +10,13 @@
  *
  */
 
+// For compilers that support precompilation, includes "wx/wx.h".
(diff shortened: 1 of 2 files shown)


ronys:
Thanks!

rafaelx:
Does it fix the reported issue [1500](https://sourceforge.net/p/passwordsafe/bugs/1500/)?

pm-kan:
Should fix. It works with my test setup (I've debugged this issue in Fedora Workstation 31 virtual machine updated at 28.03), but it 's better to check with bug author.

PR #647, opened 2020-04-15
[wx] Fix pre-selection in drop down boxes.

rafaelx:


ronys:
That simple !?!
Thanks for this and the cleanup in addEdit

PR #650, opened 2020-04-22
Update README.LINUX.md

blare-ski:
Updating the CLI installation command for Fedora.

### DIFF
--- a/README.LINUX.md
+++ b/README.LINUX.md
@@ -45,6 +45,7 @@ or
 2. Install it using yum:
    ```
    $ sudo yum install passwordsafe-*.rpm
+   $ sudo dnf install pwsafe
    ```
 
 ## Installation on Gentoo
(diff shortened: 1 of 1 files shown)


ronys:
Thanks! Is the dnf command in addition to or instead yum? Perhaps there should be an "or" between the two?

blare-ski:
You're welcome. Also, I don't know. The last update on baseurl.org was in 2011, DistroWatch doesn't track it any more, and in Fedora /bin/yum is a symlink to dnf-3, so... I guess I'd vote for 'instead of,' but I'm a newbie, so.

Karneades:
Yes, Fedora replaced yum with dnf a while back and should be the first choice.

Karneades:
For convenience, you could also add the link to the package in the Fedora section (https://apps.fedoraproject.org/packages/pwsafe).

Karneades:
And looking at pwsafe, this is not the correct passwordsafe ("pwsafe"): https://apps.fedoraproject.org/packages/pwsafe - @blare-ski could you clarify?

Karneades:
Similar topic now also in https://github.com/pwsafe/pwsafe/pull/651.

PR #657, opened 2020-05-17
[wx] Increase width of wxSpinCtrl controls for Gtk3

rafaelx:


dskrvk:
Could you attach screenshots?

ronys:
Interesting. When building against wx3-gtk2, there's no difference in the spinboxes.
When building against wx-gtk3, the "spinboxes" are better with the changes in this pull request, but still not very nice. See attached.

Note also that running the gtk3 version gives many warnings of the form:

> (pwsafe:18245): Gtk-CRITICAL **: 20:54:34.608: gtk_box_gadget_distribute: assertion 'size >= 0' failed in GtkSpinButton

when a window with spinboxes is rendered.

Master branch:
gtk3-master

This pull request applied:
gtk3-pr657

rafaelx:
These messages appear when a spinner (or probably also any other controls) cannot be fully displayed.
It seems that choosing a fixed size is generally not a good strategy, since on other systems the width may not be sufficient to fully display the control. On my Linux installation (Kubuntu 20.04) with the selected size, the messages are no longer displayed and the button on the far right of the control is no longer only partially visible.
I will change the fixed size to the default size, so it is the responsibility of the wx library to choose the appropriate width. The controls are then too wide for my feeling. The selected width allows the entry of approx. 10 digits, although typically only 2 or 3 are entered by the user.

rafaelx:
That's how it looks now with the default size.

PasswordPolicyDlg

AddEditDialog_Policy

Please pay also attention to the text entry field of the search tool. I set it to the fixed height of 30 pixels because with the default height the text entered was not completely visible.
SearchTool

rafaelx:
That's strange. I removed the fixed height for the search tool to take a screenshot showing what I was facing, and now the overall size is fine.
SearchTool2

rafaelx:
The previous screenshots were taken of pwsafe with wxGtk 3.1.3. With wxGtk 3.0.4 some letters are not completely visible. The last letters in text entry filed on the screenshot are "g" and "q".
SearchTool3

So I suggest keeping the fixed height of 30 pixels for the search tool. Unless the visual appearance is in any way inappropriate for other Linux distributions.

rafaelx:
The spinners are unnecessary wider on the policy related dialogs with wxGtk 3.0.4.
PasswordPolicyDlg_wxGtk304
AddEditDialog_Policy_wxGtk304

ronys:
The Search window definitely looks odd:
image
- The descenders for g and q are missing a pixel or two from the bottome
- The outlines for the text edit control are confusing

rafaelx:
This seems to be related to issue #526 and should be [fixed](https://github.com/wxWidgets/wxWidgets/commit/82caca68ba344d35a30a05a13e4aa658df867b52) in wxGtk 3.1.3.

PR #662, opened 2020-05-29
Gitpod: Workaround for https://github.com/gitpod-io/gitpod/issues/1565

Kreyren:
Concluded that gitpod using `sudo apt-get` results in a permission issue (https://github.com/gitpod-io/gitpod/issues/1565). This is a workaround for this issue.

File setup-deb-dev-env
- This file has extension `.sh`, but is using a shebang `#!/bin/bash` for bash and in previous commit was calling `sh` to call this script which might break on some environments i.e environments without shebang support compiled in kernel and systems without dynamic linking (not necesary linked to gitpod, but good practice)
- Was using `apt-get` without `-y` which would cause an infinite loop in gitpod's docker environment that doesn't have set `DEBIAN_FRONTEND=noninteractive` to workaround https://github.com/gitpod-io/gitpod/issues/1171

Shellcheck (https://github.com/koalaman/shellcheck) has been recommended to ensure code quality of shell scripts

This also makes it to not use `setup-deb-dev-env.sh` let me know if you want to use this file in gitpod (seems to me as bad idea since it would be harder to cache), but i am willing to make changes to allow this file to be used.

Signed-off-by: Jacob Hrbek <kreyren@member.fsf.org>

### DIFF
--- a/.gitpod.Dockerfile
+++ b/.gitpod.Dockerfile
@@ -12,29 +12,7 @@ RUN echo 'debconf debconf/frontend select Noninteractive' | sudo debconf-set-sel
 USER root
 RUN true \
   && apt-get -q update \
-  && apt-get install -yq \
-    cmake-gui \
-    cmake \
-    fakeroot \
-    g++ \
-    gettext \
-    git \
-    libgtest-dev \
-    libcurl4-openssl-dev \
-    libqrencode-dev  \
-    libssl-dev \
-    libuuid1 \
-    ibwxgtk3.0-dev \
-    libxerces-c-dev \
-    libxt-dev \
-    libxtst-dev \
-    libykpers-1-dev \
-    libyubikey-dev \
-    make \
-    pkg-config \
-    uuid-dev \
-    zip \
-    libmagic-dev \
+  && sh ./Misc/setup-deb-dev-env.sh
   && apt-get autoremove -yq \
   && rm -rf /var/lib/apt/lists/*
 
--- a/.gitpod.Dockerfile
+++ b/.gitpod.Dockerfile
@@ -9,6 +9,35 @@ RUN echo 'debconf debconf/frontend select Noninteractive' | sudo debconf-set-sel
 # RUN sudo apt-get -q update && #     sudo apt-get install -yq bastet && #     sudo rm -rf /var/lib/apt/lists/*
 #
 # More information: https://www.gitpod.io/docs/config-docker/
-RUN sudo apt-get -q update && sudo sh ./Misc/setup-deb-dev-env.sh && sudo apt install -yq cmake-gui && sudo rm -rf /var/lib/apt/lists/*
+USER root
+RUN true \
+  && apt-get -q update \
+  && apt-get install -yq \
+    cmake-gui \
+    cmake \
+    fakeroot \
+    g++ \
+    gettext \
+    git \
+    libgtest-dev \
+    libcurl4-openssl-dev \
+    libqrencode-dev  \
+    libssl-dev \
+    libuuid1 \
+    ibwxgtk3.0-dev \
+    libxerces-c-dev \
+    libxt-dev \
+    libxtst-dev \
+    libykpers-1-dev \
+    libyubikey-dev \
+    make \
+    pkg-config \
+    uuid-dev \
+    zip \
+    libmagic-dev \
+  && apt-get autoremove -yq \
+  && rm -rf /var/lib/apt/lists/*
+
 # Set debconf back to normal.
+USER gitpod
 RUN echo 'debconf debconf/frontend select Dialog' | sudo debconf-set-selections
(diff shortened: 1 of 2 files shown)


ronys:
Thanks, but this misses what I was trying to achieve: Avoiding two separate lists of the same dependencies. Experience shows that there's a > 50% chance of forgetting to update one.

ronys:
> @ronys What is the preferred way to maintain the dependencies then? Using a script that has the logic? Or should the script read from a readme or some other file?

Good question. Currently I have scripts for deb and rpm dev env setups. It would be nice to have a unified script that reads the correct list by autodetecting the distro type.
The dependencies could be in an external file or files, but they may need to have some conditional logic applied. For example, Ubuntu switched from libwxgtk3.0-dev to libwxgtk3.0-gtk3-dev in Ubuntu 20. This should be reflected in the setup script.

Any suggestion?

Kreyren:
heads-up: i am working on a solution in https://github.com/Kreyren/pwsafe/pull/1

ronys:
Thanks for the heads-up. I was planning on working on this myself. Waiting patiently for your pull request.

If it makes any difference, what I'd *really* like is the ability to use gitpod to create workspaces based on different ditros, not only Ubuntu. Any chance of this being possible?

Kreyren:
> Any chance of this being possible?

Yes you can use any dockerimage on gitpod as long as it passes security gitpod-layers :p

Choose your preferred and let me know if you encounter any issues i do my best to solve them for you, but praying to @jankeromnes to not be lazy and slow may be required to merge the changes.

Btw. i've noticed the replies just now >.<

Kreyren:
@ronys btw. if you are willing to fix the compatibility issues then i leave it to you since i am currently struggling with that~

I've made a sourcehut images for you that you can use for testing though which should make it easy ^-^ The changes seems to be minor for someone who knows the code

Kreyren:
> If it makes any difference, what I'd really like is the ability to use gitpod to create workspaces based on different ditros, not only Ubuntu. Any chance of this being possible? @ronys 

If it helps there is my debian Dockerfile for gitpod ^-^

https://github.com/RXT0112/Zernit/blob/master/gitpod/debian.Dockerfile

ronys:
> > Any chance of this being possible?
> 
> Yes you can use any dockerimage on gitpod as long as it passes security gitpod-layers :p
> 
> Choose your preferred and let me know if you encounter any issues i do my best to solve them for you, but praying to @jankeromnes to not be lazy and slow may be required to merge the changes if any needed.
> 
> Btw. i've noticed the reply just now >.<

Well, my default dev env is based on Debian 10, but when I release a version of PasswordSafe, I build (and test) it on (at least) Debian, Ubuntu (last 2 LTSs) and Fedora (last 2 releases), and others upon request (e.g., Mint).
Ideally, there would a way to specify the distro when creating a gitpod workspace. Perhaps something like a ?param=value that I can add to the creation URL.
For example: https://gitpod.io/#https://githob.com/pwsafe/pwsafe?distro=fedora32

Or am I asking for too much?

Kreyren:
> Ideally, there would a way to specify the distro when creating a gitpod workspace. 

I would also like this feature on gitpod, currently tracked on https://github.com/gitpod-io/gitpod/issues/1263 i try to force @nisarhassan12 with the implementation ^-^

> Well, my default dev env is based on Debian 10, but when I release a version of PasswordSafe, I build (and test) it on (at least) Debian, Ubuntu (last 2 LTSs) and Fedora (last 2 releases), and others upon request (e.g., Mint).

Usually you would use a docker assuming Linux Kernel used which is not working on Gitpod https://github.com/gitpod-io/gitpod/issues/52 due to the https://github.com/gitpod-io/gitpod/issues/39 which to my understanding is currently blocked for security reasons since Gitpod's workspaces are running on a docker. I've proposed a solution to @meysholdt and @svenefftinge in https://typefox.slack.com/archives/CD3892V6Z/p1590798698000400?thread_ts=1590371237.000800&cid=CD3892V6Z (Accessible only to TypeFox and major contributors) which attempts to use VM-based containers which would also fix https://github.com/gitpod-io/gitpod/issues/1251, https://github.com/gitpod-io/gitpod/issues/1265 and other related issues.
Currently waiting for them to process it and if anyone has experience with VM-based containers that i personally would appreciate their input to this as well since based on my research it's a very good upgrade for gitpod.

Currently i would recommend you using [sr.ht](https://sr.ht/) which allows you to test the source code in a VM provided by ddevault for free(?) as proposed in https://github.com/pwsafe/pwsafe/pull/663

> Perhaps something like a ?param=value that I can add to the creation URL.
> For example: gitpod.io/#https://githob.com/pwsafe/pwsafe?distro=fedora32

I like that idea ^-^

> Or am I asking for too much?

Rather too little! :p
Gitpod is supposed to be a tool to speed up the development so in my opinion these issues should be resolved for it to be considered as a viable option for cross-platform development.

PR #667, opened 2020-06-13
[wx] Change options dialog to listbook style

rafaelx:


ronys:
Nice, but a couple of small changes would make it even better:
1. Some indication as to what the current page is. Something as simple as changing the title bar from "Options" to, for example, "Options - Security" would be fine.
2. The "System" tab is a bit "confused" when displaying the "no system tray support" warning (see below). Making the dialog a bit wider (by ~10 pixels) should fix this.
image

rafaelx:
In the current implementation the options dialog is derived from 'wxPropertySheetDialog' which does not provide any events that indicate a page / tab switch. It would be necessary to use another base class that supports this.
In general, I would like to introduce wxAUI to be able to make use of some advanced UI features. This would also include 'wxAuiNotebook' which could act perfectly as base class for the options dialog. However, this possibly will result in some more work and may take some more time to implement.
One option could be to leave this pull request open until the switch to wxAUI for the options dialog is complete. Another option could be to merge the pull request version into master without any indicator pointing to the active tab, which currently also do not exist in the version on master. The 'active tab indicator' could be added later.

ronys:
I've chosen the second option - thanks!

PR #668, opened 2020-06-15
[wx] Refactor 'CreateControls' of 'OptionsPropertySheetDlg'

rafaelx:
Same as already done for 'AddEditPropSheetDlg', which makes navigation in the source code easier.


ronys:
Definitely a step forward, but I think we can do even better if we get the relevant control via ID, instead of keeping pointers to all of them in the object.

PR #675, opened 2020-07-17
Generate macOS builds using Github actions

sauravg:
This PR would probably generate a build using Github actions. There would be an Artifact link under the Checks tab, which would let you download a dmg. Users need to drag pwsafe.app from the dmg to their Applications folder.

Once merged, every subsequent PR would probably generate such a build.

However, this is a very minimal effort. The user would first need to install wxWidgets on his Mac by running "brew install wxWidgets".

This PR has another change that has been lying around in my fork: 883f38a: Enable QR code feature on Mac. It doesn't add any new dependencies. No reason to turn off the feature.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,63 @@
+# This is a basic workflow to help you get started with Actions
+
+name: CI
+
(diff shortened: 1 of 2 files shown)


ronys:
Thanks!

Issue #674, opened 2020-07-16
Yubikey button not displayed v1.10/1.11 on Fedora 32


antigenx:
The builds for Fedora 32 do not show the yubikey button when i insert my yubikey.
Instead the console fills up with the following messages:

(pwsafe:101752): Gtk-WARNING **: 11:13:15.007: Negative content height -9 (allocation 1, extents 5x5) while allocating gadget (node button, owner GtkButton)

(pwsafe:101752): Gtk-WARNING **: 11:13:15.502: Negative content width -19 (allocation 1, extents 10x10) while allocating gadget (node button, owner GtkButton)

antigenx:
from syslog:

insert event:
Jul 16 11:27:31 xx kernel: usb 2-2: new full-speed USB device number 15 using xhci_hcd
Jul 16 11:27:31 xx kernel: usb 2-2: New USB device found, idVendor=1050, idProduct=0116, bcdDevice= 3.37
Jul 16 11:27:31 xx kernel: usb 2-2: New USB device strings: Mfr=1, Product=2, SerialNumber=0
Jul 16 11:27:31 xx kernel: usb 2-2: Product: Yubikey NEO OTP+U2F+CCID
Jul 16 11:27:31 xx kernel: usb 2-2: Manufacturer: Yubico
Jul 16 11:27:31 xx kernel: input: Yubico Yubikey NEO OTP+U2F+CCID as /devices/pci0000:00/0000:00:14.0/usb2/2-2/2-2:1.0/0003:1050:0116.02D2/input/input745
Jul 16 11:27:31 xx kernel: hid-generic 0003:1050:0116.02D2: input,hidraw3: USB HID v1.10 Keyboard [Yubico Yubikey NEO OTP+U2F+CCID] on usb-0000:00:14.0-2/input0
Jul 16 11:27:31 xx kernel: hid-generic 0003:1050:0116.02D3: hiddev97,hidraw4: USB HID v1.10 Device [Yubico Yubikey NEO OTP+U2F+CCID] on usb-0000:00:14.0-2/input1
Jul 16 11:27:31 xx mtp-probe[105884]: checking bus 2, device 15: "/sys/devices/pci0000:00/0000:00:14.0/usb2/2-2"
Jul 16 11:27:31 xx mtp-probe[105884]: bus: 2, device: 15 was not an MTP device
[...]

PR #683, opened 2020-08-08
[wx] SafeCombinationEntryDlg: Resolve issue on Fedora of not appearing buttons

rafaelx:
Could someone please check if this solves issue #674 ?


ronys:
Confirmed fix on Fedora32 VM.
Thanks!

PR #685, opened 2020-08-30
[wx] Some improvements to the handling of attachments.

rafaelx:


ronys:
Delaying merging this in until 3.53 is released, since there are changes in the common core library.

PR #689, opened 2020-09-11
SF1506

rafaelx:
Resolves [1506](https://sourceforge.net/p/passwordsafe/bugs/1506/).

### DIFF
--- a/src/os/unix/file.cpp
+++ b/src/os/unix/file.cpp
@@ -310,6 +310,11 @@ bool pws_os::LockFile(const stringT &filename, stringT &locker,
         StringXStream lockerStream;
         if (PWSUtil::loadFile(lock_filename.c_str(), lockerStream)) {
           locker = stringx2std(lockerStream.str());
+
+          if (!PWSUtil::HasValidLockerData(locker)) {
+            locker = _T("");
+            Format(locker, IDSC_INVALIDLOCKER, lock_filename.c_str());
+          }
         }
         else {
           LoadAString(locker, IDSC_CANTREADLOCKER);
--- a/src/os/unix/file.cpp
+++ b/src/os/unix/file.cpp
@@ -221,18 +221,6 @@ static stringT GetLockFileName(const stringT &filename)
  */
 void pws_os::TryUnlockFile(const stringT &filename, HANDLE &lockFileHandle)
 {
-  // Provides characters from the beginning of str up to given delimiter
-  // and removes them finally from str.
-  const auto getStringToken = [](stringT &str, const stringT &delimiter) -> stringT {
-    stringT token;
-    size_t pos = str.find(delimiter);
-    if (pos != std::string::npos) {
-      token = str.substr(0, pos);
-      str.erase(0, pos + delimiter.length());
-    }
-    return token;
-  };
-
   const stringT lockFilename = GetLockFileName(filename);
 
   size_t mbsSize = wcstombs(nullptr, lockFilename.c_str(), lockFilename.length()) + 1;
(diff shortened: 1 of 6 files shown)


src/os/unix/file.cpp
```diff
@@ -310,6 +291,11 @@ bool pws_os::LockFile(const stringT &filename, stringT &locker,
         StringXStream lockerStream;
         if (PWSUtil::loadFile(lock_filename.c_str(), lockerStream)) {
           locker = stringx2std(lockerStream.str());
+
+          if (!PWSUtil::HasValidLockerData(locker)) {
+            locker = _T("");
+            Format(locker, IDSC_INVALIDLOCKER, lock_filename.c_str());
```
rafaelx, line 297: When the application is started from the command line it gets stuck when 'GetStringBufSize' is executed within 'Format'.
This behavior is not shown when pwsafe is started by the IDE.

src/ui/wxWidgets/SafeCombinationEntryDlg.cpp
```diff
@@ -373,8 +373,16 @@ void SafeCombinationEntryDlg::ProcessPhrase()
...
-      errmess += locker.c_str();
+      errmess = _("Could not lock file, opening read-only.\n");
+
+      if (PWSUtil::HasValidLockerData(locker)) {
+        errmess += _("Locked by ");
+        errmess += locker.c_str();
+      }
+      else {
+        errmess += locker.c_str();
+      }
```
rafaelx, line 384: The code could be changed as follows.

```
      if (PWSUtil::HasValidLockerData(locker)) {
        errmess += _("Locked by ");
      }

      errmess += locker.c_str();
```

src/os/unix/file.cpp
rafaelx: Resolved by updating 'core.rc2' with the newly introduced string.

src/ui/wxWidgets/SafeCombinationEntryDlg.cpp
rafaelx: Changed in new commit.

src/ui/wxWidgets/SafeCombinationEntryDlg.cpp
```diff
@@ -373,8 +373,14 @@ void SafeCombinationEntryDlg::ProcessPhrase()
     const stringT fname(m_filename.c_str());
     stringT locker(L"");
     if (!m_readOnly && !m_core.LockFile(fname, locker)) {
-      errmess = _("Could not lock file, opening read-only\nLocked by ");
+      errmess = _("Could not lock file, opening read-only.\n");
+
+      if (PWSUtil::HasValidLockerData(locker)) {
+        errmess += _("Locked by ");
```
dskrvk, line 379: Minor: line 382 could be combined with this one to append both strings in one go.

dskrvk (approved):

ronys:
Note to self: apply the refactoring to Windows-specific code.

src/ui/wxWidgets/SafeCombinationEntryDlg.cpp
rafaelx: The reason for this split is that 'locker' can provide different information.

Usually, if a lock file has been detected the resulting error message in the message box would appear as follows.

> Could not lock file, opening read-only.
> Locked by \<user@machine:nnnnnnnn>

If an error occured during the processing of the lock file 'locker' would refer to an corresponding error message.
This would lead to the following result.

> Could not lock file, opening read-only.
> Locked by \<error message>

With the current implementation it should be something as the following.

> Could not lock file, opening read-only.
> \<specific error message or reason for the read only mode>

I would have liked to leave the original string (line 376) untouched so as not to invalidate the translations.

dskrvk: Makes sense, I didn't consider the translations.

PR #690, opened 2020-09-21
Build & cache wxWidgets locally (#3)

sauravg:
Removed all dependencies on homebrew. Now we build wxWidgets ourselves and link to it statically, so there are no non-Apple runtime dependencies.

Builds produced by the new workflow should run on macOS 10.9+, though I've only tested on 10.11 and 10.15. I don't have the other versions of macOS but I'm reasonably certain that it'd work.

wxWidgets takes about 30 minutes to build, so we cache the build outputs. This PR should build wxWidgets and cache it so the subsequent builds would be faster (nearly 4 minutes).

Sometimes, the build cache is not used:

1. If the PR is from a branch A to branch B, but A didn't branch off B
2. If the cumulative build cache size exceeds 5 GB. Each build is ~ 48mb
3. If you change the wxWidgets version, macOS version being built for, or the Xcode version
4. The cache being referenced wasn't accessed in the last 7 days
5. Build cache was created by/for another user

This is just FYI, in case some PR takes too long to build.

Lastly, one of the build tool dependencies is gone: auriamg/macdylibbundler

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -22,23 +22,52 @@ jobs:
     # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
     - uses: actions/checkout@v2
 
+    - name: Set build parameters
+      # If you need to rebuild wxWidgets for some reason, say you changed the build params, just increment FORCE_REBUILD_WX
+      run: |
+        echo "::set-env name=WX_VERS::3.0.5"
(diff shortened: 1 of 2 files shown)


ronys:
Cool. I've learned a lot through this commit - thanks!

seamusdemora:
I am glad to see the requirement for `homebrew` has been dispensed with. Given you're using wxWidgets, static links strike me as a far better approach for all.

PR #692, opened 2020-10-03
Add 'Show Combination' to dialogs

rafaelx:
Since I don't have a YubiKey, I was not able to verify how the layout changes would affect the YubiKey-related controls.

### DIFF
--- a/src/ui/wxWidgets/CompareDlg.cpp
+++ b/src/ui/wxWidgets/CompareDlg.cpp
@@ -152,6 +152,7 @@ void CompareDlg::CreateControls()
 
   m_dbSelectionPane->Expand();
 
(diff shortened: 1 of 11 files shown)


ronys (changes requested):
Nice, but some more work is needed to complete this:

1. Export dialog looks bad - lower row of buttons are too close to the edge (see attached). I think it would be better to put the "Show Combination" checkbox on the same row as the Yubikey button and text.
2. The checkbox should also be added to the merge/sync/compare dialogs, as these also prompt for the master password.
3. In "Change Safe Combination", the "confirm" control should be disabled and ignored if the "Show Combination" checkbox is set.
wx-export

ronys (approved):

PR #693, opened 2020-10-03
Only run xdg-* functions in deb postint, prerm if the requried dirs exist

bblough:
Some distributions (debian, arch, maybe others) do not always create
the system-level /usr{,/local}/share/destop-directories paths which are
required for xdg-desktop-menu and similar programs to work correctly.

In addition, the prerm script uses the -e bash option (exit on non-zero
status).  However, the postinst script does not.

This can lead to the situation where the deb package can be installed,
but cannot be uninstalled. (upgrading still works)

This commit adds checks to the postinst and prerm scripts to only run the
xdg programs if at least one of the required directories exists.

This commit also adds the -e bash option to the postinst script, for
consistency.

### DIFF
--- a/install/deb/postinst
+++ b/install/deb/postinst
@@ -1,7 +1,10 @@
-#!/bin/bash
-xdg-icon-resource install --novendor --size 48 --mode system \
-                  /usr/share/icons/hicolor/48x48/apps/pwsafe.png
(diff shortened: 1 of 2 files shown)


ronys:
Thanks - reminded me that these files are missing in deb packages created via cmake/cpack.

PR #694, opened 2020-10-10
[wx] Add handling of Yubikey to Merge and Compare dialog

rafaelx:


src/ui/wxWidgets/DbSelectionPanel.cpp
```diff
@@ -137,3 +169,39 @@ void DbSelectionPanel::OnFilePicked(wxFileDirPickerEvent& WXUNUSED(event))
...
+      m_combination = response;
+
+      /* Compare Dialog
+      GetParent()->GetEventHandler()->AddPendingEvent(wxCommandEvent(wxEVT_BUTTON, ID_BTN_COMPARE));
+      */
+
+      /* Merge Dialog
+      GetParent()->GetEventHandler()->AddPendingEvent(wxCommandEvent(wxEVT_BUTTON, wxID_OK));
+      */
+
```
rafaelx, line 195: DbSelectionPanel is used as an embedded control in the Merge and Compare dialog. I'm not sure what should happen after YubiMixin::PerformChallengeResponse in DbSelectionPanel::OnYubibtnClick was successful in the case of the merge or in the comparison dialog. Should the "OK" button click event be triggered in the Merge dialog and the "Compare" button click event in the  Compare dialog?

ronys: Semantically, OnYubibtnClick() should set the m_combination field and simulate the OK / Compare button, so I think this part is OK.
Only problem I see is that the CheckPasskey() in DoValidation() should check the m_combination value set in line 106 above, and *not* what's xferred from the control in the call to wxWindow::Validate() in DoValidation.

ronys (changes requested):

ronys (changes requested):
Looks good, but I've found another problem:
PasswordSafe detects the presence of YubiKey and shows the green button, as it should.
When the YubiKey is removed (after being detected), the green button turns grey, and the text *should* change from "Click the button labeled 'YubiKey' in green" to "Please insert your YubiKey".
This is done by a timer event that triggers a check for the YubiKey.
In the Merge/Sync/Compare dialogs, the button changes from green to gray, but the text doesn't change accordingly. This is confusing.

PR #702, opened 2020-11-15
Update macOS building with several changes

KaiZin2018:
Complete Xcode and deal with wxWidget version 3.1.4. Remove all warnings and replace deprecated functions. Dark mode icons corrected to have no white background. Dark mode color set to macOS example. New setting to remove menu separators, as menu is very long on macOS. Re-open/unlock menu dialog deals with read-only flag.


KaiZin2018:
Update macOS Xcode building with Catalina.

KaiZin2018:
Please consider changes for macOS building

KaiZin2018:
WxWidget 3.1.4 version has to be used to compile with this version.

src/ui/wxWidgets/PasswordSafeFrame.cpp
```diff
@@ -868,6 +877,30 @@ void PasswordSafeFrame::RefreshToolbarButtons()
       }
     }
   }
+  tb->Realize();
+}
+
+void PasswordSafeFrame::ReCreateMainToolbarSepartor(bool bInsert)
```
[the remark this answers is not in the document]
KaiZin2018: Thank you for your hints. I followed them.

KaiZin2018:
Check build is failing here, but with my Mac it is running well.

src/core/PWScore.cpp
```diff
@@ -731,7 +731,7 @@ struct ExportRecordWriter {
 
 private:
   PWSfile *m_pout;
-  PWScore *m_pcore;
+  /*PWScore *m_pcore; // variable not used */
```
rafaelx, line 734: How about removing unused code?

src/ui/wxWidgets/SafeCombinationPromptDlg.cpp
```diff
@@ -172,15 +175,28 @@ void SafeCombinationPromptDlg::CreateControls()
...
-  flexGridSizer->AddStretchSpacer(0);
+// Show Combination is moved down to right side of new field "Open as read-only"
   auto showCombinationCheckBox = new wxCheckBox(this, wxID_ANY, _("Show Combination"), wxDefaultPosition, wxDefaultSize, 0 );
   showCombinationCheckBox->SetValue(false);
   showCombinationCheckBox->Bind(wxEVT_CHECKBOX, [&](wxCommandEvent& event) {m_scctrl->SecureTextfield(!event.IsChecked());});
-  flexGridSizer->Add(showCombinationCheckBox, 1, wxALL|wxEXPAND, 5);
 
   itemBoxSizer5->Add(flexGridSizer, 0, wxEXPAND|wxALL, 5);
 
+  auto *itemBoxSizer14 = new wxBoxSizer(wxHORIZONTAL);
```
rafaelx, line 185: In similar dialog boxes that require a password to be entered, such checkboxes are located below the password text input field and are left-justified to it. It would be nice if the application could offer the same visual appearance to the user.

src/ui/wxWidgets/SafeCombinationPromptDlg.cpp
```diff
@@ -347,6 +363,33 @@ void SafeCombinationPromptDlg::OnExitClick(wxCommandEvent& WXUNUSED(evt))
...
+ * wxEVT_COMMAND_CHECKBOX_CLICKED event handler for ID_READONLY
+ */
+
+void SafeCombinationPromptDlg::OnReadonlyClick( wxCommandEvent& event )
+{
+  m_readOnly = event.IsChecked();
+}
+
+void SafeCombinationPromptDlg::UpdateReadOnlyCheckbox(wxCheckBox *checkBox)
+{
```
rafaelx, line 376: The checkbox object could also be localized via its ID and the function thus made parameterless.

src/ui/wxWidgets/PWSafeApp.cpp
```diff
@@ -271,7 +271,7 @@ bool PWSafeApp::OnInit()
   // Get the locale environment variable 'LC_CTYPE' specified by the environment
   // For instance, the behavior of function 'wcstombs' depends on the LC_CTYPE 
   // category of the selected C locale.
-  setlocale(LC_CTYPE, "");
+  setlocale(LC_CTYPE, "en_US.UTF-8");
```
rafaelx, line 274: What is the advantage over the previous version?

src/ui/wxWidgets/CreateShortcutDlg.cpp
```diff
@@ -188,17 +188,17 @@ void CreateShortcutDlg::CreateControls()
 
   auto StaticTextShortcutGroup = new wxStaticText(this, wxID_ANY, _("Group:"), wxDefaultPosition, wxDefaultSize, 0, _T("wxID_ANY"));
   FlexGridSizer1->Add(StaticTextShortcutGroup, 1, wxALL|wxALIGN_RIGHT|wxALIGN_CENTER_VERTICAL, 5);
-  m_ComboBoxShortcutGroup = new wxComboBox(this, ID_COMBOBOX1, wxEmptyString, wxDefaultPosition, wxDefaultSize, 0, nullptr, wxCB_DROPDOWN, wxDefaultValidator, _T("ID_COMBOBOX1"));
+  m_ComboBoxShortcutGroup = new wxComboBox(this, static_cast<wxWindowID>(ID_COMBOBOX1), wxEmptyString, wxDefaultPosition, wxDefaultSize, 0, nullptr, wxCB_DROPDOWN, wxDefaultValidator, _T("ID_COMBOBOX1"));
```
rafaelx, line 191: Is it really necessary to have that many casts spread across the code? In some cases, wouldn't it be an option to adjust the data type instead?

src/core/PWScore.cpp
KaiZin2018: I had not been sure for unix or windows compilation, so stay with this as long no negative feedback is given.

src/ui/wxWidgets/SafeCombinationPromptDlg.cpp
KaiZin2018: Look and feel might be better with the tag fields aligned with the input field. I'm not sure about the needed size of the text of the two check buttons when translated, therefore moved the text left aligned to the left most place and not starting below the input field. This is a design decision I could correct, when wanted.

src/ui/wxWidgets/SafeCombinationPromptDlg.cpp
KaiZin2018: This is not needed, as the function is directly called only one time from SafeCombinationPromptDlg::CreateControls() where the reference is known and must search by identifier. The name might be misleading, but is very much the same functionality.

src/ui/wxWidgets/PWSafeApp.cpp
KaiZin2018: With my macOS Catalina the conversion from wxString (wchar_t) to UTF-8 coding is not running well as long as the setlocale function is not having the dedicated initialization. I do not know why, but this is the only chance that UTF-8 coding runs well.

src/ui/wxWidgets/CreateShortcutDlg.cpp
KaiZin2018: Yes, the various cast are introduced to come away from compiler warnings about implicit conversions.

KaiZin2018:
Don't know why Mac building fails on the server - on my iMac it run's well.

src/ui/wxWidgets/GridShortcutsValidator.cpp
```diff
@@ -544,7 +537,7 @@ void PWSMenuShortcuts::RemoveShortcutAt(size_t idx)
 void PWSMenuShortcuts::SetShorcutsGridEventHandlers(wxGrid* grid, wxButton* resetAllButton)
 {
   m_shortcutGridStatus.resize(m_midata.size());
-  std::transform(m_midata.begin(), m_midata.end(), m_shortcutGridStatus.begin(), std::mem_fun_ref(&MenuItemData::GetStatus));
+  std::transform(m_midata.begin(), m_midata.end(), m_shortcutGridStatus.begin(), [](MenuItemData& a){return a.GetStatus();} /*std::mem_fun_ref(&MenuItemData::GetStatus)*/);
```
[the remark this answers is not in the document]
KaiZin2018: Removed

src/ui/wxWidgets/AddEditPropSheetDlg.cpp
```diff
@@ -1414,11 +1414,11 @@ static struct {short pv; wxString name;}
   // - Adds " (default)" to default string
   // - Selects current value
   for (size_t i = 0; i < sizeof(dcaMapping)/sizeof(dcaMapping[0]); i++) {
-    pcbox->SetClientData(i, reinterpret_cast<void *>(dcaMapping[i].pv));
+    pcbox->SetClientData((unsigned int) i, reinterpret_cast<void *>(dcaMapping[i].pv));
     if (dcaMapping[i].pv == defDCA) {
       wxString dv = dcaMapping[i].name;
       dv += wxT(" ("); dv += _("default"); dv += wxT(")");
-      pcbox->SetString(i, dv);
+      pcbox->SetString((unsigned int) i, dv);
```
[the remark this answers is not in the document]
KaiZin2018: Used static cast now

src/ui/wxWidgets/AddEditPropSheetDlg.cpp
```diff
@@ -1414,11 +1414,11 @@ static struct {short pv; wxString name;}
   // - Adds " (default)" to default string
   // - Selects current value
   for (size_t i = 0; i < sizeof(dcaMapping)/sizeof(dcaMapping[0]); i++) {
-    pcbox->SetClientData(i, reinterpret_cast<void *>(dcaMapping[i].pv));
+    pcbox->SetClientData((unsigned int) i, reinterpret_cast<void *>(dcaMapping[i].pv));
```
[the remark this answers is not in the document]
KaiZin2018: Use static cast now

src/ui/wxWidgets/CreateShortcutDlg.cpp
dskrvk: I think in this particular case the control ID variables can be declared with type `wxWindowID`, unless that would require casting them to `int` elsewhere?

KaiZin2018: Declaration changed and static cast removed

src/ui/wxWidgets/SafeCombinationPromptDlg.cpp
```diff
@@ -175,24 +175,27 @@ void SafeCombinationPromptDlg::CreateControls()
...
 
+// Show Combination is moved down to right side of new field "Open as read-only"
+  auto showCombinationCheckBox = new wxCheckBox(this, wxID_ANY, _("Show Combination"), wxDefaultPosition, wxDefaultSize, 0 );
+  showCombinationCheckBox->SetValue(false);
+  showCombinationCheckBox->Bind(wxEVT_CHECKBOX, [&](wxCommandEvent& event) {m_scctrl->SecureTextfield(!event.IsChecked());});
+
   itemBoxSizer14->Add(itemCheckBox15, 1, wxALIGN_LEFT|wxALL, 5);
   itemBoxSizer14->AddSpacer(1);
   itemBoxSizer14->Add(showCombinationCheckBox, 1, wxALIGN_LEFT|wxALL, 5);
   itemBoxSizer14->AddStretchSpacer();
```
rafaelx, line 194: Without the stretch spacer, the dialog box is more compact and therefore less wide. But that is possibly a matter of taste.

rafaelx (commented):
The reduction of type casts by use of wxWindowID is definitely an improvement.

src/ui/wxWidgets/PWSafeApp.cpp
rafaelx: This could possibly lead to unexpected effects in other environments.
I would suggest using this as a fallback solution when "setlocale(LC_CTYPE, "")" fails to use the user’s preferred locale.

Can you run the "locale" command in a terminal to see what information it provides?

ronys:
Almost ready for merging into the project's master. Blocking the merge:
- The mac-pwsafe check still fails
- Rafael's comment about locale raises a valid concern, IMHO, and should be addressed.

KaiZin2018:
Using setlocale to determine macOS setting results in "UTF-8", but this default is not running well. I have to add a language in before. Which language is not from interest, but changing the language at the user interface I added an adaptation to the corresponding language.

src/ui/wxWidgets/PWSafeApp.cpp
```diff
@@ -272,9 +272,10 @@ bool PWSafeApp::OnInit()
...
-  wxString typeFull;
-  if (wxGetEnv(wxT("LC_CTYPE"), &typeFull)) {
-    setlocale(LC_CTYPE, typeFull.c_str());
+  char *lcCType = setlocale(LC_CTYPE, NULL);
+
+  if (lcCType && strcmp(lcCType, "UTF-8")) { // MAC OS only have UTF-8 as default, but conversion with this string is nut running well
+    setlocale(LC_CTYPE, "");
   }
   else {
     setlocale(LC_CTYPE, "en_US.UTF-8");
```
rafaelx, line 282: I wonder if "C" would be a better option as a fallback solution. Better in terms of portability if I look at the [man page](https://man7.org/linux/man-pages/man3/setlocale.3.html) of "setlocale". - Just a thought.

KaiZin2018:
"C" will be default at program start-up, but is not working well. With empty "" the default values from environment has to be used, but are also not set in Mac OS standard environment when starting from Finder. The only way to run correct is as actually implemented. I do not have new/additional ideas on that.

KaiZin2018:
Later I will add code to double check on environment variable as done by setlocale(..., ““) to implement the same in my code for the Mac OS.

ronys:
Merged in master - thanks for the effort!

PR #703, opened 2020-11-26
Update tree view to list the entries on different topics

KaiZin2018:
Update tree view to list the entries in groups of group name, first letter of the title, or sort by date (highest date value of each entry). New Menu entry "Sort Tree" to select the highest tree kind.


ronys:
Interesting idea. Two questions come to mind:
1. It seems more natural to sort entries in the table ("List" view, which already supports sorting by any field.
2. In the Windows version, one can drag and drop entries (and groups) between groups in the tree view. Don't see how this would be possible in the sorted versions, which would result in an inconsistent, and confusing, user interface.

KaiZin2018:
Thank you for your interest on the proposed change. The grid view is looking not very clearly represented as a lot of information is given on one line. You cannot reduce the information to the one of interest, as searching for a name or looking to those entries used in the past or in a time (period). In macOS/wxWidget environment you have to arrange the column wide each time when starting up as a fix width is used initially - may be this could be a point of continue work in next release. With my proposal you can reduce the view to those entries of interest, e.g. starting with the same letter or searching in a tree of dates. This is even faster than scrolling through the long list of all entries in grid view.
With wxWidget you cannot drag and drop groups and entries, therefore the macOS/Unix user cannot miss this kind of handling in alphabetic sorted view or date sorted view.
In that way I hope to give you well answers to your points. By any question, please come back to me.

ronys:
Good points, but I'd rather address them by enhancing the grid view (for example, by letting the user select which columns are displayed), and adding support for drag&drop to the tree view.
Filters are another feature that is supported in Windows but not (yet) in the wxWidgets port - this can also help limit the displayed data to the range of interest.

KaiZin2018:
Hello, but working on the grid view allows not to collapse groups as in tree view. The selection of columns is not as simple as changing the kind of tree. When leaving the name and small number of additional columns in grid view only to achieve same overview I'm limited the fields to handle later on. For me this not looking the preferred way.

The drag and drop issue looks not relevant for me, as for such operations I could change back to group tree view.

When I want to browse through all entries to check for a certain name that I do not know in which group and the exact naming a filter is not the right option for me. Similar applies about the date to check on last used or created entries, or such of a specific time period. I do not need a filter if I can direct scan through a date sorted tree, independent from creation date, modified date, or last access date.

As Windows and wxWidget look and feel is not 100% the same the additional grouping in tree view looks well for wxWidget users.

Later I can think on the proposals for the evolution of the wxWidget interface, but not in short time. For the time being my proposal stays as it is, I like it. Please think on it positiv. The user has the option to select the view as he like - I personally never used the grid view before starting the discussion. You must not use all possible options.

Regards 

Kai

KaiZin2018:
As demanded I added the drag and drop feature for wxWidgets. Please check.

KaiZin2018:
Now this branch is on same level last pwsafe/pwsafe, plus 2 commits.

ronys:
Thanks for starting work on drag & drop. A few comments:
- Moving a group into another group should be possible
- Popups at the end of a drop are very user unfriendly
- It should be possible to drop an entry to the root (move it from under a group to toplevel)
- The "..." appearing at the top is, well, odd. Any way to avoid it?
- An outline of the dragged entity (group or entry) while the drag is in effect would be very useful

KaiZin2018:
Hello,

Moving a group into another one is possible, as long as this one is not part of the moving tree, which would lead to problems.

Pop up at the end of a drop is only given in case the drop cannot be performed. Giving no notice and doing nothing in case of an error is also not user friendly.

What is the expected action on moving an entry into the root directory? Removal of the group and leaving empty? I thought an empty group should be avoided. When editing an entry I'm also limited to the list of present group names and cannot leave the name of the group empty, isn't it?

To move a group to the root you need the root inside of the tree. About using ... we can discuss about different naming, like "(root)" or different, but leaving the root name empty is also no good user experience because the empty name entry is harder to recognize and to locate for dropping.

The vxWidgets library support for the demanded outline is not clear for me. What kind of marker do you like to have? The selected entry is not marked after start of the drag action by default, as the mice is starting to select the destination folder. May be I could mark the the original selected entry by a different background that is similar to the selected entry.

Please give your answers to my questions so I can act accordingly.

KaiZin2018:
Ok, adapt to mark the selected entry while selecting the destination (in blue colour). Allow dropping also any entry to the root.

rafaelx:
1. Wow, very impressive how much work you have put in. It would be nice if a pull request could focus on just one topic instead of addressing several different issues so as not to lose overview.
1. In the case of a tree with a very deep hierarchy of elements, the entity currently outlined in blue can become invisible to the user if the tree is scrolled far enough. In order to permanently display the dragged element to the user, its symbol and its textuel representation could follow the mouse pointer in a slightly transparent visual appearance.
Some of the wxWidgets examples ([dnd](https://github.com/wxWidgets/wxWidgets/blob/master/samples/dnd/dnd.cpp), [treectrl](https://github.com/wxWidgets/wxWidgets/blob/master/samples/treectrl/treetest.cpp)) might be helpful here.
1. This out of nowhere appearing root element looks a bit strange indeed.
1. What would be really cool is if the tree view remains the same. This means that not all groups are expand when the dragging begins. A collapsed group could be expanded if the dragged element is directly above the group. If it is moved forward to the next group, the temporarily expanded group could be returned to the previous state or it just remains expanded. At least this behavior would not lead to sudden, bigger changes for the user in the interface, which could be a bit confusing in the first moment. My first impression was. Oops what happened ??? A root item appeared and where is the group I was targeting? I only lost it visually for a moment. However, such implementation as the suggested one may be more challenging and time consuming, but could be continued on a separate branch until a satisfactory solution is found.

ronys:
I've cherry-picked your last commit (1ab4d56) into pwsafe/master, as it's unrelated to the tree work, and definitely an improvement over the current behavior - thanks!

src/ui/wxWidgets/I18N/Makefile
```diff
@@ -39,7 +39,8 @@ $(POS) : $(POT)
 
 mos/%/LC_MESSAGES/pwsafe.mo : pos/pwsafe_%.po
 	@mkdir -p $(dir $@)
-	@msgfmt --check --output-file=$@ $^
+# do not remove wanted entries:	@msgfmt --check --output-file=$@ $^
```
dskrvk, line 42: Why keep old code commented out? If the new version works fine, but someone needs to look up the old one for some reason, they need only to look at git blame.

KaiZin2018: Did the change as demanded. Now only a comment about --check option is left.

KaiZin2018:
To come back to the last comments, (6) the "Show Optimized Cell Size" is an option to increase the size of the row in grid view so that all entries are fitting into the cell.
I managed the scrolling at the end or top while dragging and Collaps and Expand while dragging, when you are waiting on an entry having a child.
About the idea to have the selected item as part of the mice, this is not looking very easy. I will try next, but cannot promise.
What idea can be given about the representation of the root entry? We need one and cannot hide while dragging.

KaiZin2018:
By the way. the examples are very pure.

rafaelx:
Sorry, I forgot to mention the [wxDragImage](https://docs.wxwidgets.org/3.0/classwx_drag_image.html) [example](https://github.com/wxWidgets/wxWidgets/blob/master/samples/dragimag/dragimag.cpp).
The [discussion](https://forums.wxwidgets.org/viewtopic.php?t=40168) from the wxWidgets forum could also be helpful.

KaiZin2018:
Thank you for the tip, I will look what I can do with it.

rafaelx:
In the latest version of this pull request, I cannot expand and collapse any of the groups in the tree view.

KaiZin2018:
Sorry for not testing intensive enough, I will double check tomorrow. Sorry for the inconvenience.

KaiZin2018:
Update will now come to an intermediate version, that fixes the problem with the mouse events. Also the drag object will show the selected item text now snd the cursor changes from pointer to not selectable when selected item is not fitting. For discussion, with this hint on mice view the modal dialog on error might be double or could be removed, when this is still demanded.
One open item, I'm not able to fix so far. When the mouse pointer is went out of the window to scroll faster down or up and the mouse pointer is changing to wxCURSOR_SIZEWE or wxCURSOR_SIZENS it will not turn back to the one controlled in the ListCtrl window. Please help, when you can. I'm lost in this point.
For Cherry Picking the checkin "Avoid error message about empty password ..." could be taken.

src/ui/wxWidgets/I18N/pos/pwsafe_de.po
```diff
@@ -4108,26 +4108,21 @@ msgstr "Das Punktzeichen ist in Gruppennamen nicht erlaubt"
...
+#: ../../../ui/wxWidgets/TreeCtrl.cpp:1065
+#: ../../../ui/wxWidgets/TreeCtrl.cpp:1071
 msgid "Drag and Drop failed"
 msgstr "Drag und Drop fehlgeschlagen"
 
-#: ../../../ui/wxWidgets/TreeCtrl.cpp:1045
+#: ../../../ui/wxWidgets/TreeCtrl.cpp:1065
 #, fuzzy
 msgid "Destination is no group"
 msgstr "Zielfestplatte ist voll."
```
rafaelx, line 4119: Would you be willing to update the translation here and then remove the "fuzzy" keyword?

KaiZin2018: I prepared already a change correcting all fuzzy entries (in pwsafe_de.po, but waiting for feedback from wxWidgets. At first I tried by forum, but now opened a ticket.

https://forums.wxwidgets.org/viewtopic.php?p=202975&sid=037f43db4826987975793ecaec1815ca#p202975

https://trac.wxwidgets.org/ticket/19021

There is one left item with the toolbar menu as well. When translated not always the extended menu ">>" is shown, but can be selected. I have to analyze later.

src/ui/wxWidgets/I18N/Makefile
```diff
@@ -39,7 +39,7 @@ $(POS) : $(POT)
 
 mos/%/LC_MESSAGES/pwsafe.mo : pos/pwsafe_%.po
 	@mkdir -p $(dir $@)
-# do not remove wanted entries:	@msgfmt --check --output-file=$@ $^
+# "--check" option would remove potential used entries, so do not add it to the following line
```
rafaelx, line 42: Could someone please explain to me what effect this actually has on the translations in practice?
Would the resulting effect have an advantage over the missing checks, especially the [format checks](https://www.gnu.org/software/gettext/manual/html_node/msgfmt-Invocation.html)?

KaiZin2018: We have several items in MacOS that are not found in the source tree. Using the clean option always comment out those items, not aligned to a source of line, but needed for text translations on MacOS standard menu.

rafaelx:
I do not see any drag image during the dragging phase. Shouldn't the item's icon be visible?

src/ui/wxWidgets/TreeCtrl.cpp
```diff
@@ -901,6 +1013,436 @@ void TreeCtrl::OnKeyDown(wxTreeEvent& evt)
...
+       ((GetRootItem() != itemDst) || (GetRootItem() != GetItemParent(m_drag_item)))) { // Do not Drag and Drop on its own
+      if(GetRootItem() != itemDst) {
+        if(! ItemIsGroup(itemDst)) {                   // Only group may be destination
+          wxMessageBox(_("Drag and Drop failed"), _("Destination is no group"), wxOK|wxICON_ERROR);
+          evt.Skip();
+          m_drag_item = NULL;
+          return;
+        }
+        else if(IsDescendant(itemDst, m_drag_item)) {  // Do not drag and drop into the moved tree
+          wxMessageBox(_("Drag and Drop failed"), _("Destination cannot be inside source tree"), wxOK|wxICON_ERROR);
```
rafaelx, line 1347: It seems that the message and the title are interchanged in the message boxes.

src/ui/wxWidgets/I18N/pos/pwsafe_de.po
```diff
@@ -2439,179 +2419,188 @@ msgid "Show Menu Separator"
 msgstr "Menütrenner anzeigen"
 
 #: ../../../ui/wxWidgets/OptionsPropertySheetDlg.cpp:538
+msgid "Show Optimized Cell size"
+msgstr "Zeige optimierte Zell-Größe im Grid"
```
rafaelx, line 2423: Since "Grid" refers to the "List View", I would suggest using a similar German term for "Grid" as it is already used in the application.
Additionally, based on the already available translations, I would also suggest putting the verb at the end of the sentence.

For instance:
> Optimierte Zellgröße in der Tabellenansicht anzeigen

src/ui/wxWidgets/I18N/Makefile
rafaelx: At least in my opinion, the "check-format" option is very valuable if it is not included elsewhere on the toolchain.

src/ui/wxWidgets/I18N/pos/pwsafe_de.po
rafaelx: The discussion and reported bug are related to a problem with the cursor during drag and drop, if I'm not mistaken. I don't understand how the strings marked as fuzzy relate to this.

src/ui/wxWidgets/TreeCtrl.cpp
KaiZin2018: May be, please advice which order is preferred.

KaiZin2018:
The items text is visible near to the cursor while drag and drop inside of the tree view. Icon and text is not implemented.

src/ui/wxWidgets/I18N/Makefile
KaiZin2018: There is no missing format check, but the entries defined in the pwsage_xx.po not related to a source line, e.g. those for MacOS standard menu, are comment out when the check option is present.

src/ui/wxWidgets/I18N/pos/pwsafe_de.po
KaiZin2018: I have overtaken the proposal, thanks a lot.

src/ui/wxWidgets/TreeCtrl.cpp
KaiZin2018: Proposal overtaken (twi places).

src/ui/wxWidgets/I18N/pos/pwsafe_de.po
KaiZin2018: Text corrected - make pos seam to change to fuzzy wording.

rafaelx:
In my case, the text doesn't appear with Wx 3.1.3 (Gtk 3.24.18) and Wx 3.0.4 (Gtk 3.24.14) under Kubuntu, but with Wx 3.0.4 (Gtk 2.24.32) under LinuxMint. It would be interesting to find out if it is due to the different window managers or Gtk versions.

A little remark. What I noticed is that the text is slightly covered by the mouse pointer when the drag starts. When the mouse pointer is moved to a potential drop target, the text is repositioned slightly so that it becomes fully readable.

src/ui/wxWidgets/I18N/Makefile
KaiZin2018: I think I have to change back. The problems occurs more when the po file is newly generated. I will update the makefile soon.

src/ui/wxWidgets/TreeCtrl.h
```diff
@@ -51,6 +52,44 @@ class TreeCtrl;
...
+
+  virtual void Notify();
+
+private:
+  TreeCtrl *m_owner;
+
+  wxDECLARE_NO_COPY_CLASS(TreeScrollTimer);
+};
+
+class TreeCollapseTimer: public wxTimer
```
rafaelx, line 75: Now that we have different uses (YubiMixin, TreeCtrl) for timers, it would be a good opportunity to introduce a generic timer implementation at wxUtilities.

rafaelx:
> What idea can be given about the representation of the root entry? We need one and cannot hide while dragging.

Just for a better understanding. Is the root element needed in case a password element needs to be moved out of a group to make it groupless?
Would it be an option to extend the hit test to see if the object being dragged is outside the tree? If the entity is dropped outside the tree, it no longer belongs to any group.

src/ui/wxWidgets/TreeCtrl.cpp
```diff
@@ -901,6 +1013,436 @@ void TreeCtrl::OnKeyDown(wxTreeEvent& evt)
...
+  wxPoint mousePt = ScreenToClient(wxGetMousePosition());  // Find where the mouse is in relation to the (exited) pane
+  
+  GetSize(&width, &height);  // Store window dimensions
+  
+  if((mousePt.x < 0) || (mousePt.x >= width) || (mousePt.y < 0) || (mousePt.y >= height)) { // Not in same area
+    m_last_mice_item_in_drag_and_drop = NULL;
+    return;
+  }
+  
+  int flags = wxTREE_HITTEST_ONITEMICON|wxTREE_HITTEST_ONITEMLABEL;
```
rafaelx, line 1104: Just as a hint. The combination of 'wxTREE_HITTEST_ONITEMICON|wxTREE_HITTEST_ONITEMLABEL' is provided by 'wxTREE_HITTEST_ONITEM'.

src/ui/wxWidgets/TreeCtrl.cpp
```diff
@@ -901,6 +1013,436 @@ void TreeCtrl::OnKeyDown(wxTreeEvent& evt)
...
+      GetEventHandler()->ProcessEvent(scrollEvent);
+      m_scroll_timer.Start( TreeScrollTimer::DELAY, true );
+    }
+    else {
+      resetScrolling();
+    }
+  }
+}
+
+void TreeCtrl::CheckCollapsEntry()
```
rafaelx, line 1090: Small typo ('CheckCollapsEntry' -> 'CheckCollapseEntry').

src/ui/wxWidgets/TreeCtrl.cpp
```diff
@@ -901,6 +1013,436 @@ void TreeCtrl::OnKeyDown(wxTreeEvent& evt)
...
+{
+  if(! m_last_mice_item_in_drag_and_drop.IsOk()) return;
+  
+  int width, height;
+  wxPoint mousePt = ScreenToClient(wxGetMousePosition());  // Find where the mouse is in relation to the (exited) pane
+  
+  GetSize(&width, &height);  // Store window dimensions
+  
+  if((mousePt.x < 0) || (mousePt.x >= width) || (mousePt.y < 0) || (mousePt.y >= height)) { // Not in same area
+    m_last_mice_item_in_drag_and_drop = NULL;
```
rafaelx, line 1100: I would suggest using 'nullptr'.

src/ui/wxWidgets/TreeCtrl.cpp
```diff
@@ -901,6 +1013,436 @@ void TreeCtrl::OnKeyDown(wxTreeEvent& evt)
...
+        Expand(currentItem);
+        //scroll the last child of this node into visibility
+        EnsureVisible(GetLastChild(currentItem));
+        //but if that scrolled the parent out of the view, bring it back
+        EnsureVisible(currentItem);
+      }
+    }
+  }
+  else {
+    m_last_mice_item_in_drag_and_drop = NULL;
```
rafaelx, line 1125: I would suggest using 'nullptr'.

src/ui/wxWidgets/TreeCtrl.cpp
```diff
@@ -901,6 +1013,436 @@ void TreeCtrl::OnKeyDown(wxTreeEvent& evt)
...
+    wxDELETE(m_drag_image);
+  }
+  Refresh();
+  
+  wxTreeItemId selectedItem = GetSelection();
+  if(selectedItem.IsOk()) UnselectAll();
+  
+  if((mousePt.x < 0) || (mousePt.x >= width) || (mousePt.y < 0) || (mousePt.y >= height)) { // Not in same area
+    evt.Skip();
+    m_drag_item = NULL;
```
rafaelx, line 1321: I would suggest using 'nullptr'.

src/ui/wxWidgets/TreeCtrl.cpp
```diff
@@ -901,6 +1013,436 @@ void TreeCtrl::OnKeyDown(wxTreeEvent& evt)
...
+      return;
+    }
+
+    if(itemDst && itemDst.IsOk() &&
+       (itemDst != m_drag_item) &&
+       (itemDst != GetItemParent(m_drag_item)) &&
+       ((GetRootItem() != itemDst) || (GetRootItem() != GetItemParent(m_drag_item)))) { // Do not Drag and Drop on its own
+      if(GetRootItem() != itemDst) {
+        if(! ItemIsGroup(itemDst)) {                   // Only group may be destination
+          wxMessageBox(_("Destination is no group"), _("Drag and Drop failed"), wxOK|wxICON_ERROR);
```
rafaelx, line 1341: I would suggest "Destination is not a group.".

src/ui/wxWidgets/TreeCtrl.cpp
```diff
@@ -901,6 +1013,436 @@ void TreeCtrl::OnKeyDown(wxTreeEvent& evt)
...
+          sxNewPath = tostringx(GetItemGroup(itemDst) + GROUP_SEL_STR + label);
+        }
+        else {
+          sxNewPath = tostringx(label);
+        }
+        if(! doOverride) {
+          wxTreeItemId si;
+          if(ExistsInTree(itemDst, tostringx(label), si)) {
+            evt.Veto();
+            wxMessageBox(_("Same group name exists (use Shift to overrule)"), _("Double group name"), wxOK|wxICON_ERROR);
```
rafaelx, line 1400: I would suggest "Duplicate group name".

src/ui/wxWidgets/SafeCombinationChangeDlg.cpp
```diff
@@ -295,7 +301,7 @@ void SafeCombinationChangeDlg::OnOkClick(wxCommandEvent& WXUNUSED(evt))
       wxMessageDialog err(this, _("Cannot verify old safe combination - file gone?"),
                           _("Error"), wxOK | wxICON_EXCLAMATION);
       err.ShowModal();
-    } else if (m_confirm != m_newpasswd) {
+    } else if (m_isPasswordHidden && (m_confirm != m_newpasswd)) {
```
rafaelx, line 304: Shouldn't it be or'ed?

src/ui/wxWidgets/PasswordSafeFrame.cpp
```diff
@@ -684,6 +706,25 @@ void PasswordSafeFrame::CreateMenubar()
...
+/**
+ * Update check for sorting and enable or disable the menu bar for orting
+ */
+void PasswordSafeFrame::UpdateTreeSortMenu()
+{
+  auto menuBar = GetMenuBar();
+  
+  menuBar->Check(ID_SORT_TREE_BY_GROUP, IsTreeSortGroup() ? true : false);
+  menuBar->Check(ID_SORT_TREE_BY_NAME, IsTreeSortName() ? true : false);
+  menuBar->Check(ID_SORT_TREE_BY_DATE, IsTreeSortDate() ? true : false);
```
rafaelx, line 721: Wouldn't it be sufficient to use the return value of the function directly?

src/ui/wxWidgets/PasswordSafeFrame.cpp
```diff
@@ -684,6 +706,25 @@ void PasswordSafeFrame::CreateMenubar()
...
+
+  menuBar->Enable(ID_SORT_TREE_MENU, IsTreeView() ? true : false);
+  menuBar->Enable(ID_SORT_TREE_BY_GROUP, IsTreeView() ? true : false);
+  menuBar->Enable(ID_SORT_TREE_BY_NAME, IsTreeView() ? true : false);
+  menuBar->Enable(ID_SORT_TREE_BY_DATE, IsTreeView() ? true : false);
```
rafaelx, line 726: Is this still needed if already handled by OnUpdateUI?

src/ui/wxWidgets/GridTable.cpp
```diff
@@ -162,6 +162,24 @@ void GridTable::SetView(wxGrid* newGrid)
...
+        for(unsigned int row = 0, noRows = GetNumberRows(); row < noRows; ++row) {
+          if(!IsEmptyCell(row, idx)) {
+            dc.GetTextExtent(GetValue(row, idx), &width, &height);
+            width += border;
+            if(width > PWSGridCellData[idx].width) {
+              PWSGridCellData[idx].width = width;
+            }
+          }
+        }
+      }
```
rafaelx, line 182: Would it be an option to use [wxGrid::AutoSizeColumns](https://docs.wxwidgets.org/3.0/classwx_grid.html#a412837fe2932fcde8473e6fbf64f8ad5) and let the framework do the calculation of the best column width?

src/core/PWSprefs.cpp
```diff
@@ -132,6 +132,8 @@ const PWSprefs::boolPref PWSprefs::m_bool_prefs[NumBoolPrefs] = {
   {_T("ListSortAscending"), true, ptApplication},           // application
   {_T("EnableWindowTransparency"), false, ptApplication},   // application
   {_T("ShowMenuSeparator"), true, ptApplication},           // application
+  {_T("OptimizedCellSize"), false, ptApplication},          // application
+  {_T("ToolbarShowText"), true, ptApplication},             // application
```
rafaelx, line 136: I would suggest not to surprise the user with a change in the user interface by default and thus set the default value for "ToolbarShowText" to "false".

src/ui/wxWidgets/TreeCtrl.h
KaiZin2018: The generic timer is wxTimer, as the specific one, with dedicated time and handler is the generate, as the given reference.

src/ui/wxWidgets/TreeCtrl.cpp
KaiZin2018: Replaced three times to wxTREE_HITTEST_ONITEM. Thank you for the hint.

src/ui/wxWidgets/TreeCtrl.cpp
KaiZin2018: Corrected, thank you for the hint.

src/ui/wxWidgets/TreeCtrl.cpp
KaiZin2018: It's not 100% used as nullptr throughout the source tree - NULL can be seen at other places also for pointers - but I did as proposed for the new code.

src/ui/wxWidgets/TreeCtrl.cpp
KaiZin2018: I did as proposed for the new code.

src/ui/wxWidgets/TreeCtrl.cpp
KaiZin2018: I did as proposed for the new code.

src/ui/wxWidgets/TreeCtrl.cpp
KaiZin2018: Proposal is fine with me.

src/ui/wxWidgets/TreeCtrl.cpp
KaiZin2018: Proposal is fine with me.

src/ui/wxWidgets/PasswordSafeFrame.cpp
KaiZin2018: Although the logic is better readable with the explicit values I followed the demanded change.

src/ui/wxWidgets/PasswordSafeFrame.cpp
KaiZin2018: You are right, 4 lines removed.

src/ui/wxWidgets/GridTable.cpp
KaiZin2018: I did not managed to adapt the code using wxGrid::AutoSizeColumns to have the expected functionality.

src/core/PWSprefs.cpp
KaiZin2018: Adapted to false as demanded.

KaiZin2018:
To answer to the questions above, the root is needed to remove the group from an item or place a group direct to the highest level when using drag and drop.
The test is already done to be inside of the tree window. When inside of the tree window it is double checked in dropping end action if the destination item is also beneath the cursor (hit test).

src/ui/wxWidgets/GridTable.cpp
rafaelx: Hm, an event handler for double-click events is set up in the GridCtrl [constructor](https://github.com/pwsafe/pwsafe/blob/0e789f083e1f3dfff8254af819642d1dd4ed03f2/src/ui/wxWidgets/GridCtrl.cpp#L66), from which [wxGrid::AutoSizeColumn](https://github.com/pwsafe/pwsafe/blob/0e789f083e1f3dfff8254af819642d1dd4ed03f2/src/ui/wxWidgets/GridCtrl.cpp#L81) is already successfully used when a double-click event is executed on one of the column separators.

src/ui/wxWidgets/TreeCtrl.cpp
rafaelx: It is not so dramatic having NULL in the source code, but there have already been some replacements of NULL by nullptr in the past. However, there could still be some occurrences of NULL here and there in the source code. It would at least be  good if NULL were no longer introduced. - Thanks for the changes.

src/ui/wxWidgets/TreeCtrl.h
rafaelx: By generic I mean a timer implementation that better suits the needs of the project so that a new class implementation does not have to be created for each new timer, that follows the same pattern and which only encapsulates the timeout value and a handler. Both information, the timeout value and the callback function, could be specified when setting up the timer. Such a timer could also offer a parameterless start function. If it already knows the timeout value to be used there is no need to pass the already known value to its own function again. I just think about preventing code duplication. - This is just an idea.

src/ui/wxWidgets/GridTable.cpp
KaiZin2018: Had to call function AusSizeColumns outside of the loop.

src/ui/wxWidgets/TreeCtrl.h
KaiZin2018: Please provide a proposal class, I'm not good in implementing virtual classes/functions.

src/core/PWSprefs.cpp
```diff
@@ -210,6 +212,8 @@ const PWSprefs::stringPref PWSprefs::m_string_prefs[NumStringPrefs] = {
   {_T("AddEditFont"), _T(""), ptApplication },                      // application
   {_T("AddEditSampleText"), _T("AaBbYyZz 0O1IlL"), ptApplication},  // application
   {_T("AltNotesEditorCmdLineParms"), _T(""), ptApplication},        // application
+  {_T("lastsort"), _T("group"), ptApplication},                     // application
```
ronys, line 215: This should be renamed to "TreeSort":
- "last" is redundant
- "lastsort" doesn't make clear what's sorted
- Capitalization is inconsistent with the other preferences.

src/core/PWSprefs.cpp
```diff
@@ -341,7 +345,7 @@ StringX PWSprefs::GetAllStringPrefs(const bool bUseCopy)
     LastView, TreeFont, BackupPrefixValue, ListColumns,
     ColumnWidths, MainToolBarButtons, PasswordFont,
     TreeListSampleText, PswdSampleText,
-    LastUsedKeyboard, VKeyboardFontName, VKSampleText, LanguageFile
+    LastUsedKeyboard, VKeyboardFontName, VKSampleText, LanguageFile, LastSort
```
ronys, line 348: LastSort -> TreeSort (see previous comment)

src/core/PWSprefs.cpp
```diff
@@ -132,6 +132,8 @@ const PWSprefs::boolPref PWSprefs::m_bool_prefs[NumBoolPrefs] = {
   {_T("ListSortAscending"), true, ptApplication},           // application
   {_T("EnableWindowTransparency"), false, ptApplication},   // application
   {_T("ShowMenuSeparator"), true, ptApplication},           // application
+  {_T("OptimizedCellSize"), false, ptApplication},          // application
```
ronys, line 135: Why is this a user-visible preference? When would a user set or clear this?

src/ui/wxWidgets/PasswordSafeFrame.h
```diff
@@ -189,7 +193,7 @@ class PasswordSafeFrame : public wxFrame, public Observer
 
 private:
     enum class ViewType { TREE, GRID };
-
+    enum class SortType { GROUP, NAME, DATE };
```
ronys, line 196: SortType --> TreeSortType
for readability/maintainability

ronys (changes requested):
It's really unclear what "Show Optimized Cell Size" means as a display preference.

It makes much more sense to me to implement this as a "reset widths" action / menu item in the popup menu displayed when a user right-clicks on the column header rows in list view.

src/ui/wxWidgets/TreeCtrl.h
rafaelx: The idea is to encapsulate the handler function as well, perhaps by replacing the specific receiver of type wxTreeCtrl with an interface/abstract class/lambda. The timer implementation could provide a handler registration function which is then called within the notification function, as is already the case in the current version, but this could be based on a more generic mechanism, so that it can be reused in other places as well.
Another possibility could be to make use of [wxTimer::SetOwner](https://docs.wxwidgets.org/3.0/classwx_timer.html#aa3966b37329a4fad69ad384733eab27e).
Certainly the current implementation is specially designed for TreeCtrl and thus the receiver of the type wxTreeCtrl is very specific here, but could perhaps also be of the type wxControl or wxEvtHandler. For now, however, the implementation can be left as is, and architectural improvements can be added later.

rafaelx:
> It's really unclear what "Show Optimized Cell Size" means as a display preference.
> 
> It makes much more sense to me to implement this as a "reset widths" action / menu item in the popup menu displayed when a user right-clicks on the column header rows in list view.

We, the people, want "Adjust widths". ;-)

rafaelx:
> To answer to the questions above, the root is needed to remove the group from an item or place a group direct to the highest level when using drag and drop.
> The test is already done to be inside of the tree window. When inside of the tree window it is double checked in dropping end action if the destination item is also beneath the cursor (hit test).

I was just thinking if the user drags the item out of the tree so that it is still within the valid drag area, but not over an existing group, it could be used as an indicator that the user wants to have the item at the top level, therefore as a groupless item. Perhaps the Hit-Test could be used for this purpose in order not to have to display the root element.

src/core/PWSprefs.cpp
KaiZin2018: Replaced last sort/LastSort by TreeSort

src/core/PWSprefs.cpp
KaiZin2018: Replaced last sort/LastSort by TreeSort

src/core/PWSprefs.cpp
KaiZin2018: This is to set at users demand before using the grid view or while he is working the first time with it. Either the screen is big enough to expand and scrolling in the grid view is not needed or no problem for the user or the user is not using this key and align the column wide by his own. A command in the menu might be also an idea, instead of hiding in the options menu.

src/ui/wxWidgets/PasswordSafeFrame.h
KaiZin2018: Done as proposed

KaiZin2018:
> > To answer to the questions above, the root is needed to remove the group from an item or place a group direct to the highest level when using drag and drop.
> > The test is already done to be inside of the tree window. When inside of the tree window it is double checked in dropping end action if the destination item is also beneath the cursor (hit test).
> 
> I was just thinking if the user drags the item out of the tree so that it is still within the valid drag area, but not over an existing group, it could be used as an indicator that the user wants to have the item at the top level, therefore as a groupless item. Perhaps the Hit-Test could be used for this purpose in order not to have to display the root element.

Dragging outside the tree is for me a user that is misliking his action and do no know how to stop the action. The Hit-Test is only to make sure the item is correctly selected. When leaving the dragging object at any place, e.g. outside or not over correctly selected item, it is unexpected when this one would be moved to the top level.

KaiZin2018:
> > It's really unclear what "Show Optimized Cell Size" means as a display preference.
> > It makes much more sense to me to implement this as a "reset widths" action / menu item in the popup menu displayed when a user right-clicks on the column header rows in list view.
> 
> We, the people, want "Adjust widths". ;-)

This is to set at users demand before using the grid view or while he is working the first time with it. Either the screen is big enough to expand and scrolling in the grid view is not needed or no problem for the user or the user is not using this key and align the column wide by his own. A command in the menu might be also an idea, instead of hiding in the options menu.

KaiZin2018:
I have added a generic class for the two TreeCtrlTimers. May be the code can be extend later as general one for other timers.

src/ui/wxWidgets/TreeCtrl.h
```diff
@@ -54,42 +54,33 @@ typedef std::map<pws_os::CUUID, wxTreeItemId, std::less<pws_os::CUUID> > UUIDTIM
...
+    DELAY_SCROLLING = 250,
     // start Collapse or Expand after 1.5 second (if the mouse hasn't been clicked/moved)
-  enum { DELAY = 1500 };
-    
-  TreeCollapseTimer();
+    DELAY_COLLAPSE = 1500
+  };
+  
+  TreeCtrlTimer(TreeCtrl *aowner, TreeCtrlMemberFncPtr acallback, int atime, bool aonetime = true) : m_owner(aowner), m_callback(acallback), m_time(atime), m_one_time(aonetime) { }
+  TreeCtrlTimer(TreeCtrlMemberFncPtr acallback, int atime, bool aonetime = true) : m_owner(nullptr), m_callback(acallback), m_time(atime), m_one_time(aonetime) { }
```
rafaelx, line 70: The compiler issues many warnings about the initialization order of the timer-related members.

rafaelx:
Based on the reported issue #704, it would be great if another pwsafe instance could be a drop target so that items can be easily copied between different databases. - That would be super terrific.

src/ui/wxWidgets/TreeCtrl.h
KaiZin2018: > * The compiler issues many warnings about the initialization order of the timer-related members.
Next check in solves the issues about -Wreorder
> * Do we need the second constructor and TreeCtrlTimer::setOwner ?
I do not need this, but left for later deriving a general timer class from this code part.

KaiZin2018:
> Based on the reported issue #704, it would be great if another pwsafe instance could be a drop target so that items can be easily copied between different databases. - That would be super terrific.

This kind of function is covered by the proposed merge, compare etc. from functional point of view. For drag and drop between two instances a different mechanism as already started in the drag-bar must be used. Mixing this two things may be taken later, but not at the moment. At first I want to stabilize (and overtake in main stream) this first drag and drop inside one instance.

ronys:
Playing with this a bit more, and comparing it to the drag & drop implementation in Windows, a few comments:
- When the drop target is an entry, instead of popping up an error message ("Destination is not a group"), it would be more user-friendly to find the group associated with the entry and make that the effective drop target. This would also make it easier to drop to the root - just drag to a toplevel entry.
- Associating a drop to the blank area with the toplevel seems natural for the Windows version. It's also consistent with what happens when you right-click to create a new entry or group in the wx version. This action doesn't change the display, so it's inconsistent to see drag & drop doing this.

In any case, I appreciate the work you're doing here, and hope to continue collaborating next year as well. Here's to a happy & healthy 2021!

KaiZin2018:
> Playing with this a bit more, and comparing it to the drag & drop implementation in Windows, a few comments:
> 
> * When the drop target is an entry, instead of popping up an error message ("Destination is not a group"), it would be more user-friendly to find the group associated with the entry and make that the effective drop target. This would also make it easier to drop to the root - just drag to a toplevel entry.

Follow the proposal to use the parent instead of the actual flat entry when this one is selected in dropping. 

> * Associating a drop to the blank area with the toplevel seems natural for the Windows version. It's also consistent with what happens when you right-click to create a new entry or group in the wx version. This action doesn't change the display, so it's inconsistent to see drag & drop doing this.

The root item is now gone (as long as no hard coded XML change is performed). To place into the root the item must be dropped below the list, which is similar to windows or macOS finder behavior, when looking into the root folder of a drive.

> In any case, I appreciate the work you're doing here, and hope to continue collaborating next year as well. Here's to a happy & healthy 2021!

Happy new year and stay healthy

rafaelx:
I really like the way drag and drop behaves now. Very user friendly.

I noticed that after moving an element to which a shortcut exists, the icon indicating that the item is being used as a base entry is replaced with an icon that an item gets if it has no shortcut.

src/ui/wxWidgets/I18N/pos/pwsafe_de.po
```diff
@@ -5956,3739 +5956,33 @@ msgstr "Deutsch"
...
-#~ msgid "subset: '%s' field '%s' '%s' %s"
-#~ msgstr "Teilmenge: '%s' Feld '%s' '%s' %s"
+msgid "Bring All to Front"
+msgstr "Alle Einblenden"
 
-#~ msgid "this entry's password and data"
-#~ msgstr "Passwort und Daten dieses Eintrags"
+msgid "Quit PasswordSafe (wxWidgets)"
+msgstr "PasswordSafe beenden"
+msgid "Hide PasswordSafe (wxWidgets)" msgstr "PasswordSafe ausbelenden"
```
rafaelx, line 5964: - "msgstr" could be on a line of its own
- Typo in "PasswordSafe ausbelenden" -> "PasswordSafe ausblenden"

src/ui/wxWidgets/I18N/pos/pwsafe_de.po
```diff
@@ -5956,3739 +5956,33 @@ msgstr "Deutsch"
...
 
-#~ msgid "top"
-#~ msgstr "Beginn"
+# wxString standard strings
+msgid "Create new document"
+msgstr "Erzeuge ein neues Dokument"
```
rafaelx, line 5968: Just a suggestion.

- "Ein neues Dokument erstellen"

src/ui/wxWidgets/I18N/pos/pwsafe_de.po
```diff
@@ -5956,3739 +5956,33 @@ msgstr "Deutsch"
...
 
-#~ msgid "whatever"
-#~ msgstr "egal"
+# wxString standard strings
+msgid "Open an existing document"
+msgstr "Öffnen eines vorhandenen Dokumentes"
```
rafaelx, line 5972: Just a suggestion.

- "Ein vorhandenes Dokument öffnen"

rafaelx: Hmm, should the translation of "document" possibly be replaced by "database" in the context of pwsafe?
However, I am not a Mac user and so I am not familiar with the concept of a Mac user interface, nor with the standard menu.

src/ui/wxWidgets/I18N/pos/pwsafe_de.po
KaiZin2018: > * "msgstr" could be on a line of its own
> * Typo in "PasswordSafe ausbelenden" -> "PasswordSafe ausblenden"

corrected

src/ui/wxWidgets/I18N/pos/pwsafe_de.po
KaiZin2018: Good idea, I just have overtaken the proposal.

src/ui/wxWidgets/I18N/pos/pwsafe_de.po
KaiZin2018: To not deviate too much from English translation I prefer to stay with document/Dokument.

KaiZin2018:
> I really like the way drag and drop behaves now. Very user friendly.
> 
> I noticed that after moving an element to which a shortcut exists, the icon indicating that the item is being used as a base entry is replaced with an icon that an item gets if it has no shortcut.

Corrected the problem with the shortcut and also the problems with selection of empty group in edit/create dialog.

ronys:
Like @rafaelx, I'm very pleased with the drag & drop at this stage.

Question now is whether we wait for the "Show Optimized Cell size" comments to be addressed, or merge this as-is into master, and immediately create a branch to do this work.

What say?

KaiZin2018:
> Like @rafaelx, I'm very pleased with the drag & drop at this stage.
> 
> 
> 
> Question now is whether we wait for the "Show Optimized Cell size" comments to be addressed, or merge this as-is into master, and immediately create a branch to do this work.
> 
> 
> 
> What say?

Sorry, I'm not aware of any addressed change on "Show Optimized Cell size". The actual state is as aligned with Rafael. Point of discussion might be location of switching, either options as it actually implemented or new menu item that can be selected to switch the function on and off. Please give your opinion on that and I will implement as demanded.

ronys:
Re "Show Optimized Cell size", I was referring to [this ](https://github.com/pwsafe/pwsafe/pull/703#issuecomment-752397533) comment.

Is it really a persistent boolean state, or an action that one-time resets the widths? I'm assuming the latter.

This should (a) be moved to the View menu, (b) displayed only when the current view is grid, (c) made a regular menu  entry - not a checkbox and finally (d) renamed to "Adjust column widths".

KaiZin2018:
> Is it really a persistent boolean state, or an action that one-time resets the widths? I'm assuming the latter.

In my idea it is a persistent boolean state for all grid. The cell size in the grid has to fit always, as long as the flag is set and no need is present to double click on column border to resize the column.

> This should (a) be moved to the View menu, (b) displayed only when the current view is grid, (c) made a regular menu  entry - not a checkbox and finally (d) renamed to "Adjust column widths".

Due to the different assumption the place in options looks well. The resize of all cells as a single action could be implemented in addition as additional solution for those who do not like to have the growing of cell width permanently fit to the maximum size, but call this function on demand.

ronys:
Aha. *Now* I understand what this is supposed to do. Thanks.
So, in light of this:
1. It should indeed remain a Boolean preference / checkbox
2. The checkbox should be renamed to "Auto-adjust column width"
3. The preference should be renamed to "AutoAdjColWidth"

After this I think we'll be ready to merge.

KaiZin2018:
> Aha. _Now_ I understand what this is supposed to do. Thanks.
> So, in light of this:
> 
> 1. It should indeed remain a Boolean preference / checkbox
> 2. The checkbox should be renamed to "Auto-adjust column width"
> 3. The preference should be renamed to "AutoAdjColWidth"
> 
> After this I think we'll be ready to merge.

I followed your advice and has renamed the checkbox and the preference.

PR #706, opened 2021-01-04
Add some support for building/running on Apple silicon

sauravg:
Added some instructions on what to do to run pwsafe on the new M1 Macs.

This undoes eaf2c3b27. I'm not sure why that hard-coded a path into the configs. The config generator script (Xcode/generate-configs) will automatically add the path if the dependencies are to be found in that path.


ronys:
Thanks!

PR #708, opened 2021-01-08
Remove warning about implicit conversion and fix problem with filter view

KaiZin2018:
The last change on master had introduced implicit conversion of parameter, which lead to warnings. This is fixed with the change.
In addition the filter, e.g. not saved entries, had not been updated at the gui when undo or redo is done while the filter is active. This problem is fixed as well.


ronys:
Thanks!

PR #709, opened 2021-01-08
Extend update on filter for GridView and correct for_each statement on grid entry deletion

KaiZin2018:
Sorry for this high frequency of changes, on looking for adapting the previous change to the grid view an error on an earlier change had been noticed that can lead to assertion failure when deleting or updating (undo/redo) entries in grid view.

### DIFF
--- a/src/core/Command.cpp
+++ b/src/core/Command.cpp
@@ -404,6 +404,8 @@ DBEmptyGroupsCommand::DBEmptyGroupsCommand(CommandInterface *pcomInt,
 int DBEmptyGroupsCommand::Execute()
 {
   if (!m_pcomInt->IsReadOnly()) {
+    SaveDBInformation();
+      
     bool bChanged(false);
     if (m_bSingleGroup) {
       // Single Empty Group functions
@@ -414,7 +416,7 @@ int DBEmptyGroupsCommand::Execute()
           break;
         case EG_DELETE:
           bChanged = m_pcomInt->RemoveEmptyGroup(m_sxEmptyGroup);
-          m_pcomInt->AddChangedEmptyGroups(m_sxEmptyGroup);
+          // m_pcomInt->AddChangedEmptyGroups(m_sxEmptyGroup);
           break;
         case EG_RENAME:
           bChanged = m_pcomInt->RenameEmptyGroup(m_sxOldGroup, m_sxNewGroup);
@@ -503,6 +505,7 @@ void DBEmptyGroupsCommand::Undo()
           break;
       }
     }
+    RestoreDBInformation();
     if (m_bNotifyGUI) {
       m_pcomInt->NotifyGUINeedsUpdating(UpdateGUICommand::GUI_REFRESH_TREE,
         CUUID::NullUUID());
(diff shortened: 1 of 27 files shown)


ronys:
The functionality added in the last commit (471a172) is problematic, and will be even more so when full filter functionality is added (arbitrary filtering criteria).
For the predefined subgroups:
- Unsaved changes: If the user added an empty group, this should always be shown
- Entries with expiry dates: The result should never contain an empty group
- Last find results: If the find includes an empty group (not sure this is possible, as find iterates only over entries, as far as I recall), then it should be shown.

Can you remove/revert this commit so that I can merge in the other fixes, or would you rather I do this locally?

KaiZin2018:
> The functionality added in the last commit ([471a172](https://github.com/pwsafe/pwsafe/commit/471a172c42f88b366d44bdc556788263d688b78e)) is problematic, and will be even more so when full filter functionality is added (arbitrary filtering criteria).
> For the predefined subgroups:
> 
> * Unsaved changes: If the user added an empty group, this should always be shown

Indication of empty groups is activated by default now.

> * Entries with expiry dates: The result should never contain an empty group

Indication of empty groups is disabled by default now.

> * Last find results: If the find includes an empty group (not sure this is possible, as find iterates only over entries, as far as I recall), then it should be shown.

Indication of empty groups is disabled by default now (iteration is over entries only).

> Can you remove/revert this commit so that I can merge in the other fixes, or would you rather I do this locally?

I hope the resulting implementation is sufficient for you. I will look for a long time implementation of scanning the undo-command vector to give back a list of changed (empty) groups at a later point of time - this will need some time.

ronys:
I'd also like the menu entry (which has a spelling mistake) to be removed.

PasswordSafe's user interface is already much more complicated than I'd like, but most of the complexity is in dialogs and tabs that the typical user doesn't see. Adding another menu item is much more intrusive. In this case I don't think the added functionality justifies this.

KaiZin2018:
> I'd also like the menu entry (which has a spelling mistake) to be removed.

As the menu item is needed for drag and drop in filter view I propose to stay with it, but move to lower menu level, as it is used only in special cases.

> PasswordSafe's user interface is already much more complicated than I'd like, but most of the complexity is in dialogs and tabs that the typical user doesn't see. Adding another menu item is much more intrusive. In this case I don't think the added functionality justifies this.

Moving down the menu item to a lower level will reduce confusion for normal usage but allow special use cases - free drag and drop of filtered items.

I hope you can follow my idea - if not if will make the compilation of the menu item depending on compilation flag. Standard would be not present.

ronys:
Moving the "Empty Group visible in Filter" menu item to the Filters submenu is a clever idea, but:
- The use-case for explicitly turning it on (or off) doesn't seem very practical to me. I don't think people will use drag & drop in a filtered view. I expect filters to be used to find entries that meet specified criteria. Editing these seems untypical, and editing them by moving them to an empty group seems even more unlikely.
- The Filters submenu is about managing filters. Adding an action that has to do with the display isn't consistent with the topic of this submenu.
Bottom line: While nice for completeness, I don't think the ability to toggle the visibility of empty groups in the filter view is useful enough to add as a submenu item.
Where it *would* make sense perhaps is as part of a filter definition, when the ability to add user-defined filters is implemented.

KaiZin2018:
I hope the last changes will fit to your idea of interface. The "Empty Group visible in Filter" is left in Filter menu, but this one is disabled. The predefined filter on all changed entries is now including only changed empty groups, maintained in a separate list.

KaiZin2018:
That's it. No more changes expected, except you will address something to me.
Unfortunately an additional fix is included into the change history.

ronys:
Made a small change, tested on Windows and Linux.
Thanks!

PR #715, opened 2021-02-08
Add Drag and Drop between instances on wxWidgets

KaiZin2018:
This change will add drag and drop facility between instances of pwSafe. In addition the deletion of groups will give warning to be confirmed in case of alias/shortcut outside of deleted tree only.


KaiZin2018:
Hint for drag and drop: wxWidget 3.1.4 is not running correct, but 3.1.5 will do, same as 3.0.5. Use Shift-Key on drag for forcing using the full path and in drop to force plains into root. You can change the keyboard while dragging. Use Ctrl-Key to copy instead of move - but this is not handled in macOS, so default for macOS is copy always.

KaiZin2018:
Long time ago, I implemented as well file dropping function to start opening or merging.
Also some more corrections included.

src/ui/Windows/PWTreeCtrl.cpp
```diff
@@ -2087,7 +2087,7 @@ bool CPWTreeCtrlX::ProcessData(BYTE *in_buffer, const long &inLen,
   inDDmemfile.Attach((BYTE *)in_buffer, inLen);
 
   // Get all the entries
-  in_oblist.DDUnSerialize(inDDmemfile);
+  in_oblist.DD/ialize(inDDmemfile);
```
ronys, line 2090: This broke the Windows build. Fixing it locally.

KaiZin2018:
Sorry for the problem in windows built - I toke the implementation as source for the wxWidget built. There had been a typo into the wrong window...

ronys:
Can you rebase to master and fix the conflicts?

KaiZin2018:
Sorry, KaiZin2018/pwSafe looks already up to date: "This branch is 17 commits ahead of pwsafe:master.". I do not know what to do for you. There is no conflict, as we did the same change in 3295bc7 and 5c214db.

KaiZin2018:
Hello Ronys, you can try to overtake the change in core to avoid useless questions about alias/shortcut to be deleted when call Delete() for a base-entry, by following patch (after merging the above source). I do not have Windows development environment to double check if side effects are present.
--- src/ui/Windows/MainEdit.cpp
+++ src/ui/Windows/MainEdit.cpp
@@ -1141,7 +1141,7 @@ Command *DboxMain::Delete(const CItemData *pci)
   // ConfirmDelete asks for user confirmation
   // when deleting a shortcut or alias base.
   // Otherwise it just returns true
-  if (m_core.ConfirmDelete(pci))
+  if (m_core.ConfirmDelete(pci, m_sxOriginalGroup)) // Do not ask for confirmation on alias or shortcut that is part of the group to be deleted, therefore add the group to be deleted here as second parameter
     return DeleteEntryCommand::Create(&m_core, *pci);
   else
     return NULL;

Issue #721, opened 2021-02-26
gtest: allow to be provided instead of download and build on-fly


yopito:
While packaging release 1.13 for VoidLinux distribution, I've enabled the self unit tests provided.
Since `gtest` is available as a package, I've disabled its download and build on-fly with this quick patch [1].
It's working fine (if providing gtest of course)

Request: provide a cmake flag to use provided gtest instead of download and build an-fly.
Something like `-DGTEST_USE_INTERNAL` for instance.

What do you think about ? 
I might provide a PR if needed.

[1] https://github.com/void-linux/void-packages/blob/master/srcpkgs/passwordsafe/patches/use-gtest-from-package.patch

ronys:
I've had trouble with packaged `gtest` in the past, which is why I went this route.

It would be nice to skip the download & build at make time, and instead do the work in `Misc/setup-linux-dev-env.sh` - either installing the `gtest` package or download and build it once while running this script.

A PR with this would be most welcome!

yopito:
I prefer to add an additional cmake's option to circumvent it, since the current way to do is already fine.
As a packager, I don't use `Misc/setup-linux-dev-env.sh`, so I feel less comfortable to move gtest stuff into it.

PR #722, opened 2021-02-28
add cmake option GTEST_BUILD to control gtest download and build on-fly

yopito:
Closes: #721

Is working as expected.


ronys:
Thanks!

PR #724, opened 2021-03-07
[wx] Fix compiler warnings about comparison of integer expressions

rafaelx:


KaiZin2018 (commented):
I personal prefer static_cast<...> instead of constructor on the given file, but this is something of personal taste. If compiler warning is given in environment it is good to implement the changes.

KaiZin2018 (approved):

Issue #734, opened 2021-03-25
Password expiry warning doesn't work [PasswordSafe Linux 1.13]


madmartin:
Hi,
since some time I changed from the pwsafe windows version (used on Linux with Wine) to the linux version, started with 1.11.
Now I discover that the warning about expired paswords does not work. 
The warning option is on, but all the passwords expiring within the time range (or already have expired) keep their green rectangle instead of turning yellow or red.
This feature has worked with the windows version I used previously.
The database version of the PW DB is 3.14.

I will try to help finding that in the source code, can someone give me some guidance where I find related code in the source?

BTW I am the maintainer of passwordsafe on Gentoo Linux. see  [here](https://packages.gentoo.org/packages/app-admin/passwordsafe)

ronys:
The problem is that this feature hasn't been implemented yet in the wx port, so it's a matter of porting the Windows implementation of CheckExpiredPasswords(), as well as the logic in the Tree control that colors them correctly.

Volunteers?

madmartin:
I can have a look, but dont give me more than 10% success rate....

KaiZin2018:
You might double check Windows
 int CCWTreeCtrl::GetEntryImage(const CItemData &ci) const
and wxWidgets
 void TreeCtrl::SetItemImage(const wxTreeItemId &node, const CItemData &item)
Be aware about ordering of the icons in xpmList.

May be I can help next week.

madmartin:
@KaiZin2018  and @ronys  - I tested your patch 65e4ea68e5 - it seems to work on my database - excellent!!!! 
Can you please arrange that this will be merged into the next release?

ronys:
Merged, will be in next release.

PR #735, opened 2021-03-26
Implement issue #734

KaiZin2018:
Show different Item icon in tree view for entries that will be shortly expired (warning) or are expired.

### DIFF
--- a/src/ui/wxWidgets/TreeCtrl.cpp
+++ b/src/ui/wxWidgets/TreeCtrl.cpp
@@ -83,6 +83,7 @@ IMPLEMENT_CLASS( TreeCtrl, wxTreeCtrl )
 
 // Image Indices - these match the order images are added
 //                 in TreeCtrl::CreateControls()
+// Warning: Do not change ordering on ..._EXP_II, ..._WARN_II and base entry for ABASE, NORMAL and SBASE
 enum {
   ABASE_EXP_II,    // 0
   ABASE_WARN_II,   // 1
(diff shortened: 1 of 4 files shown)


src/ui/wxWidgets/TreeCtrl.cpp
```diff
@@ -838,14 +839,50 @@ bool TreeCtrl::Remove(const CUUID &uuid)
 void TreeCtrl::SetItemImage(const wxTreeItemId &node,
                                const CItemData &item)
 {
-  // TODO: modify to display warning and expired states
+  int offset = 0;
+  CItemData::EntryType entrytype = item.GetEntryType();
+  
+  if((entrytype == CItemData::ET_NORMAL) || (entrytype == CItemData::ET_ALIASBASE) || (entrytype == CItemData::ET_SHORTCUTBASE)) {
+    time_t tttXTime(static_cast<time_t>(0));
```
dskrvk, line 846: Minor: the method names GetXTime, GetPMTime, and GetCTime are not very intuitive and make the calculations here a bit hard to follow. Maybe we could at least name the corresponding local variables differently - e.g. `daysToExpiration` & `creationModificationTimestamp`?

src/ui/wxWidgets/TreeCtrl.cpp
```diff
@@ -838,14 +839,50 @@ bool TreeCtrl::Remove(const CUUID &uuid)
...
+    item.GetXTime(tttXTime);
+    if (tttXTime > static_cast<time_t>(0) && tttXTime <= static_cast<time_t>(3650)) {
+      time_t tttCPMTime(static_cast<time_t>(0));
+      item.GetPMTime(tttCPMTime);
+      if (tttCPMTime == static_cast<time_t>(0))
+        item.GetCTime(tttCPMTime);
+      tttXTime = static_cast<time_t>((long)tttCPMTime + (long)tttXTime * 86400L);
```
dskrvk, line 853: What's happening on this line? Is it trying to correct an invalid tttXTime?

src/ui/wxWidgets/TreeCtrl.cpp
KaiZin2018: I added comments for those who are not so familiar with the short form of field names.

src/ui/wxWidgets/TreeCtrl.cpp
KaiZin2018: Add comment about handling of time versus handling of distance of days before expiry.

KaiZin2018:
The change in TreeCtrl::Remove() had lead to left wrongly empty groups when undo tree Drag and Drop. Recursive deletion as done before would have removed empty groups when present before empty groups. With additional change the group structure in tree view is left correct when undo tree Drag and Drop action.

ronys:
Nice work - thanks!

PR #738, opened 2021-04-20
Correct copy and paste failure in PWFiltersDlg.cpp

KaiZin2018:
Number of Policies is given instead of those for Attachments.

### DIFF
--- a/src/ui/wxWidgets/TreeCtrl.cpp
+++ b/src/ui/wxWidgets/TreeCtrl.cpp
@@ -83,6 +83,7 @@ IMPLEMENT_CLASS( TreeCtrl, wxTreeCtrl )
 
 // Image Indices - these match the order images are added
 //                 in TreeCtrl::CreateControls()
+// Warning: Do not change ordering on ..._EXP_II, ..._WARN_II and base entry for ABASE, NORMAL and SBASE
 enum {
   ABASE_EXP_II,    // 0
   ABASE_WARN_II,   // 1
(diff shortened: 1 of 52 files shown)


src/ui/wxWidgets/TreeCtrl.cpp
```diff
@@ -703,6 +718,8 @@ void TreeCtrl::AddItem(const CItemData &item)
   wxTreeItemId gnode = AddGroup(GroupNameOfItem(item));
   const wxString disp = ItemDisplayString(item);
   wxTreeItemId titem = AppendItem(gnode, disp, -1, -1, data);
+  if(GetRootItem() != gnode && GetItemImage(gnode) == EMPTY_NODE_II)
+    wxTreeCtrl::SetItemImage(gnode, NODE_II); // Could be empty before
```
rafaelx: Seems to have the same purpose as the code on line 519 and 520. How about a little helper function to avoid code duplication?

src/ui/wxWidgets/TreeCtrl.cpp
```diff
@@ -820,14 +837,9 @@ bool TreeCtrl::Remove(const CUUID &uuid)
...
     wxTreeItemId parentId = GetItemParent(id);
     Delete(id);
-    while (parentId != GetRootItem()) {
-      wxTreeItemId grandparentId = GetItemParent(parentId);
-      if (GetChildrenCount(parentId) == 0) {
-        Delete(parentId);
-        parentId = grandparentId;
-      } else
-        break;
-    } // while
```
rafaelx: What happened to the tree traversing loop? Is it no longer needed?

KaiZin2018: The tree traversing loop is no longer needed. It is faulty in some cases and therefore had been corrected (removed). In case an element is removed, e.g.
+ -
|_ A group
      |_ B group
            |_ C entry
The recursive deletion would remove group A and B from the tree view in case C entry is deleted. Internally the group A and B are still present and will come back when tree view is rebuild (e.g. after coming back from grid view).

src/ui/wxWidgets/TreeCtrl.cpp
KaiZin2018: Demanded change is handled in "Use Function to set EMPTY_NODE_II or set back to NODE_II."

KaiZin2018: This change is already in master repository. I don‘t know what is wrong with my local repository. The last changes / commit are repeated... and k do not know how to solve this problem except setting back the repository to master. But next time the story is repeated.

PR #742, opened 2021-05-17
Add "StartupWMClass=pwsafe" to passwordsafe.desktop

Amr-Ibra:
The app icon is missing from the dash, the top panel and the window switcher in GNOME Shell under Wayland.

This is solved by adding "StartupWMClass=pwsafe" to passwordsafe.desktop to enable GNOME Shell to map the application to its icon under Wayland.

Closes https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=988062


ronys:
Thanks!

PR #743, opened 2021-05-17
Update macOS build documentation and procedure to cover Apple Silicon…

dhocker:
… (M1)

build target.


ronys:
Nice work - thanks!

PR #746, opened 2021-06-02
Add missing function Manage filter and New/Edit filter

KaiZin2018:
This pull request is adding the manage filter and new/edit filter functionality as in Windows environment given.
The DCA filter is extend with PRESENT/NOTPRESENT, with small adaptation in Windows only.
The filter XSD is corrected to handle all filters as possible.


KaiZin2018:
Mac build failed last time with resource busy, but is working (also on test build - see previous time some minutes ago).

ronys:
Nice work! Many thanks.

PR #749, opened 2021-06-18
Implement XML import for wxWidgets using Pugi library

KaiZin2018:
Without need of Xerces or Microsoft XML library the added code is supporting XML file import including full syntax check, without using xsd file definition.


ronys:
So you've basically hand-coded a schema validator for XML. Impressive!
Future developers/maintainers need to know that this should be updated as well as the xsd file whenever the schema changes. Can you document this somewhere?
This is also a good occasion to add unit tests for XML operations. Let me know if you'd like to do these as well.

KaiZin2018:
> So you've basically hand-coded a schema validator for XML. Impressive!
> Future developers/maintainers need to know that this should be updated as well as the xsd file whenever the schema changes. Can you document this somewhere?
> This is also a good occasion to add unit tests for XML operations. Let me know if you'd like to do these as well.

Thank you Rony, I have now add a README.XML.md to address your demand on documentation on the parallel syntax definition of XML structures. I'm not aware of the goal and way of writing unit tests. Actual I will focus more on extending the functions to the same level as Windows has already reached. Next on my agenda is
- dealing with aliases in wxWidgets
- Manage Password Policies in wxWidgets (Check function, as I encountered problems with version 3.1.5 of wxWidgets)

ronys:
1. Looks like there's a typo in src/core/CMakeLists.txt:75 - "NOT" shouldn't be there ( "list (NOT APPEND CORE_SRCS)")
2. There's a conflict with the latest src/ui/wxWidgets/I18N/pos/pwsafe_ru.po - My Russian isn't good enough to resolve this...

KaiZin2018:
> * Looks like there's a typo in src/core/CMakeLists.txt:75 - "NOT" shouldn't be there ( "list (NOT APPEND CORE_SRCS)")

NOT removed from file

> * There's a conflict with the latest src/ui/wxWidgets/I18N/pos/pwsafe_ru.po - My Russian isn't good enough to resolve this...

Overtake last change from Master and re-run local make pos - new texts will present for translation...

PR #752, opened 2021-06-27
Option to show full password in password subset dialog

pm-kan:
Sometimes full password should be entered manually (on some "remote desktops" or when entering password on another device), so I've added `*` char to subset dialog to show full password.

Not sure, maybe it's better to use another char for full password substitution, or use distinct context menu item to show full password.

Also with few changes for Linux build


src/ui/Windows/PasswordSubsetDlg.cpp
```diff
@@ -85,8 +87,12 @@ void CNumEdit::OnChar(UINT nChar, UINT nRepCnt, UINT nFlags)
   if (m_bLastSeparator && (nChar == L'0' || bSeparator))
     return;
 
+  // Ignore wildcard if string is not empty, or last char is not separator
+  if (bWildcard && !m_bLastSeparator && LineLength(LineIndex(0)) != 0)
+    return;
+  
   // Do not pass on any character that is not a digit, minus sign, separator or backspace
```
dskrvk, line 94: Here's an example of a comment that simply restates the code in English and gets outdated when the code changes. Such comments have negative value IMO because they don't really clarify anything and also need to be maintained. Lines 112 & 118 are even better examples.

Just an observation, but personally I'd have removed such comments in this function rather than updating them.

ronys:
Well, a bit redundant, since there are several ways to copy the whole password, but still makes sense for completeness.
Thanks for implementing both in Windows and wx.

PR #753, opened 2021-06-30
Update language files

KaiZin2018:


KaiZin2018:
Sorry for incorrect tittle - update language files and small correction on base entry handling.

PR #756, opened 2021-07-16
Handle Alias in wxWidgets environment

KaiZin2018:
The implementation of Alias in wxWidgets environment had not been completed so far. This pull request is adding full support for alias in a user friendly way, providing a dialog for alias selection. In addition the information on alias is clearly visible on New/Add Entry Dialog by changing the label in front of password. On selecting a shortcut base or shortcut as alias or alias base the related shortcuts are removed and warning message is given before the action takes place.

### DIFF
--- a/Makefile.windows
+++ b/Makefile.windows
@@ -157,7 +157,7 @@ endif
 		$(GH_UPLOAD_CMD) upload -t $(RELEASENAME) -n $(BINRELNAME64).zip -f $(BINRELNAME64).zip; \
 		$(GH_UPLOAD_CMD) upload -t $(RELEASENAME) -n $(BINRELNAME64).zip.sig -f $(BINRELNAME64).zip.sig; \
 		$(GH_UPLOAD_CMD) upload -t $(RELEASENAME) -n pwsafe-$(RELEASENAME)-map+pdb.zip \
-			-f pwsafe-$(RELEASENAME)-map+pdb.zip) \
+			-f pwsafe-$(RELEASENAME)-map+pdb.zip; \
 		$(GH_UPLOAD_CMD) upload -t $(RELEASENAME) -n pwsafe64-$(RELEASENAME)-map+pdb.zip \
 			-f pwsafe64-$(RELEASENAME)-map+pdb.zip)
 	(cd docs; $(GH_UPLOAD_CMD) upload -t $(RELEASENAME) -n ReleaseNotes.md -f ReleaseNotes.md)
--- a/Misc/I18N/ru/omegat_wx/omegat/project_save.tmx
+++ b/Misc/I18N/ru/omegat_wx/omegat/project_save.tmx
@@ -5085,6 +5085,14 @@ Action:</seg>
     </tu>
     <tu>
       <tuv lang="en-US">
+        <seg>Enter positions of password characters separated by spaces, commas or semi-colons (or * for full password).</seg>
+      </tuv>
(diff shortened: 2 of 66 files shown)


ronys:
Nice work in terms of code.

The UI, however, is problematic, in that it exposes the implementation detail of aliases being an "encoded" password to the user, who really shouldn't know or care about this.
Worse, the average uses, who will probably never need the alias functionality, will be pretty confused by the "alias" button appearing around the password controls.

In Windows this functionality is deliberately hidden, available only to people who bother to read the documentation, exactly because we didn't find a way to present it in a simple manner to users.

Perhaps moving the Alias button to the "Additional tab with a tooltip text  "Alias to an existing Entry" would be less confusing.

Finally, when a button results in a new window appearing, the convention is that it ends with "..." I suggest "Alias" be renamed to "Alias To..."

KaiZin2018:
> Nice work in terms of code.

Thank you.

> 
> The UI, however, is problematic, in that it exposes the implementation detail of aliases being an "encoded" password to the user, who really shouldn't know or care about this.

Internal implementation is by linked UUID, but user representation is by the shown encoded password.

> Worse, the average uses, who will probably never need the alias functionality, will be pretty confused by the "alias" button appearing around the password controls.

Hide the button until selection on new boolean "ShowAliasSelection" in options menu.

> 
> In Windows this functionality is deliberately hidden, available only to people who bother to read the documentation, exactly because we didn't find a way to present it in a simple manner to users.

The possible typo error by manual typing is avoided when selection is done in the new dialog. I would propose to add a similar dialog also for Windows.

> 
> Perhaps moving the Alias button to the "Additional tab with a tooltip text "Alias to an existing Entry" would be less confusing.

See new option, so the button gets visible in case of switching this new option on.

> 
> Finally, when a button results in a new window appearing, the convention is that it ends with "..." I suggest "Alias" be renamed to "Alias To..."

Ok, Name is changed to "Alias To..."

ronys:
*Much* better!
A few more tweaks and it's ready to be merged:
1. Setting an Alias to an existing entry removes that entry's password. This is obvious when making an alias the old way, by explicitly editing the password field, but not at all clear to someone who's trying out the feature. I think a confirmation popup "This will replace the current password with that of the aliased entry" would make sense here.
2. The "Remove" button replaces an alias [g:t:u] password with [Alias], which I don't find very useful. Clearing the password entirely or restoring it to the non-alias password would make more sense.
3. Likewise, disabling Remove when the current value of the password field isn't a [g:t:u] would make it less confusing.

Let me know if you'd prefer I do one or more of the above.

KaiZin2018:
> _Much_ better!
> A few more tweaks and it's ready to be merged:
> 
> 1. Setting an Alias to an existing entry removes that entry's password. This is obvious when making an alias the old way, by explicitly editing the password field, but not at all clear to someone who's trying out the feature. I think a confirmation popup "This will replace the current password with that of the aliased entry" would make sense here.

Warning not needed, as the old password will stay unchanged.

> 2. The "Remove" button replaces an alias [g:t:u] password with [Alias], which I don't find very useful. Clearing the password entirely or restoring it to the non-alias password would make more sense.

Overwriting with [Alias] is removed. The old password stay stored.

> 3. Likewise, disabling Remove when the current value of the password field isn't a [g:t:u] would make it less confusing.

Handling of Remove button is done as demanded.

> Let me know if you'd prefer I do one or more of the above.

Everything done. Thanks for the offer.

ronys:
Excellent! I still manage to get "[Alias]" in the password field, but this looks like it also happened previously.

Thanks once again for your work and patience.

PR #761, opened 2021-09-07
Hotfix to avoid assert with no delimiter configured in menu bar, enable compilation for MacOS using Xcode

KaiZin2018:
- Avoid assert with no delimiter setting in display options
- Set/unset text beneath icons in menu bar according to display options
- Enable compilation/linking on MacOS using Xcode as build environment
Hint: wxWidgets 3.0.5 is not able to run with the actual development on MacOS, while 3.1.4 is able, but background of tree view is grey and not white as the icons background is - the last one is fixed with 3.1.5 development branch of wxWidgets.

### DIFF
--- a/src/core/PWSprefs.cpp
+++ b/src/core/PWSprefs.cpp
@@ -134,7 +134,7 @@ const PWSprefs::boolPref PWSprefs::m_bool_prefs[NumBoolPrefs] = {
   {_T("EnableWindowTransparency"), false, ptApplication},   // application
   {_T("ShowMenuSeparator"), true, ptApplication},           // application
   {_T("AutoAdjColWidth"), false, ptApplication},            // application
-  {_T("ToolbarShowText"), false, ptApplication},            // application
+  {_T("ToolbarShowText"), true, ptApplication},             // application
   {_T("DragAndDropShowRoot"), false, ptApplication},        // application
   {_T("ShowAliasSelection"), false, ptApplication},         // application
 };
--- a/src/ui/wxWidgets/PasswordSafeFrame.cpp
+++ b/src/ui/wxWidgets/PasswordSafeFrame.cpp
@@ -890,21 +890,35 @@ void PasswordSafeFrame::RefreshToolbarButtons()
   if (toolbar->GetToolCount() > 0) {
     toolbar->ClearTools();
   }
+  bool toolbarShowText = pref->GetPref(PWSprefs::ToolbarShowText);
+  bool showMenuSeprator = pref->GetPref(PWSprefs::ShowMenuSeparator);
 
   for (const auto & PwsToolbarButton : PwsToolbarButtons) {
-    if (pref->GetPref(PWSprefs::ShowMenuSeparator) && (PwsToolbarButton.IsSeparator())) {
-      toolbar->AddSeparator()->SetId(PwsToolbarButton.id);
+    if (PwsToolbarButton.IsSeparator()) {
+      if (showMenuSeprator)
+        toolbar->AddSeparator()->SetId(PwsToolbarButton.id);
     }
     else {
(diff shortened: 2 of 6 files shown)


src/ui/wxWidgets/PasswordSafeFrame.cpp
```diff
@@ -890,21 +890,35 @@ void PasswordSafeFrame::RefreshToolbarButtons()
...
     toolbar->ClearTools();
   }
+  bool toolbarShowText = pref->GetPref(PWSprefs::ToolbarShowText);
+  bool showMenuSeprator = pref->GetPref(PWSprefs::ShowMenuSeparator);
 
   for (const auto & PwsToolbarButton : PwsToolbarButtons) {
-    if (pref->GetPref(PWSprefs::ShowMenuSeparator) && (PwsToolbarButton.IsSeparator())) {
-      toolbar->AddSeparator()->SetId(PwsToolbarButton.id);
+    if (PwsToolbarButton.IsSeparator()) {
+      if (showMenuSeprator)
```
dskrvk, line 898: Why not combine the two if conditions?

src/ui/wxWidgets/PasswordSafeFrame.cpp
```diff
@@ -890,21 +890,35 @@ void PasswordSafeFrame::RefreshToolbarButtons()
...
+          wxEmptyString,
+          nullptr);
+      else
+        toolbar->AddTool(
+          PwsToolbarButton.id,
+          wxGetTranslation(PwsToolbarButton.toollabel),
+          PwsToolbarButton.GetBitmapForEnabledButton(),
+          PwsToolbarButton.GetBitmapForDisabledButton(),
+          wxITEM_NORMAL,
+          wxEmptyString,
```
dskrvk, line 919: This is the only line that's different, why not use a ternary operator here?
```suggestion
          toolbarShowText ? wxGetTranslation(PwsToolbarButton.tooltip) : wxEmptyString,
```

src/ui/wxWidgets/PasswordSafeFrame.cpp
KaiZin2018: Because of the logic, the following else depends only from type of entry. In case of combination an assert is the result of execution in case of configuration to not showing separation.

src/ui/wxWidgets/PasswordSafeFrame.cpp
KaiZin2018: Because compiler error about ambiguous return value of wxTranslate.

ronys:
Thanks!

rafaelx:
Strange that I can't reproduce the assert.  :-|

src/core/PWSprefs.cpp
```diff
@@ -135,7 +135,7 @@ const PWSprefs::boolPref PWSprefs::m_bool_prefs[NumBoolPrefs] = {
   {_T("EnableWindowTransparency"), false, ptApplication},   // application
   {_T("ShowMenuSeparator"), true, ptApplication},           // application
   {_T("AutoAdjColWidth"), false, ptApplication},            // application
-  {_T("ToolbarShowText"), false, ptApplication},            // application
+  {_T("ToolbarShowText"), true, ptApplication},             // application
```
rafaelx, line 138: I prefer to see the labels on the toolbar buttons, but is there a technical reason to change the default behavior?

src/ui/wxWidgets/PasswordSafeFrame.cpp
rafaelx: If someone doesn't know that this is a hotfix and stumbles upon it, then I wouldn't be surprised if they change that again as part of a code clean-up. Smells like code duplication.  8^)

PR #763, opened 2021-09-12
[wx] Fix size of search status text

rafaelx:


rafaelx:
When the search bar appears, it initially has a blank text for the search status text. No status text is displayed during a search and only appears if the size of the main frame is changed slightly.
This commit is intended to fix this.

PR #767, opened 2021-09-27
Add German translations (in wxWidgets)

KaiZin2018:


ronys:
I only see the pwsafe_de.po file in the commit. Perhaps you forgot to add the Makefile?
(Note that in general  it's a good idea to keep unrelated changes in separate commits)

KaiZin2018:
The makefile is not needing any change, as only translations are added. The present makefile in above directory is sufficient.

ronys:
OK. The commit message confused me.

PR #773, opened 2021-10-03
Add missing copy for languages in Makefile

KaiZin2018:
Correct generation of Mac OS package dmg file


ronys:
Instead of dealing with the conflicts, can you create a new pull request with just the Makefile changes?

PR #805, opened 2021-11-22
[cpack] Update definition of CPACK_PACKAGE_FILE_NAME

jrolli:
This PR makes two changes to how the CPack variable `CPACK_PACKAGE_FILE_NAME` is defined - neither change should affect the definition for `dpkg`-based systems.

The first change simply removes the architecture tag from the file name when `dpkg` is not present (rather than adding "i386" regardless of how the package was actually built).

The second change fixes an unsanitized use of the `NAME` variable from os-release which is permitted to contain whitespace.  This led CPack to interpret the filename as a list on Fedora 35 (value is "Fedora Linux") and error out during RPM generation.  This corrects that behavior by using `ID` from os-release which is guaranteed by spec (`man 5 os-release`) to contain no whitespace.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -495,11 +495,12 @@ if (NOT WIN32)
 
       if (NOT DPKG_EXECUTABLE)
         # if we can't find dpkg, we don't really care, as this variable won't be used in this run.
-        set (CPACK_DEBIAN_PACKAGE_ARCHITECTURE i386)
+        set (PACKAGE_ARCHITECTURE "")
       else ()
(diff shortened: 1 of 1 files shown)


ronys:
Thanks - can you fix the syntax error?

CMake Error at CMakeLists.txt:503 (set):
  Syntax error in cmake code at

    /home/runner/work/pwsafe/pwsafe/CMakeLists.txt:503

  when parsing string

    -${CPACK_DEBIAN_PACKAGE_ARCHITECTURE

  syntax error, unexpected end of file, expecting } (37)
-- Configuring incomplete, errors occurred!

PR #817, opened 2021-12-10
fix small typo in english source text

Karneades:


dskrvk (approved):

ronys (commented):
This is not a typo. "e.g." is an abbreviation for "exempli gratia", meaning "for example". Following it with a comma is correct. Read it as "...(for example, time)..."

ronys:
See review comment

PR #819, opened 2021-12-10
[wx] Update to English text in SelectionCriteria

Karneades:
During translation updates for German at this line (separate PR), the following strange English translation came up. Maybe in the past it only stated "were:" but now the operation is also given. I would replace the "were" with _from_ <operation> or with _with_.

Current report looks like this where the text is used
```
...
All entries were Find with corresponding entries from "/path/test.psafe3"

All fields in matching entries were Find
...
```


Karneades:
Hm, this _were_ is also used with a verb besides the case with a noun from the first comment. So we have once "were compared" and once "were Find ...". The operation which is given as placeholder in the code is either a verb or a noun.
```
All Enties were compared...
```

ronys (changes requested):
This function is called with the operation parameter set to one of the following:
- compared
- Find
- synchronized

Seems to me that the correct fix is to change the call with "Find" to "found", and leave "were" as-is.

Karneades:
Changed it accordingly.

PR #831, opened 2022-01-30
[wx] Fix file path name for Yubi

rafaelx:
This is an attempt to fix the reported issue [1548](https://sourceforge.net/p/passwordsafe/bugs/1548/) on SourceForge.
It's untested because I don't have a Yubikey.

### DIFF
--- a/src/ui/wxWidgets/SafeCombinationEntryDlg.cpp
+++ b/src/ui/wxWidgets/SafeCombinationEntryDlg.cpp
@@ -633,6 +633,9 @@ void SafeCombinationEntryDlg::OnYubibtnClick(wxCommandEvent& WXUNUSED(event))
 {
   m_combinationEntry->AllowEmptyCombinationOnce();  // Allow blank password when Yubi's used
 
+  // For the validation process, put the full file path name back into the combo box.
+  m_filenameCB->ChangeValue(m_filename);
+
   if (Validate() && TransferDataFromWindow()) {
     if (!pws_os::FileExists(tostdstring(m_filename))) {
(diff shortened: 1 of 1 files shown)


b-schliessmann (approved):
Built and tested with my Yubikey 5, it solves the "file or path not found" problem (that pswafe uses the ellipsised filename instead of the "real" one to try and open the password database file).

(This issue, however, raises doubts with me if it wouldn't be prudent to remove the ellipsising code altogether instead. To me, personally, this feature is of questionable merit.)

rafaelx:
> Built and tested with my Yubikey 5, it solves the "file or path not found" problem (that pwsafe uses the ellipsised filename instead of the "real" one to try and open the password database file).
> 
Thanks for testing.  :)

> (This issue, however, raises doubts with me if it wouldn't be prudent to remove the ellipsising code altogether instead. To me, personally, this feature is of questionable merit.)

For some, this feature is of questionable value and for [others](https://sourceforge.net/p/passwordsafe/discussion/134801/thread/1dc4628928/?limit=25#402f), it makes the application (hopefully) easier to use.  ;-)

b-schliessmann:
You're welcome! :) And thank _you_ for the quick fix.

Ah well, the usual YMMV disclaimer applies ;) IMHO, it was unfortunate that a fix for a user's "luxury problem" produced a show stopper for others.

ronys:
Showing long paths with ellipsis in the middle is common UI these days, I wouldn't consider this "luxury".

In any case, thanks to you both for the report, fix & test.

PR #845, opened 2022-05-16
Read-only checkbox comes first in the Safe Combination Entry dialog

alexis-m:
Somewhere between v3.45 and v3.58 Show Combination was placed first.
This is less convenient for keyboard-only input, as the read-only
box is used more often.

Use case: enter combination, press Tab to focus the read-only box,
press Space to check it, press Enter to continue.

Open dialog before
image

Open dialog after
image

Unlock dialog with RO box hidden
image


ronys:
Nice - thanks!

PR #846, opened 2022-05-17
[SF1553] Restore Unicode awareness in add-edit dialogs

alexis-m:
Commit 35eae8d (accidentally?) changed these two dialogs from
RichEdit20W to RichEdit20A in v3.58.

First, this made double-click word selection glitchy as described in
SourceForge bug 1553. Second, notes with ASCII and non-ASCII text
(if created in a previous version) don't show correctly in v3.58 --
the text is truncated.


ronys:
Thanks once again.

PR #854, opened 2022-06-06
[SF1545] Restore maximized status of the window on program start

alexis-m:
Since commit 103134a6e, the maximized window is saved in the prefs
as an empty rect (0, 0, 0, 0). So do the maximized check first.
Previously, the visibility check (RectInRegion) would always win and
ignore the empty rect thus not restoring the maximized state.

The fix is not complete. The following situation is still not right:
1. Place Window in the middle of the screen.
   Normally, if you close the program and open it again, the window
   is shown in the same place.
2. Click Maximize button and close the program.
   Windows position is saved as (0, 0, 0, 0).
3. Open the program again. Window is shown as Maximized.
   This didn't work. This commit fixes it.
4. Click Restore button.
   Window is restored to the top-left corner with some default size.
   Expectation: Windows is restored to its previous state from 1.


dskrvk (approved):

ronys:
Thanks!

Issue #878, opened 2022-09-02
macOS: The main window does not disappear when the close button is clicked


nobugshere:
On macOS, when I click the red close button on the main window (with the menu bar icon enabled) the content of the window vanishes but the window frame remains.  This does not happen with Cmd-H or the yellow minimize button.  The attached patch fixed the problem for me.  There's a comment in the code that I don't really understand:
    //Don't call Hide() here, which just calls Show(false), which is overridden in
    //derived classes, and wxDialog actually cancels the modal loop and closes the window

However, it seems we need to let it call the derived class function.

Environment:
M1 Mac Studio macOS 12.5.1
Xcode 13.4.1
PasswordSafe 1.15 pulled from git master
wxWidgets 3.2.0 (installed via homebrew)

```
diff --git a/src/ui/wxWidgets/PasswordSafeFrame.cpp b/src/ui/wxWidgets/PasswordSafeFrame.cpp
index 2f08a5093..c2a6e7994 100644
--- a/src/ui/wxWidgets/PasswordSafeFrame.cpp
+++ b/src/ui/wxWidgets/PasswordSafeFrame.cpp
@@ -3100,7 +3100,7 @@ void PasswordSafeFrame::HideTopLevelWindows()
     //derived classes, and wxDialog actually cancels the modal loop and closes the window
     pws_os::Trace(L"Hide window <%ls> (%ls)\n", ToStr(win->GetTitle()), ToStr(win->GetName()));
     win->ClearBackground();
-    win->wxWindow::Show(false);
+    win->Show(false);
   }
   m_hiddenWindows = std::move(windows);
 }
@@ -3109,7 +3109,7 @@ void PasswordSafeFrame::ShowHiddenWindows(bool raise)
 {
   for(const auto win : m_hiddenWindows) {
     pws_os::Trace(L"Show window <%ls>\n", ToStr(win->GetTitle()));
-    win->wxWindow::Show(true);
+    win->Show(true);
     if (raise) {
       win->Raise();
     }
```

ronys:
Patch tested against other builds, merged in #880 
Thanks!

PR #880, opened 2022-09-14
BR878: macOS UX fix

ronys:
(Can't test this directly, but doesn't seem to break Debian/Fedora/Ubuntu.)

See #878 for details.


rafaelx:
Didn't notice anything unusual with Manjaro, Kubuntu (GTK3) and openSUSE (GTK2) either.

PR #886, opened 2022-10-03
Avoid error message "Dots are not allowed in group names" with Read O…

KaiZin2018:
…nly data base. I do not know how, but sometimes I can clock in the list view and this is opening the editing of the entry label. In case this label is including a dot by chance the before given error is shown. To avoid entering the real edit dialog and the error message about a dot I like to include the Read Only condition into the initial condition with veto() without error message.

### DIFF
--- a/src/ui/wxWidgets/TreeCtrl.cpp
+++ b/src/ui/wxWidgets/TreeCtrl.cpp
@@ -1016,8 +1016,9 @@ void TreeCtrl::OnEndLabelEdit( wxTreeEvent& evt )
 {
   const wxString &label = (evt.IsEditCancelled() ? GetItemText(evt.GetItem()) : evt.GetLabel());
 
-  if (label.empty()) {
-    // empty entry or group names are a non-no...
+  if (label.empty() || IsReadOnly()) { // end editing without message to avoid error "Dots are not allowed in group names" with Read Only data base and a dot in a shown name
+    // 1st empty entry or group names are a non-no...
+    // 2nd In Read-Only data base we do not change anything
     evt.Veto();
     return;
   }


src/ui/wxWidgets/TreeCtrl.cpp
```diff
@@ -1012,12 +1013,23 @@ void TreeCtrl::OnRenameGroup(wxCommandEvent& WXUNUSED(evt))
   }
 }
 
+void TreeCtrl::OnStartLabelEdit( wxTreeEvent& evt )
+{
+  wxTreeItemId item = evt.GetItem();
+  if(IsReadOnly() || !item.IsOk() ||! ItemIsGroup(item)) {
```
ronys, line 1019: Why the check for !ItemIsGroup() ? We allow editing a group's name same as for an entry.

src/ui/wxWidgets/TreeCtrl.cpp
```diff
@@ -1012,12 +1013,23 @@ void TreeCtrl::OnRenameGroup(wxCommandEvent& WXUNUSED(evt))
...
+  }
+}
+
 void TreeCtrl::OnEndLabelEdit( wxTreeEvent& evt )
 {
   const wxString &label = (evt.IsEditCancelled() ? GetItemText(evt.GetItem()) : evt.GetLabel());
 
-  if (label.empty()) {
-    // empty entry or group names are a non-no...
+  if (label.empty() || IsReadOnly()) { // end editing without message to avoid error "Dots are not allowed in group names" with Read Only data base and a dot in a shown name
```
ronys, line 1030: The check for IsReadOnly() is unnecessary and confusing, since this is checked in OnStartLabelEdit().

ronys (changes requested):

src/ui/wxWidgets/TreeCtrl.cpp
KaiZin2018: It is a nice idea to allow editing of entry names in the list view. But for that additional handling must be added into TreeCtrl::OnStartLabelEdit() and TreeCtrl::OnEndLabelEdit(). In actual handling go TreeCtrl::OnEndLabelEdit() only renaming or adding of group is implemented (by FinishAddingGroup() or FinishRenamingGroup()). In addition in TreeCtrl::OnStartLabelEdit() the optional shown additional fields in the tree view label must be removed before editing of the name. otherwise the entry <entry> with username [xyz@abc.de] would occur as "entry [xyz@abc.de]" for editing - only entry has to be content while on place editing is active. So change the label to the entry name only in TreeCtrl::OnStartLabelEdit() and rebuild to the resulting entry, including additional shown part, in TreeCtrl::OnEndLabelEdit(). If you like to allow editing of the sub-fields as well the parser would become much more complex...

src/ui/wxWidgets/TreeCtrl.cpp
KaiZin2018: Ok, removed in last check in to handle in TreeCtrl::OnStartLabelEdit() only.

src/ui/wxWidgets/TreeCtrl.cpp
ronys: IIRC, the Windows version does this extra parsing, but in any case this is way beyond the scope of this PR.

ronys (approved):

Issue #888, opened 2022-10-11
TreeCtrl editing is not really implemented


nobugshere:
A recent patch tried to prevent the direct editing of the name of Tree View password entries because the code is not there to support it.  However, group names can still be selected for editing, even though it doesn't work properly.  This patch completely disables direct editing of names in the TreeCtrl.  Indirect rename via the context menu still works, which seems more consistent with the behavior on Windows.

Environment:
M1 Mac Studio macOS 12.6
Xcode 14.0.1
PasswordSafe 1.15 pulled from git master on 2022-10-09
wxWidgets 3.2.1 (installed via homebrew)

```
diff --git a/src/ui/wxWidgets/PasswordSafeFrame.cpp b/src/ui/wxWidgets/PasswordSafeFrame.cpp
index c2a6e7994..e8165551d 100644
--- a/src/ui/wxWidgets/PasswordSafeFrame.cpp
+++ b/src/ui/wxWidgets/PasswordSafeFrame.cpp
@@ -756,9 +756,10 @@ void PasswordSafeFrame::CreateControls()
   wxASSERT(m_grid);
   itemBoxSizer83->Add(m_grid, wxSizerFlags().Expand().Border(0).Proportion(1));

+  // Add the wxTR_EDIT_LABELS style flag to enable direct editing of Tree labels
   m_tree = new TreeCtrl( panel, m_core, ID_TREECTRL, wxDefaultPosition,
                             wxDefaultSize,
-                            wxTR_EDIT_LABELS|wxTR_HAS_BUTTONS |wxTR_HIDE_ROOT|wxTR_SINGLE );
+                            wxTR_HAS_BUTTONS|wxTR_HIDE_ROOT|wxTR_SINGLE );
   wxASSERT(m_tree);
   if(IsTreeSortGroup()) {
     m_tree->SetSortingGroup();
```

PR #889, opened 2022-10-11
Correct issue #888. Allow direct group name editing in tree view

KaiZin2018:
Edit tree list view group name by direct editing.


ronys:
Thanks!

PR #895, opened 2022-10-30
[i18n] Updates in Russian .po and help

pm-kan:


ronys (commented):
The encoding in [Misc/I18N/ru/omegat_help/omegat_help-omegat.tmx](https://github.com/pwsafe/pwsafe/pull/895/files#diff-e5f8dc6cb85d02ffa32a1ba4de82da83c89bbbe79c0a7333839b543e7886e021) seems incorrect - I'm seeing non-Cryllic text in the review panel. Can you check?

pm-kan:
This file (and shown lines) looks fine when opening full file. It's in utf-8 encoding, but github for some reason re-encode it in diff preview (as if it was in some one-byte encoding and than converted to unicode).

PR #896, opened 2022-11-07
[wx] Raise the wx version for current Ubuntus

rafaelx:
With the upgrade to Kubuntu 22.10, Wx version 3.2.0 is available from the official repositories.
To benefit from improvements in more recent Wx versions, this PR upgrades the Wx dependency to Wx version 3.2.0 for Ubuntus from 22.10.


CMakeLists.txt
```diff
@@ -613,7 +613,9 @@ if (${RELEASE_ID_SHORT} STREQUAL "Ubuntu")
   else (NOT ${RELEASE_NUM_SHORT} VERSION_LESS "19.0")
     set (LIBQRENCODE "libqrencode3 (>= 3.4.4-1)")
   endif (NOT ${RELEASE_NUM_SHORT} VERSION_LESS "19.0")
-  if (NOT ${RELEASE_NUM_SHORT} VERSION_LESS "20.0")
+  if (NOT ${RELEASE_NUM_SHORT} VERSION_LESS "22.10")
```
dskrvk, line 616: I’m not too familiar with the cmake syntax, but it seems that previously the expression in parentheses was the same in the if and the else branches. Not sure why an else branch needs a condition or whether it should always match the condition in the first if.

PR #897, opened 2022-11-17
[wx] Raise the wx version for Ubuntu build env

rafaelx:


rafaelx:
Created as a draft as it has yet to be tested on other distributions.

dskrvk:
I wonder if instead of this ever more complicated logic with hard-coded versions there’s a way to query the distro’s package manager for the latest available wxwidgets. Or is there such a thing as dependency management in C-land? (Akin to Pom.xml for Java, package.json for node.js or pipfile for python.)

PR #905, opened 2023-01-12
Clean-up desktop file

Amr-Ibra:
- Remove key "Encoding"
  - It is deprecated and thus useless
  - Can be verified from `$ desktop-file-validate pwsafe.desktop`

- Do not enforce X11
  - pwsafe runs well on Wayland through GTK3 and does not need forcing X11
  - Forcing X11 has the disadvantage of not respecting the display's hi-dpi fractional scaling
  - This has been tested to run well on GNOME Wayland, Debian 11

- Do not hardcode the icon path
  - This enables using a themed icon, named `pwsafe.png` too, coming from a third-party icon theme

- Remove key/value "Terminal=false"
  - "Terminal=false" is the default and thus useless

- Remove useless blank line

All changes have been tested to run well on GNOME Wayland, Debian 11 and up-to-date Debian testing.


ronys:
Thanks!

PR #906, opened 2023-01-13
Initialize the newly constructed buffer instead of setting it to null…

henricj:
Setting a pointer nullptr right after allocating it is almost certainly not the intended behavior.  This PR is the minimal change to avoid this, but one might also ask if the whole buffer should be zeroed and if the allocations should be managed through something like `std::unique_ptr<>`.


ronys:
Nice catch - thanks!

PR #909, opened 2023-01-17
Make the XPM arrays "const".

henricj:
This complicates editing XPM files, but moves many things to read-only data segments.  The first commit only changes the code to treat the arrays as `const`, which works even for XPM arrays that are not `const` and is simply good practice.  The second commit is the one that would mean more work when editing XPM files, because the resulting `.xpm` files are no longer canonical XPM files since they would have declarations of the form,
``` cpp
static const char * const <filename>_xpm[] = {
```
instead of,
``` cpp
static const char * <filename>_xpm[] = {
```
This conversion could also be done at build time, at the cost of a little more build complexity.

As to why a three-decade old image standard only added `const` to the individual strings..?  Being able to change individual strings in the XPM image arrays at runtime would be a rather niche requirement.

Note also that unless the linker is set up to merge identical data symbols, some images will be included in the resulting binary multiple times:
```
$ find src -type f -iname "*.cpp" | xargs grep -h "#include" | grep "\.xpm" | sort | uniq -d -c
      4 #include "graphics/cpane.xpm"
      4 #include "graphics/toolbar/new/copypassword.xpm"
      3 #include "graphics/toolbar/new/copypassword_disabled.xpm"
      6 #include "graphics/Yubikey-button.xpm"
```
One might consider handling those four XPMs from a central repository instead of `#include`ing them directly.

### DIFF
--- a/src/ui/wxWidgets/DragBarCtrl.cpp
+++ b/src/ui/wxWidgets/DragBarCtrl.cpp
@@ -87,10 +87,10 @@ enum
 struct DragbarToolInfo {
   const wxWindowID id;
   const wxString name;
-  const char** bitmap;
-  const char** bitmap_disabled;
-  const char** classic_bitmap;
-  const char** classic_bitmap_disabled;
+  const char* const* const bitmap;
+  const char* const* const bitmap_disabled;
+  const char* const* const classic_bitmap;
+  const char* const* const classic_bitmap_disabled;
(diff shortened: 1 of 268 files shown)


ronys:
Thanks. Indeed the change means that adding new xpms will require additional work/thought.
However, this is rare enough so that it doesn't justify tweaking the build process to automate.

PR #913, opened 2023-01-19
Make sure the source files are encoded as UTF-8.

henricj:
Use UTF-8 to encode the source files.


ronys (changes requested):
Something's wrong here - the corrections aren't mapped to valid characters.
or example, the 'degree' sign in the status bar now looks like this:
broken-status-bar

henricj:
Interesting.  I didn't notice that this set of changes touched anything not in a comment.  It's from the change in [`src/ui/Windows/DboxMain.cpp`](https://github.com/pwsafe/pwsafe/pull/913/files#diff-fd53dc42d2370babb0c8e11d597dd94cd161e94c5a1bfe2de2313fb517048e1f).  It should result in something more interesting if the build computer's locale is set to something not in America or Western Europe (that could be true both with and without this change).  Adding the `/utf-8` flag should fix the problem, regardless of the build computer's locale.

I've been pulling apart the changes I've been making, and while I got the hardening flags out in "[Add create-harden-interface.cmake](https://github.com/henricj/pwsafe/commit/5861bd74d36fb9c85dfb72c885108bf7cbe76e1c)," I haven't gotten around to the rest of the flags yet.  On the plus side, that hardening change turns on MSVC's mitigations and even makes `hardening-check` happy on Debian. Trying to get wxWidgets to compile on Windows resulted in a lot of build file changes.  That still doesn't work because `TreeCtrl` is calling internal methods its parent `wxTreeCtrl` class, and they've hidden them away in wx 3.2 (at least, I think that's what's going wrong).

In the meantime, I can either add a quick `/utf-8` to this PR or drop the `DboxMain.cpp` change..?  The character could also be replaced with the equivalent `\x` or `\u`.

PR #914, opened 2023-01-19
Update how the .chm files are built and build pwsafeZH.chm.

henricj:
The missing html files have been recovered by extracting them from the repo's pwsafeZH.chm.  The CMake build for the help directory was updated so that all the CHM files built reliably (and it was tested to fail reliably when there is an error of some kind).

### DIFF
--- a/help/CMakeLists.txt
+++ b/help/CMakeLists.txt
@@ -75,4 +75,10 @@ foreach (LANG ${LANGS})
   list(APPEND OUTPUT_FILES "${OUTPUT_FILE}")
 endforeach(LANG)
 
-add_custom_target(helpfiles ALL DEPENDS ${OUTPUT_FILES})
+option(PWS_BUILD_HELP "Build the help files by default." OFF)
+
+if (PWS_BUILD_HELP)
+  add_custom_target(helpfiles ALL DEPENDS ${OUTPUT_FILES})
+else()
+  add_custom_target(helpfiles DEPENDS ${OUTPUT_FILES})
+endif ()


henricj:
> Nice work! There's no need to build .chm files for both Release and Debug, though. Could you make the chm files build only under Release?

It turns out that making the build of a target depend on the configuration is complicated.  Theoretically, the "[$<COMMAND_CONFIG>](https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html)" generator expression can do it, but that only seems to work for the ["Ninja Multi-Config" generator](https://cmake.org/cmake/help/latest/generator/Ninja%20Multi-Config.html).  Neither generator expressions on the "helpfiles" target nor on the .chm file dependencies of that target made any impression on the ["Visual Studio 17 2022" generator](https://cmake.org/cmake/help/latest/generator/Visual%20Studio%2017%202022.html) (it is also a multi-config generator, but it always built the help files anyway).  As an alternative, I added an option (`PWS_BUILD_HELP`) that lets one choose if the "helpfiles" target will be part of the default build.  If true, doing a normal build will include the helpfiles target.  If not true, then one has to specify the "helpfiles" target to get them to build:
```
cmake --build . --target helpfiles
```

The default for the `PWS_BUILD_HELP` options is off.  On the other hand, the dependencies are set up properly now so the help files should only get built on the first build (or if one changes any of the help source files, of course).

henricj:
> PWS_BUILD_HELP is nice. hhc is not Microsoft's best work, to put it mildly, but I'm surprised that there's no clear replacement (not to mention a migration path to the other tools I've found).

My guess is that they would view browser-based help as the successor.  One could write a static website and launch the user's default browser to view locally installed help.  The index and catalog stuff should be doable without too much trouble (not that I'm volunteering).

I do have to say that having a security tool like Password Safe still using [bits of Internet Explorer](https://stackoverflow.com/questions/74853761/microsoft-html-help-and-internet-explorer-retirement) to display its help is a bit dodgy.  (Open a .chm file, click "Options", then "Internet Options" and look at that old, familiar dialog.)

PR #915, opened 2023-01-19
Build hardening.

henricj:
Add a new interface library called `harden_interface`.  Adding it to a target will add linker and compiler switches in order to harder the resulting binary.  MSVC, GCC, and Clang are supported, with recent options being checked before they are applied.  It will not add unsuitable options to debug builds.

Some minor build changes were also added in order to build from VS2022 (even without the hardening changes).


henricj:
With the [PIC/PIE commit](https://github.com/pwsafe/pwsafe/pull/915/commits/c19cea9421019eab7f83001ce45774060450504a), the implementation should be done.  It builds and passes my basic smoke test on Windows and Ubuntu jammy/gcc-12, with the latter now reporting,
```
~/src/crypto/pwsafe$ hardening-check build/pwsafe
build/pwsafe:
 Position Independent Executable: yes
 Stack protected: yes
 Fortify Source functions: yes (some protected functions found)
 Read-only relocations: yes
 Immediate binding: yes
 Stack clash protection: yes
 Control flow integrity: yes
```

This is what Microsoft's "[BinSkim](https://github.com/microsoft/binskim) analyze" reports on a `pwsafe.exe` built for `Release` from a VS2022 command prompt set up to use the Spectre-mitigation libraries[^1]:
```
Analyzing...
C:\src\crypto\pwsafe\pwsafe\build\Release\pwsafe.exe: warning BA2027: The PDB for 'pwsafe.exe' does not contain SourceLink information, compromising frictionless source-driven debugging and increasing latency of security response. Enable SourceLink by configuring necessary project properties and adding a package reference for your source control provider. See https://aka.ms/sourcelink for more information.
C:\src\crypto\pwsafe\pwsafe\build\Release\pwsafe.exe: warning BA6001: 'pwsafe.exe' appears to be compiled as release but enables incremental linking, increasing binary size and further compromising runtime performance by preventing enabling maximal code optimization.
C:\src\crypto\pwsafe\pwsafe\build\Release\pwsafe.exe: warning BA6004: 'pwsafe.exe' was compiled with COMDAT folding (/OPT:ICF) disabled, increasing binary size.
C:\src\crypto\pwsafe\pwsafe\build\Release\pwsafe.exe: warning BA6005: 'pwsafe.exe' was compiled with Optimize References (/OPT:REF) disabled, increasing binary size.

Done. 1 files scanned.
Analysis completed successfully.

One or more rules was disabled for an analysis target, as it was determined not to be applicable to it (this is a common condition). Pass --verbose on the command-line for more information.
```

Since these kinds of changes can expose latent bugs in other code and has many compiler/linker option checks during CMake's configure, this needs a lot of testing (e.g., it has never been built on macOS).

[^1]: This is how the command prompt was initialized: `VsDevCmd.bat -arch=x64 -host_arch=x64 -vcvars_spectre_libs=spectre
`

ronys:
What copyright notice would you like for the new files?
Putting them under my copyright like the rest of the project keeps things simple and uniform, but of course if you want to be the copyright owner that's fine too. License should be the same, though.

henricj:
What I publish to the [henricj/pwsafe](https://github.com/henricj/pwsafe) repo or that I publish (e.g., submit as a PR) to the [pwsafe/pwsafe](https://github.com/pwsafe/pwsafe) repo should be considered published under the Artistic License 2.0.  Whatever notice works best for the project is fine by me as long as it is understood that this is a non-exclusive grant for the pwsafe collective work to republish said works under the Artistic License 2.0 (and, obviously, for others to then use the same under the terms of that same license).  That does not mean that such notices that may appear on individual files that identify the copyright holder of the collective work needs to mention me or any of the above.

E.g., I may soon be putting some of that same .cmake gunk into my own dunelegacy project under the GPLv2.

Is that acceptable?

henricj:
The last forced push took things a little further than expected, but linker and compiler flag changes were needed get rid of several more [BinSkim](https://github.com/microsoft/binskim) warnings and it seemed wasteful to not use the existing changes for cleaning up the MSVC compiler flag handling.  This required some other build changes when these flags propagated to the other targets in the codebase (e.g., previously the LTO/LTCG setting was only applied to pwsafe, and not the DLLs it uses).

This can perhaps be broken down into smaller bits, but I think this is about the minimum set of changes needed to make BinSkim happy with all the binaries.

henricj:
> "E.g., I may soon be putting some of that same .cmake gunk into my own dunelegacy project under the GPLv2. Is that acceptable?"
> 
> It's your code - if it's OK with you then that's what matters.

It's all good.  I just wasn't sure if you were fishing for some informal CLA.

ronys:
Tested on:
- Debian 11 x86_64
- Ubuntu 20.04.5 x86_64
- VS2022 x64
- Ubuntu 20.04.5 aarch64
- Fedora37 x64
-  Debian 10 x86

On *all* of the Linuxen:
Performing Test HARDEN_COMPILE_SANITIZE_SAFE_STACK - Failed
Performing Test HARDEN_LINK_SANITIZE_SAFE_STACK - Failed

On what platform(s) are these expected to succeed?

(See attached for all CMAKE config outputs)
[pr915-cmakes.txt](https://github.com/pwsafe/pwsafe/files/10483407/pr915-cmakes.txt)

henricj:
> On *all* of the Linuxen:
> Performing Test HARDEN_COMPILE_SANITIZE_SAFE_STACK - Failed
> Performing Test HARDEN_LINK_SANITIZE_SAFE_STACK - Failed
> 
> On what platform(s) are these expected to succeed?

Those options are only available for Clang (so far, anyway).  Off the top of my head, I don't recall if any are GCC-specific, but at least one of the stack protection options isn't available on Debian 11's default version of GCC.

The Control Flow Integrity option defaults to off since it doesn't work unless linked with libraries that have also been built with CFI.

henricj:
One more thing: To build with VS2022's Spectre mitigation, makes sure the Spectre mitigation libraries are installed.  Run "Visual Studio Installer", click "Modify" for the VS2022 install in question, go to the "Individual Components" tab, enter "Spectre latest" in the "Search components" search box, then make sure the MSVC Spectre-mitigated libs, C++ ATL with Spectre mitigations, and C++ MFC with Spectre mitigations are installed for the architectures in question (likely only x86/x64, but ARM64 should almost work too--minus the translation satellite DLLs).
image
To actually use the Spectre-mitigation libraries, the build must be done from an environment set up something like this:
```
VsDevCmd.bat -arch=x64 -host_arch=x64 -vcvars_spectre_libs=spectre
```
For example, I duplicated the "Developer Command Prompt for VS 2022" in [Windows Terminal](https://learn.microsoft.com/en-us/windows/terminal/) and added `-vcvars_spectre_libs=spectre` to the end of the original command line.
image
The same should be settable through `CMakeConfig.json` or `CMakePresets.json` as well, I think.  Microsoft's [`/Qspectre` documentation](https://learn.microsoft.com/en-us/cpp/build/reference/qspectre) only mentions how to build from the command line.

ronys:
Thanks, but specifically for PasswordSafe, I'm not sure that Spectre mitigation is relevant.
That is, if an attacker is able to mount a Spectre attack on a system running PasswordSafe to discover, say, a given file's encryption key, then the attacker is almost certainly able to mount a much simpler attack, e.g., replacing the pwsafe.exe executable with a modified version.
Or am I missing something?

henricj:
> Thanks, but specifically for PasswordSafe, I'm not sure that Spectre mitigation is relevant. That is, if an attacker is able to mount a Spectre attack on a system running PasswordSafe to discover, say, a given file's encryption key, then the attacker is almost certainly able to mount a much simpler attack, e.g., replacing the pwsafe.exe executable with a modified version. Or am I missing something?

Yes, you are mostly right.  The other mitigations don't require an observer on the same system at exactly the right time to be useful; only the Spectre mitigations address the case where an observer is present when PasswordSafe is reading a malicious file using sensitive material (e.g., a decryption key).  An attacker might gain that visibility through malicious code hosted on a web page (although, I gather that is much harder than it used to be) or from running as another user on a shared Terminal Server system.  As you say, if the user's session is compromised, the attacker can do simpler things like logging keystrokes (the EXE might be harder to muck with than their password file if the user doesn't have admin rights).  The other mitigations also require the attacker devise some means of observing what has happened on the system--albeit without the had real-time requirements needed for a Spectre exploit, since the point of the exercise was presumably to gain access to the compromised passwords and that requires some means of exfiltration (true, one might consider other threats like an attacker modifying or denying access to stored passwords, using PasswordSafe as the means of gaining access to the rest of the system, etc., but I would guess that where PasswordSafe is concerned, most attackers' primary interest would be the actual passwords).

OTOH, it's just a matter of having the Spectre libs installed and building with them.  I can see about what flags would be needed in CMakeConfig.json or CMakePresets.json; I assume there is something to set there (if VS doesn't handle it automatically when it sees the Spectre-mitigation flag), but I haven't investigated.

TL;DR: If it doesn't complicate the release procedure, it might still help in some circumstances (which is true for most mitigations).

PR #916, opened 2023-01-21
EXE/DLL "Details" Copyright Symbol Corruption

henricj:
The RC files have `#pragma code_page(1252)` which forces `rc.exe` to interpret the files as Windows-1252 code page when parsing them.  Modern text editors tend to save files in UTF-8, so this makes keeping the version string copyright symbol from degenerating into mojibake difficult.  These commits switch the RC files to declare the UTF-8 code page (65001) and fixes the current copyright symbol corruption.  Round-tripping through VS2022's Resource Editor was done and the code page directive did survive.

Microsoft's Raymond Chen has a blog post about this problem: [_The Resource Compiler defaults to CP_ACP, even in the face of subtle hints that the file is UTF-8_](https://devblogs.microsoft.com/oldnewthing/20190607-00/?p=102569)

Note that the encoding for the strings inside the resulting binaries was already Unicode as can be seen from the "1200" in the "[VarFileInfo](https://learn.microsoft.com/en-us/windows/win32/menurc/varfileinfo-block)":
```
    BLOCK "VarFileInfo"
    BEGIN
        VALUE "Translation", 0x409, 1200
    END
```

This PR is marked "draft" since it is on top of PR #915.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -94,11 +94,11 @@ if (NOT DEFINED pwsafe_VERSION_MINOR)
   message (FATAL_ERROR "VER_MINOR undefined in ${VERSION_FILE} - unable to proceed")
 endif (NOT DEFINED pwsafe_VERSION_MINOR)
 
-if (pwsafe_REVISION)
-   set (pwsafe_VERSION "${pwsafe_VERSION_MAJOR}.${pwsafe_VERSION_MINOR}.${pwsafe_REVISION}")
-else ()
-   set (pwsafe_VERSION "${pwsafe_VERSION_MAJOR}.${pwsafe_VERSION_MINOR}.0")
-endif (pwsafe_REVISION)
+if (NOT pwsafe_REVISION)
+  set (pwsafe_REVISION 0)
+endif ()
+
+set (pwsafe_VERSION "${pwsafe_VERSION_MAJOR}.${pwsafe_VERSION_MINOR}.${pwsafe_REVISION}")
 
 # Configurable options:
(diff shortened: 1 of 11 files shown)


ronys (changes requested):
The value of VER_MINOR is version.mfc appears correctly in the exe properties, but in the opening window and Help->About it's not handled correctly
image
image

henricj:
> The value of VER_MINOR is version.mfc appears correctly in the exe properties, but in the opening window and Help->About it's not handled correctly.

I added the version major/minor/patch directly to `version.h` so that the PWSversion ctor can just use them instead of playing string parsing games.  Even the "modified" flag could be handled similarly.

One other thought: The `version.h` file is only updated at configuration time.  It is usually safer to update it at build time so that if something other than a build-system file is updated, the new build will get an updated `version.h` file.  It can cause release process issues since if a file is modified after the configuration but before the build, then the release build will produce a binary with the modified flag.  This is probably an, "after a new release, not just before," change.

ronys (approved):
Fixed "0.0" problem + simplified code.

ronys:
Agh - looks like I missed commit [eb8dbd4](https://github.com/pwsafe/pwsafe/pull/916/commits/eb8dbd47d298823b338f7de94bf327baa624d5a2) and duplicated it myself!

henricj:
> Agh - looks like I missed commit [eb8dbd4](https://github.com/pwsafe/pwsafe/pull/916/commits/eb8dbd47d298823b338f7de94bf327baa624d5a2) and duplicated it myself!

The merge collided with the same branch for me locally, on what is now #921.  Most of the difference plus some other changes I was working are there.

PR #921, opened 2023-01-28
Improve translation file handling.

henricj:
Well, it's back.

I was pushing some changes that moves the `configure_file()` for `version.h` from configuration to the build.  After this, doing something that changes what is in `version.h` should cause anything that depends on that file to get rebuilt with the new values.  For example, changing the version number or adding a git tag should cause `version.h` to change during the next build, thereby causing anything that depends on it to also get rebuilt, like the pwsafe executable.

I'm marking this "draft" since I'm not sure what parts of the merge you are actually interested in nor what impact it might have on the release build process.

Oh, you might want to snipe [99ed710](https://github.com/pwsafe/pwsafe/commit/99ed710a4922a8667c3474bbcdfd1ffca3707a56) right now.


src/ui/Windows/ThisMfcApp.cpp
```diff
@@ -1515,26 +1515,25 @@ void ThisMfcApp::GetLanguageFiles()
...
-        delete[] szLanguage_NativeUpper;
       }
       // Get language name in English
-      int lnum = ::GetLocaleInfo(lcid, LOCALE_SENGLANGUAGE, NULL, 0);
+      int lnum = ::GetLocaleInfo(lcid, LOCALE_SENGLANGUAGE, nullptr, 0);
       if (lnum > 0) {
-        szLanguage_English = new wchar_t[lnum + 1];
-        ::GetLocaleInfo(lcid, LOCALE_SENGLANGUAGE, szLanguage_English, lnum);
-        if (wcscmp(szLanguage_Native, szLanguage_English) != 0) {
+        szLanguage_English = std::make_unique<wchar_t[]>(lnum + 1);
```
dskrvk, line 1522: Could we reduce this variable’s scope to the enclosing if block on line 1521?

src/ui/Windows/ThisMfcApp.cpp
```diff
@@ -1474,39 +1471,42 @@ void ThisMfcApp::GetLanguageFiles()
...
         }
       }
 
       // Now try to make first character of language name in that language upper case
       // Should use LCMapStringEx but that is for Vista + and we support XP & later
-      wchar_t *szLanguage_NativeUpper(NULL);
       int iu = LCMapString(lcid, LCMAP_LINGUISTIC_CASING | LCMAP_UPPERCASE,
-                           szLanguage_Native, inum,
-                           szLanguage_NativeUpper, 0);
+                           szLanguage_Native.get(), inum,
```
dskrvk, line 1503: This seems like a change in behavior - is ‘szLanguage_Native’ guaranteed to be null here?

src/ui/Windows/ThisMfcApp.cpp
```diff
@@ -1474,39 +1471,42 @@ void ThisMfcApp::GetLanguageFiles()
     // Create LCID
     LCID lcid = MAKELCID(FoundResLangID, SORT_DEFAULT);
 
-    inum = ::GetLocaleInfo(lcid, LOCALE_SNATIVELANGNAME, NULL, 0);
+    inum = ::GetLocaleInfo(lcid, LOCALE_SNATIVELANGNAME, nullptr, 0);
     if (inum > 0) {
+      std::unique_ptr<wchar_t[]> szLanguage_English;
+      std::unique_ptr<wchar_t[]> szCountry_Native;
+      std::unique_ptr<wchar_t[]> szCountry_English;
```
dskrvk, line 1478: Could we similarly reduce the scope of the last two variables to the if block below?

src/ui/Windows/ThisMfcApp.cpp
henricj: [LCMapString()](https://learn.microsoft.com/en-us/windows/win32/api/winnls/nf-winnls-lcmapstringw) ignores the `lpDestStr` argument (the next-to-last one) if the `cchDest` (the last argument) is zero:
> The application can set cchDest to 0. In this case, the function does not use the lpDestStr parameter and returns the required buffer size for the mapped string or sort key.

src/ui/Windows/ThisMfcApp.cpp
henricj: Yup.  Done in the [c985970](https://github.com/pwsafe/pwsafe/commit/c9859707191a9d1aaaec5518523c52da4fc33424) force push.

src/ui/Windows/ThisMfcApp.cpp
henricj: > Could we similarly reduce the scope of the last two variables to the if block below?

Unfortunately, they are initialized in the scope starting on line 1484, but then used in the scope starting on line 1520.

henricj:
This is stacked on top of PR #926.

PR #927, opened 2023-01-31
Add the "harden_interface" target to YkLib22 and pwsafe_base.

henricj:
The `harden_inteface` target was added to the `pws_at` target three times.  Let the `YkLib22` and `pwsafe_base` targets get their share, as originally intended.

This is not expected to have a meaningful impact since `YkLib22` is a static library and `pwsafe_base` is a resource-only satellite DLL.


src/ui/Windows/language/CMakeLists.txt
```diff
@@ -9,6 +9,6 @@ set(CMAKE_CXX_FLAGS
   "${CMAKE_CXX_FLAGS} /D languageDLL_EXPORTS")
 
 add_library(pwsafe_base SHARED ${LANG_SRCS} ${LANG_RES})
-target_link_libraries(pws_at PRIVATE harden_interface)
+target_link_libraries(pwsafe_base PRIVATE harden_interface)
```
ronys, line 12: Not strictly needed, as pwsafe_base is only used to extract strings for i18n support.
For consistency I guess it's worth it, though.

henricj: From a practical security perspective, you're probably right.  From the point of view of `LoadLibrary()` and security tools that go looking for PE header flags that are not set, well... I'm not sure.  That DLL is used as a template for the translation DLLs, so the PE flags and what-not go along for the ride.

Also, until [633f720b](https://github.com/pwsafe/pwsafe/pull/921/commits/633f720b62b432eb150646bb29ff379823fd1635) (part of PR #921), there actually is some pointless initialization code in that DLL (that code also gets copied to the translation DLLs).

PR #928, opened 2023-01-31
Force core.rc2 and cli.rc to be interpreted as UTF-8.

henricj:
This resolves the mojibake in `coretest.exe` and `pwsafe-cli.exe`.  The `core.rc` file was already UTF-8, so let `RC.exe` know to interpret it as such.  The other file, `cli.rc` was UTF-16 and just `#include`s `core.rc`.  Since the `#pragma code_page()` is only recognized in top-level RC files, we add the pragma to `cli.rc` and convert the file from UTF-16 to UTF-8.


henricj:
This is included in PR #921.

PR #932, opened 2023-02-04
Switch to CMakePresets.json and update GitHub CI builds.

henricj:
Switch from Microsoft's [CMakeSettings.json](https://learn.microsoft.com/en-us/cpp/build/cmakesettings-reference) to [CMakePresets.json](https://cmake.org/cmake/help/latest/manual/cmake-presets.7.html).

Use the presets to do GitHub CI builds for Linux and Windows.

This is stacked on top of PR #921 since the parallel builds don't work properly without it.


.github/workflows/cmake-build.yml
```diff
@@ -0,0 +1,48 @@
...
+    runs-on: ${{ matrix.os }}
+
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v3
+      with:
+        fetch-depth: 0
+
+    - name: Get cmake and ninja
+      uses: lukka/get-cmake@latest
```
dskrvk, line 27: Why use latest rather than a specific version? Not great for repeatable builds.

henricj: Are you're referring to the CMake/Ninja versions or the action's version?  Either way, you are right and it's not just about repeatability, but potentially about security too.  For a project like this, all the actions should probably use the [commit id](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions) for referencing the version.

I'll take a look after getting all the presets to work.

ronys:
Thanks! Updating this was on my todo list, but the learning curve for GitHub's automation was a bit too much for me.

Only question for me is regarding Ninja, but I'm willing to try something new for the new year :-)

PR #952, opened 2023-04-24
[macOs] Update to wxWidgets 3.2.2.1

nobugshere:
I'd like to suggest we update to wxWidgets 3.2.2.1 (the current version).  3.2.1 has a Mac specific bug that does strange things to the font rendering.  For example, in the tree view, as I click around, the list font seems to toggle between normal and a highlight or bold look.  It might depend on font size and whether or not you have a HiDPI monitor (I don't.). I've been using 3.2.2.1 since its release and I haven't noticed any new problems.

This commit only changes the macOS build and sets FORCE_REBUILD_WX=6.


dskrvk (approved):

ronys:
Thanks!

PR #988, opened 2023-06-01
macOS: Check for new version crashes the app

nobugshere:
Clicking the link to check for a newer version from the About dialog causes the app to abort when closing the database.  (See attached screenshot.)
The first commit fixes that problem by delaying closing the About dialog until after the process to close the database.  I don't currently have a Linux environment set-up to test this patch.  For that matter, I'm not really sure if the problem even happens on Linux, there are some differences in that part of wxWidgets.  I hope someone can test the Linux build before merging this PR.  If there is a problem, or some other reason for discomfort, these changes should be easy to ifdef.

Secondary problem:  Once it stopped crashing, I discovered it was checking the PC version number instead of the combined Mac/Linux WX version.  The second commit fixes that.

evtloop abort 2023-05-25

### DIFF
--- a/src/core/CheckVersion.cpp
+++ b/src/core/CheckVersion.cpp
@@ -62,8 +62,8 @@ CheckVersion::CheckLatestVersion(const stringT &xml, stringT &latest) const
 
         const stringT variant(pVariant);
         // Determine which variant is relevant for us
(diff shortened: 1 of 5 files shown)


ronys:
Works fine on Linux - thanks!

PR #989, opened 2023-06-01
macOS: Let the External Keyboard Button.cpp do something

nobugshere:
On macOS, the external keyboard button on the open dialog doesn't do anything.  I'm not sure this is good but I can't figure out how to open the virtual keyboard, so this patch will open the System Settings app to the right page and prompt the user to enable it.  Any thoughts?

### DIFF
--- a/src/ui/wxWidgets/ExternalKeyboardButton.cpp
+++ b/src/ui/wxWidgets/ExternalKeyboardButton.cpp
@@ -95,7 +95,7 @@ void ExternalKeyboardButton::HandleCommandEvent(wxCommandEvent& evt)
   wxString command = wxString("open x-apple.systempreferences:com.apple.preference.universalaccess?Keyboard");
 
   if ( wxExecute(command, wxEXEC_ASYNC, nullptr) > 0) {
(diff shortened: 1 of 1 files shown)


src/ui/wxWidgets/ExternalKeyboardButton.cpp
```diff
@@ -90,4 +90,16 @@ void ExternalKeyboardButton::HandleCommandEvent(wxCommandEvent& evt)
   }
 #endif
 
+#ifdef __WXOSX__
+  // If we can't open the virtual keyboard, at least open the settings app so the user can do it for us!
+  wxString command = wxString("open x-apple.systempreferences:com.apple.preference.universalaccess?Keyboard");
+
+  if ( wxExecute(command, wxEXEC_ASYNC, nullptr) > 0) {
+    wxMessageBox(_("Please, enable the Accissibility Keyboard in System Settings"), "", wxOK | wxICON_INFORMATION);
```
ronys, line 98: 1. Typo - should be "Accessibility"
2. I'd add " and try again"

PR #990, opened 2023-06-04
Make it possible to build on OpenBSD

sauravg:
Didn't test extensively, but whatever I use for my purposes seems to work...


ronys:
Need to s/_S/_T/ in Windows as well. I can get to it in a couple of days.

src/os/typedefs.h
```diff
@@ -30,7 +30,6 @@ typedef std::string  cstringT;
 
 typedef std::wstring stringT;
 typedef wchar_t charT;
-#define _S(x) L ## x
```
dskrvk, line 33: Looks like the comment on 21-24 is now obsolete

src/ui/wxWidgets/Makefile
```diff
@@ -57,7 +59,7 @@ OUTPUTPATH=GCCDebug
 OBJECTPATH=GCCDebug
 BUILDPATHS=$(OBJECTPATH)
 PROGRAM=pwsafe
-LIBS=`$(WX_CONFIG) --debug=yes --unicode=no --libs` -lcore -los -lcore -luuid -lXtst -lX11  -lcurl -lssl -lmagic -l$(YBPERS_LIBS) ${LIBCXX_LD_EXTRA_FLAGS}
+LIBS=`$(WX_CONFIG) --debug=yes --unicode=no --libs std,aui` -lcore -los -lcore -luuid -lXtst -lX11  -lcurl -lssl -lmagic -l$(YBPERS_LIBS) ${LIBCXX_LD_EXTRA_FLAGS}
```
dskrvk, line 62: Does this change affect every platform? What does it do?

src/os/typedefs.h
ronys: Yes, comment was removed.

Makefile.openbsd
```diff
@@ -0,0 +1,148 @@
...
+	$(MAKE) CONFIG=unicodedebug -C src/test clean
+
+release-clean:
+	$(MAKE) CONFIG=unicoderelease -C src/os/unix clean
+	$(MAKE) CONFIG=unicoderelease -C src/core clean
+	$(MAKE) CONFIG=unicoderelease -C src/ui/wxWidgets clean
+	$(MAKE) CONFIG=unicoderelease -C src/test clean
+
+# dist prepares stuff for upload - deprecated for Linux, until
+# we find an elegant way to tell deb and rpm distros apart.
```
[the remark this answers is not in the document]
ronys: Nope, rpm commands can be installed on debian based systems, and vice versa.
In any case, cmake/cpack are the recommended way to compile and build packages for pwsafe.

src/ui/wxWidgets/Makefile
sauravg: Yes, it affects every platform (but shouldn't hurt).

Without this change, it's equivalent to `--libs std`, which lists only standard wx libraries. With `aui` added, wx-config emits wxAui* libraries in addition. pwsafe needs the wxAui libraries.

PR #994, opened 2023-06-13
[wx] Fix repeated search with enter key

### LINKED ISSUE #963: Linux PS: Enter not working to repeat a string search
I have used PS on Windows extensively in the past, and I have noticed a feature that is not apparent in the Linux version. When I Ctrl+F to search and enter a string, I press enter to search which works. In Windows I could press enter again and it continues to search for the same string. In the Linux version, however, pressing enter again does nothing. Pressing the magnifying glass icon solves this, but it does slow things down.

For consideration.

rafaelx:
Fix #963.

### DIFF
--- a/src/ui/wxWidgets/PasswordSafeSearch.cpp
+++ b/src/ui/wxWidgets/PasswordSafeSearch.cpp
@@ -226,6 +226,7 @@ void PasswordSafeSearch::OnDoSearchT(Iter begin, Iter end, Accessor afn)
   }
 
   UpdateView();
+  txtCtrl->SelectNone();
 
   // Replace the "Find" menu item under Edit menu by "Find Next" and "Find Previous"
   wxMenu* editMenu = nullptr;
@@ -250,6 +251,8 @@ void PasswordSafeSearch::UpdateView()
     m_parentFrame->SelectItem(*m_searchPointer);
   }
   statusArea->SetLabel(m_searchPointer.GetLabel());
+
+  SetFocusIntoEditField();
 }
 
 void PasswordSafeSearch::FindNext()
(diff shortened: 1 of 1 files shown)


src/ui/wxWidgets/PasswordSafeSearch.cpp
```diff
@@ -630,12 +635,27 @@ void PasswordSafeSearch::OnChar(wxKeyEvent& event)
...
+        // If nothing is marked in search text field,
+        // the item's username shall be copied.
+        wxCommandEvent copy_username_event(wxEVT_MENU, ID_COPYUSERNAME);
+        m_parentFrame->GetEventHandler()->AddPendingEvent(copy_username_event);
+      }
+      else if ((event.GetKeyCode() == wxT('x')) || (event.GetKeyCode() == wxT('X'))) {
+        // If nothing is marked in search text field,
+        // the search text field shall be cleared.
+        wxCommandEvent search_clear_event(wxEVT_MENU, ID_FIND_CLEAR);
+        GetEventHandler()->AddPendingEvent(search_clear_event);
```
dskrvk, line 654: No need for `m_parentFrame` here?

src/ui/wxWidgets/PasswordSafeSearch.cpp
```diff
@@ -630,12 +635,27 @@ void PasswordSafeSearch::OnChar(wxKeyEvent& event)
...
+      }
+      else {
```
dskrvk, line 656: Why needed?

src/ui/wxWidgets/PasswordSafeSearch.cpp
```diff
@@ -630,12 +635,27 @@ void PasswordSafeSearch::OnChar(wxKeyEvent& event)
         // then normal copy event shall be handled.
         event.Skip();
       }
-      else {
+      else if ((event.GetKeyCode() == wxT('c')) || (event.GetKeyCode() == wxT('C'))) {
```
dskrvk, line 638: Nit: assigning `event.GetKeyCode()` to a variable would make these if conditions easier to read. Also, there's no need for () around equality checks since `==` has higher precedence than `||`.

src/ui/wxWidgets/PasswordSafeSearch.cpp
```diff
@@ -620,7 +623,9 @@ void PasswordSafeSearch::OnChar(wxKeyEvent& event)
     HideSearchToolbar();
   }
   else if ((event.GetModifiers() == wxMOD_CONTROL) && 
-    ((event.GetKeyCode() == wxT('c')) || (event.GetKeyCode() == wxT('C')))) {
+    ((event.GetKeyCode() == wxT('c')) || (event.GetKeyCode() == wxT('C')) || 
+     (event.GetKeyCode() == wxT('u')) || (event.GetKeyCode() == wxT('U')) ||
+     (event.GetKeyCode() == wxT('x')) || (event.GetKeyCode() == wxT('X')))) {
```
dskrvk, line 628: Nit: this list of keys might grow in the future. Can we do something like `wxT('CcUuXx').contains(event.GetKeyCode())` (sorry, don't know the right string operator to use).

PR #996, opened 2023-06-23
Fix clear clipboard on minimize option

nobugshere:
The clear clipboard on minimize option does not work unless the lock DB on minimize option is set.  This change will clear  the clipboard in either case.  This would affect both Mac and Linux, but I've only tested on Mac.


ronys:
Nice catch - thanks!

PR #998, opened 2023-07-08
Cancel changes to a policy with custom symbol list

nobugshere:
When using a custom PW policy symbol list, the policy change comparison would always say the dialog content has changed when an entry is opened and canceled without changes.

This is because the symbols string in the old PWPolicy object was not being populated, so it never matched the dialog box.  Since the symbol list is being compared earlier in the function, I changed this part to just compare the policy options as encoded strings.


ronys:
Nice fix - thanks!

Issue #954, opened 2023-04-26
Input Focus in Linux does not return to the password List after autolock


SMSmith:
The referenced fix works for when pwsafe is first launched, but if the safe is locked (manually or after timeout), an unlock fails to focus in the password list.  Is there a way to make it focus on the password list after unlock after a timeout-lock?

_Originally posted by @SMSmith in https://github.com/pwsafe/pwsafe/issues/791#issuecomment-1508539657_

rafaelx:
I can confirm the behavior caused by implementation changes from the past. An error correction is required.

SMSmith:
This still does not work.  Is there anywhere in the code I can look at to help contribute a solution?

nobugshere:
I can confirm this happens on macOS as well, though it might be clouded by some Mac specific behavior.  PR #1008 did not, specifically, address keyboard focus.  That fix only saved/restored the state of the tree view.

@SMSmith, If you want to delve into it, you might start at src/ui/wxWidgets/PasswordSafeFrame.cpp  PasswordSafeFrame::UnlockSafe, called from the bottom of SystemTray.cpp.

PR #1001, opened 2023-07-22
Small fixes - mostly macOS 

nobugshere:
These were developed and tested on macOS Ventura 13.4.1(c) with Xcode 14.3.1 and wxWidgets 3.2.2.1.  
These have not been tested on Linux, review and comments from the Linux devs are welcome.

- Add a call to setrlimit() to prevent core files in a release build (macOS only, already in Unix versions) (In normal execution, the soft limit defaults to 0, but I was able to get around it.)
- Add a call to pws_os::DisableDumpAttach() to pwsafe-cli so it too will have core files and debugging restricted.

- The PW history list is not stored in an order that is consistent with usage, this led to two problems:
    - First time after two or more PW changes, opening and canceling an entry would cause a spurious change notification  because the order in the history was different than expected.
    - When the limit is reached, the wrong entry is deleted if the entry was ever re-saved because of the above problem.

- Mac: UI issues with the system tray icon enabled.  These were mentioned in the discussion for PR #933 
    - (Mac only) When closing the window, locking the DB, or lock on timeout, the window cannot be restored by clicking on the dock icon
    - When a locked DB is unlocked, the window is restored, but is left blank.   I ifdef’ed the fix for this one to be Mac only, but it might also be applicable to Linux.   See SystemTray.cpp:315 and PasswordSafeFrame.cpp:2439
    - Fixed a few override warnings in PWSafeApp.h


src/ui/wxWidgets/PasswordSafeFrame.cpp
```diff
@@ -2432,6 +2432,9 @@ void PasswordSafeFrame::UnlockSafe(bool restoreUI, bool iconizeOnCancel)
       {
         if (ReloadDatabase(scp->GetPassword())) {
           m_sysTray->SetTrayStatus(SystemTray::TrayStatus::UNLOCKED);
+#ifdef __WXMAC__
+          m_InitialTreeDisplayStatusAtOpen = true;
```
rafaelx, line 2436: This change would also be interesting for Linux, so that the tree view is correctly restored after unlocking. Unfortunately, this functionality has been lost with changes to dialog handling.

In the settings (Manage > Options > Display: Initial Tree View) there are three options that affect the tree view once it is displayed.

- Fully collapsed
- Fully expanded
- Same as when last saved

The last option doesn't seem to work correctly anymore. The tree view may be saved incorrectly by calling PasswordSafeFrame::ShowTree and GuiInfo::SaveTreeViewInfo multiple times.

nobugshere: I'm not quite sure what you're suggesting here.  Do you want me to remove the ifdef for that line?  On the Mac, I also need the part in SystemTray.cpp to have the view restored at all, otherwise the window comes-up blank.  As for the "last saved" option, it does have an affect for me, but it restores the view to the way it was when the **_file_** was last saved, same as it does for an initial open, since it basically is closing and re-opening the file.  I'm not sure what the original intent was.

rafaelx: From my understanding, it should behave as follows regarding the "Initial Tree View" option.
After opening the database, all groups should initially be collapsed, expanded or displayed as last used by the user in the tree view. The "Initial Tree View" option should not have any further effects on the tree view at runtime. Only the changes made by the user to the tree view should be relevant. This means that when switching between views or unlocking the database, the changes should always be saved and restored accordingly.

I was initially under the impression that introducing "m_InitialTreeDisplayStatusAtOpen = true;" fixed the problem of restoring the tree view after an unlock. However, this was due to my set options. It seems like calling "m_guiInfo->RestoreTreeViewInfo(m_tree);" lost with some changes in the past.

The erroneous behavior of the tree view after unlocking the database also results in the loss of focus on the element last selected by the user. (see issue #954)

nobugshere: I think what's really happening is _PasswordSafeFrame::SaveAndClearDatabaseOnLock()_ calls _PasswordSafeFrame::ClearAppData()_ which explicitly clears the saved grid and tree info, so there's no state to restore upon unlock.  I'm not sure why that was done. Maybe it was to clear sensitive info from memory?  But, I'm not too sure it accomplishes that.

rafaelx: I see the expected behavior of the tree view when calling HideUI(false) in PasswordSafeFrame::IconizeOrHideAndLock() instead of HideUI(true).
However, if there is a breakpoint in GuiInfo::SaveTreeViewInfo, it can be observed that multiple attempts are made to lock the database, which cannot be correct either.
The Show-Hide-Iconize-Lock implementation seems quite opaque to me at the moment.

nobugshere: Yes, I understand.  I'm working on a better fix, but I got sidetracked.  I'll update this PR soon.

nobugshere: You are right, there are multiple redundant calls to save and restore the tree/grid views.  The last call to save view info happens after the views have been cleared.  GuiInfo::SaveGridViewInfo() has an explicit check at the top (grid->GetNumItems() == 0) that avoids the problem.  As a workaround, I inserted an equivalent check into GuiInfo::SaveTreeViewInfo() and that seems to restore the correct behavior from the user perspective.  I have not tested this on Linux, but I think this might fix #954 as well.  At least, it works as expected on Mac.

I also bumped the version number to 1.17.2 in case new issues appear.

I'm finding the Show-Hide-Iconize-Lock implementation rather confusing as well.  Since any re-work to streamline it would be a complex project, I would like to suggest logging that as a separate issue and letting this PR focus on the user visible behavior.

ronys: The change to ItemData broke the unit test:

[  FAILED  ] ItemDataTest.PasswordHistory

rafaelx: Yes! Commit da6b74881898c91f291eeb75d4ede649e03ab0bd brings back the old behavior to the tree view. :+1:

src/ui/wxWidgets/SystemTray.cpp
```diff
@@ -302,6 +302,7 @@ void SystemTray::ShowSetDatabaseIdDialog()
 void SystemTray::ProcessSysTrayMenuItem(int itemId)
 {
   switch(itemId) {
+    case ID_SYSTRAY_UNLOCK:
```
rafaelx, line 305: The expected behavior in case of Unlock is to only unlock the database without showing the UI to the user.

src/ui/wxWidgets/SystemTray.cpp
```diff
@@ -302,6 +302,7 @@ void SystemTray::ShowSetDatabaseIdDialog()
...
     case ID_SYSTRAY_RESTORE:
```
rafaelx, line 306: In case of Restore, the database should be unlocked and shown to the user again.

src/ui/wxWidgets/SystemTray.cpp
nobugshere: I see now, on the windows version, left-clicking the systray icon to unlock restores the window, but right-clicking and selecting the menu item does not.  What I was seeing on the Mac is that an unlock from the menu would raise the window but not display any content, tree or grid, until the window was minimized and restored.  The systray icon on the Mac displays its menu with either left or right click.  How does the 1.17.0 release behave on Linux?

src/ui/wxWidgets/PasswordSafeFrame.cpp
nobugshere: How do I retrieve the log from the test run?
Where is the code for the failed test?

I'm sure it's because I changed the the way the history entry was sorted.  I just discovered that when I transfer my test DB from Mac to Windows, the newly saved history was displayed with oldest at the top.

At this point, I think I should retract this PR, resubmit the changes that seem to be okay, and open issues for the other two.

src/ui/wxWidgets/SystemTray.cpp
rafaelx: On Linux, the system tray menu appears only on a right click and I have never observed the case of an empty main frame.

nobugshere: Does a left-click on the system tray icon do anything?

rafaelx: A single left-click triggers an unlock and shows the main UI.

nobugshere: Thanks for checking that for me.

PR #1006, opened 2023-08-03
wx: Prevent core files and debugging in release builds

nobugshere:
- Add a call to setrlimit() to prevent core files in a release build (macOS only, already in Unix versions) 
(In normal execution, the soft limit defaults to 0, but I was able to get around it.)
- Add a call to pws_os::DisableDumpAttach() to pwsafe-cli so it too will have core files and debugging restricted. (Potentially affects all builds.)

This was tested on macOS Ventura 13.5 with Xcode 14.3.1 and wxWidgets 3.2.2.1.   pwsafe-cli should be tested on Linux as well.


ronys:
Thanks for closing these corners.

PR #1008, opened 2023-08-03
Mac and Linux: Save/Restore tree view on lock/unlock

nobugshere:
Ensure the tree view information is preserved on lock so it can be restored when the file is unlocked.
See discussion in #1001 and #954.  The crux of the problem is redundant calls to GuiInfo::SaveTreeViewInfo, the last one happening after the info has been cleared.  I noted that GuiInfo::SaveGridViewInfo already had a check for an empty grid, so I added a similar check to GuiInfo::SaveTreeViewInfo.  Note that there are also redundant calls to restore the grid/tree views. 

This was tested on macOS Ventura 13.5 with Xcode 14.3.1 and wxWidgets 3.2.2.1.   It needs re-testing on Linux.

### DIFF
--- a/src/ui/wxWidgets/GuiInfo.cpp
+++ b/src/ui/wxWidgets/GuiInfo.cpp
@@ -55,6 +55,10 @@ void CollectExpandedNodes(TreeCtrl* tree, wxTreeItemId root, wxArrayString& expa
 
 void GuiInfo::SaveTreeViewInfo(TreeCtrl* tree)
 {
+  //has the tree been cleared?
+  if (tree->HasItems() == 0)
+    return;
+
   //save the first visible item
   wxTreeItemId treeItem = tree->GetFirstVisibleItem();
   if (treeItem.IsOk() && treeItem != tree->GetRootItem()) {
--- a/src/ui/wxWidgets/GuiInfo.cpp
+++ b/src/ui/wxWidgets/GuiInfo.cpp
@@ -53,6 +53,12 @@ void CollectExpandedNodes(TreeCtrl* tree, wxTreeItemId root, wxArrayString& expa
   }
 }
 
+//
+// NOTE: The Save/Restore Tree/Grid view routines are called multiple times for
+// a single lock/restore operation.  When locking, the final call to Save*ViewInfo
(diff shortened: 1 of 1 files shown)


dskrvk:
The check makes sense to add, but since you've already found the redundant calls, maybe at least mark them with "TODO: redundant call - remove"?

nobugshere:
What I meant was that I can see the routines being called more than once, I don't actually know which calls, if any, can be removed.  I suspect they are all there to handle different edge cases.  I added a comment to that affect.  I find all the different paths to lock/minimize/hide/unlock/restore difficult to follow, re-factoring would be a major project.

rafaelx:
If the database is locked from the main toolbar, all items in the tree view are always collapsed when the database is unlocked.

nobugshere:
> If the database is locked from the main toolbar, all items in the tree view are always collapsed when the database is unlocked.

This is the change from PR #1001 that should fix that; minus the part you didn't like :-)

rafaelx:
There seems to be a difference in behavior whether an item is selected in the tree view or not.

First initial situation: A group of the tree view is expanded and an element is selected.
When locking, the console shows that it sometimes takes up to three attempts to minimize the window.

> pwsafe:[7688]: OnIconize: will LockDb()
> pwsafe:[7688]: OnIconize: beingIconized=0
> pwsafe:[7688]: TryIconize attempt
> pwsafe:[7688]: TryIconize attempt
> pwsafe:[7688]: TryIconize attempt

=> After unlocking, the tree view is correctly restored.


Second initial situation: A group of the tree view is expanded and no element is selected.
When locking the database, the console shows that it often takes up to five attempts to minimize the window.

> pwsafe:[7688]: OnIconize: will LockDb()
> pwsafe:[7688]: OnIconize: beingIconized=0
> pwsafe:[7688]: TryIconize attempt
> pwsafe:[7688]: TryIconize attempt
> pwsafe:[7688]: TryIconize attempt
> pwsafe:[7688]: TryIconize attempt
> pwsafe:[7688]: TryIconize attempt

=> In this case, all elements of the tree view are shown collapsed after unlocking.

This may have been the case with previous versions as well. So this correction is at least an improvement.

PR #1011, opened 2023-08-16
Update ReleaseNotesWX.md

nobugshere:


docs/ReleaseNotesWX.md
```diff
@@ -7,10 +7,26 @@ https://pwsafe.org/. Details about changes to older releases may be found in the
 
 In the following, SFxxxx refers to Bug Reports, Feature Requests and Service Requests in PasswordSafe SourceForge Project tickets, and GHxxxx refers to issues in the PasswordSafe GitHub project.
 
+
+PasswordSafe 1.18 Release XXX 2023
+====================================
+
+Bugs fixed in 1.18
```
dskrvk, line 14: Nit: probably no need to mention the version here, since it's already mentioned in the release header. Same for "Changes".

nobugshere: A fair point, but I'm just matching the style in ReleaseNotes.md.

seamusdemora:
Any idea on the timeline for release of ver 1.18? 

BTW, kudos on the progress that's been made on the macOS version! :)  

I do wonder though if "bundling" the Mac & Linux versions together is going to be viable over the longer term... has there been any discussion on this?

ronys:
Hm. Sept-Oct, I guess.

Most of the code is common, so I don't see a good reason to break "wx" into "linux" & "macos".

seamusdemora:
Is the 1.18 code undergoing some sort of a pre-release test phase for the next several weeks?  If so, can I volunteer to help with this pre-release testing?

ronys:
Great idea.
On what platform would you like to test a pre-release?

seamusdemora:
On my Mac - a 2023 M2 running Ventura 13.4.  I currently have the V 1.17 release installed, and would like to avoid removing/disabling/upgrading it for the time being.

nobugshere:
IIRC, in another thread you mentioned you also have a Mac running Mojave. I'm curious to know if you've tried 1.17 on it?

seamusdemora:
@nobugshere -  Good memory :)   Yes, I do have an older Mac running Mojave, but I've been using the old, deprecated (?) Java version... PasswordSafe SWT IIRC.  I still use the machine for a couple of things; I'd be OK using it as a testbed, but I'd need to make sure I could get back to the SWT version. 

So... how do I go about this... where do I find the code to d/l, install & test??

I'd like to begin testing on my new machine. I can test on the Mojave system once I get through the Ventura testing & have verified I can get recover SWT if needed.

nobugshere:
I'm using the current head of the master branch as my daily driver on my M1 Ventura system.  You don't have to remove 1.17, just place this version in a different folder; it will use the same config file.  It's universal, so the same app should work on both systems, it does work under Rosetta, at least.  I don't know anything about the Java version, so I can't say whether or not there are any compatibility issues with the .cfg or .psafe3 files. 

You have good backups, right? :-)

Unless @ronys objects, you can get it here:
https://github.com/pwsafe/pwsafe/suites/15259885708/artifacts/871826679

The About dialog identifies as: "v1.17.1 (g88ee23d) OSX"

Issue #1014, opened 2023-08-30
Password input field is not foreground with focus after restore when Password Safe is in taskbar mode (i.e., when "Put icon in System Tray" is disabled/unchecked).


AshleyT3:
Password input field is not foreground with focus after restore when Password Safe is in taskbar mode (i.e., when "Put icon in System Tray" is disabled/unchecked).

OS: Windows 10 and Windows 11 with version v3.63 but issue goes back to circa 2012, observed on Windows 7 at that time.

Steps:

1. Install Password Safe.
2. Manage, Options, System, uncheck "Put icon in System Tray" to disable system tray usage.
3. Lock Password Safe. You should see its icon remain in the taskbar given configuration step 2.
4. Single click Password Safe taskbar icon and, on repro, notice Password Safe restores to the password entry dialog but the password field does not have focus.

PR #1015, opened 2023-08-30
Keep pwsafe foreground on taskbar activation.

AshleyT3:
When a locked pwsafe is activated by a user taskbar click, it should not hide its window because doing relinquishes its foreground status, preventing the password input dialog from becomming foreground window. This affects pwsafe when "Put icon in System Tray" is unchecked. https://github.com/pwsafe/pwsafe/issues/1014


AshleyT3:
Closing this PR. This is not the right fix, and there are other issues which should be addressed first which will likely naturally include refactoring addressing this issue. Details with suggestions presented to @ronys for consideration.

PR #1016, opened 2023-08-31
password entry is foreground window when DB locked

AshleyT3:
When pwsafe DB is locked while pwsafe is in non-systray "taskbar" mode, the password entry dialog becomes the pwsafe app window. This allows pwsafe DB unlock on activation to present the password entry dialog as the foreground window with focus.
https://github.com/pwsafe/pwsafe/issues/1014


src/ui/Windows/MainView.cpp
```diff
@@ -2558,7 +2558,11 @@ void DboxMain::OnTimer(UINT_PTR nIDEvent)
     if (prefs->GetPref(PWSprefs::UseSystemTray)) {
```
ronys, line 2558: I've negated the predicate and moved the call to ShowWindow(SW_HIDE) outside of the if statement to make it a bit clearer.

src/ui/Windows/PasswordSafe.rc
ronys, line 1: No idea why a couple of minor changes made git mark this entire file as changed. Line endings?

ronys: Line endings it is. Fixing.

ronys:
Nice work - thanks!

src/ui/Windows/PasskeyEntry.cpp
```diff
@@ -259,13 +274,6 @@ BOOL CPasskeyEntry::OnInitDialog(void)
     return FALSE;
   }
 
-  // Following works fine for other (non-hotkey) cases:
-  if (m_index == GCP_RESTORE || m_index == GCP_WITHEXIT) {
-    SetWindowPos(&CWnd::wndTopMost, 0, 0, 0, 0, SWP_NOMOVE | SWP_NOSIZE);
-    SetActiveWindow();
-    SetForegroundWindow();
-  }
```
dskrvk, line 267: This blocks seems to have been removed completely - why?

AshleyT3: It was redundant.

PR #1017, opened 2023-09-01
[wx][FR909] Search (Find) bar visibility is now persistent

rafaelx:


ronys:
Thanks @rafaelx - I've added a line in the wx release notes documenting this.

Issue #1022, opened 2023-09-06
Exclude Password Safe from screen capture.


AshleyT3:
Use Windows features to block screen capture of sensitive Password Safe windows to whatever extent achieved by Windows.  This will obviously not prevent physical photographs, or anything circumventing Windows prevention of screen capture (i.e., ring-0 access to display data, but will generally prevent typical user land programmatic screen capture, including capture via PrtScr, Snipping tool, and PowerShell scripts using Drawing.Graphics.

PR #1023, opened 2023-09-06
Exclude Password Safe from screen capture.

AshleyT3:
Excluding pwsafe from OS screen capture is applicable to the Windows build of pwsafe to whatever extent the OS can prevent such. It obviously does not prevent physical photographs, or any access to screen data (i.e., ring 0 access bypassing OS blocks of screen capture) but will generally prevent typical user land automated/programmatic screen capture, including PrtScr key, Snipping tool, and PowerShell scripts using Drawing.Graphics. https://github.com/pwsafe/pwsafe/issues/1022


src/ui/Windows/ThisMfcApp.h
```diff
@@ -24,6 +24,17 @@
 #include <afxmt.h>
 //-----------------------------------------------------------------------------
 
+inline void SetWindowExcludeFromScreenCapture(HWND hwnd) {
```
ronys, line 27: I think it would make more sense to put this under winutils.h.

ronys:
Where is the screen capture blocked unconditionally for the opening master password prompt?

AshleyT3:
@ronys: > Where is the screen capture blocked unconditionally for the opening master password prompt?

It is enabled (true) by default...

[PWSprefs.cpp@143](https://github.com/pwsafe/pwsafe/pull/1023/files#diff-0078912ffed701e8f978342c77688f5a775b1c094aea7688453200083b151750R143):
```
const PWSprefs::boolPref PWSprefs::m_bool_prefs[NumBoolPrefs] = {
...
  {_T("ExcludeFromScreenCapture"), true, ptDatabase},       // database
};
```

[Copy_PWS_prefs.cpp@100](https://github.com/pwsafe/pwsafe/pull/1023/files#diff-1b0da6b18e6d3de01dbd1cb60d6527149b419c0199253a7d13edcfdd6ac69a7fR100):
```
const PWSprefs::boolPref PWSprefs::bool_prefs[NumBoolPrefs] = {
...
  {_T("ExcludeFromScreenCapture"), true, ptDatabase},       // database
};
```

ronys:
> @ronys: > Where is the screen capture blocked unconditionally for the opening master password prompt?
> 
> It is enabled (true) by default...
> 

Hm.
Relying on the default value for this behavior's problematic:
1. Someone changing the default value in code will break this without any indication.
2. More realistically, consider the following scenario:
- User opens a database where this preference is set to false
- User closes the database. Preference remains in effect, (IIRC, there's no revert-to-default for dbase prefs upon db close)
- User opens another (or even same) database. Since preference is now 'false', the master password dbox is not protected from screen capture.

Adding a forceExclude bool parameter with default false to WinUtil::SetWindowExcludeFromScreenCapture might be a good solution.

AshleyT3:
> Hm. Relying on the default value for this behavior's problematic:
> 
> 1. Someone changing the default value in code will break this without any indication.

I think you may be expressing concern #&NoBreak;1 in relation to concern #&NoBreak;2 so I suggest we resolve #&NoBreak;2 first which may help discuss #&NoBreak;1. I say this because concern #&NoBreak;1, when viewed alone, does not offhand seem like a problem since any code change may break a feature without any indication, and changing a default typically demands close scrutiny.

Concern #&NoBreak;2 seems valid... _**good catch**_ ...see comments below the following quote...

> 2. More realistically, consider the following scenario:
> 
> * User opens a database where this preference is set to false
> * User closes the database. Preference remains in effect, (IIRC, there's no revert-to-default for dbase prefs upon db close)
> * User opens another (or even same) database. Since preference is now 'false', the master password dbox is not protected from screen capture.
> 
> Adding a forceExclude bool parameter with default false to WinUtil::SetWindowExcludeFromScreenCapture might be a good solution.

Your observations reveal the state transitions leading to DB Close should certainly cause a reset to DB pref defaults. Today that does not happen. This is a bug removed from this PR but perhaps revealed by this PR's feature being visible after DB Close.

We could include the fix in this PR but I suggest we resolve the "DB pref reset to defaults upon DB Close" issue in a separate PR first... I can merge the fix into this PR for cleaner consideration of this PR... your call.

It seems pwsafe is meant to use a DB's prefs while the DB is locked. If that is accurate, I do not see any more issues.

The above rationale is derived from the following...

Current pwsafe state transition behavior...

|State|Today|Desired|Comment|
|------|------|--------|-----------|
|DB Lock -> DB Open|DB pref ✅|DB pref|Seems correct.|
|DB Open -> DB Lock|DB pref ❓|DB pref or default pref?|Seems "by design" that DB pref is active during lock.|
|DB Open -> DB Close|DB pref ❌|should be DB _default_ pref|Seems like a bug, we should fix.|
|DB Lock -> DB Close|DB pref ❌|should be DB _default_ pref|Seems like a bug, we should fix.|

Unrelated to this PR, we need to fix "DB Open -> DB Close" and "DB Lock -> DB Close" by resetting the current DB prefs to DB default prefs. More briefly, closing a DB should revert DB prefs to DB default prefs.

That leaves "DB Open -> DB Lock" ... it seems we consider it "by design" to leave a DB's prefs active while it is locked for the following reasons...

- DB prefs can affect UX around DB usage whether locked or open (i.e., prefs for features adding/removing menu options available while a DB is locked, features such as ExcludeFromScreenCapture, or options otherwise affecting the unlock UX for that one DB.)
- DB prefs do not seem to contain secure information requiring scrubbing when locked (otherwise both transition to Lock and Close should reset to DB default prefs). 

If you agree a DB's prefs are "by design" still active while a DB is locked, then ExcludeFromScreenCapture should maintain its DB pref setting while a DB is locked, meaning the PR is fine as is in this regard. If you agree, it would only leave the one case of fixing DB prefs to reset to DB default prefs upon transition from either DB Lock/Open to DB Close state (all transitions to Close scrub DB prefs to be DB default prefs).

AshleyT3:
@ronys Whether fixed as part of this PR or a separate PR, here is [Issue 1025 - Reset to pwsafe DB default preferences upon DB close](https://github.com/pwsafe/pwsafe/issues/1025). Since ExcludeFromScreenCapture may be the only post-close feature that is visible, it may be easier to fix as part of this PR after all. Offhand, I did not see other DB-bound options in effect during close... so maybe this is a first. (?) Let me know if not. I will begin fixing as part of this PR but let me know if you wish to keep it separate.

AshleyT3:
@ronys See [latest commit](https://github.com/pwsafe/pwsafe/pull/1023/commits/b472ddd37971faba73b0d6cff5c1fcd56119ccc2) for "reset DB prefs upon DB close" change. This variation is based on the assumption you consider it "by design" that pwsafe retains/uses a DB's prefs upon transition to Lock, that only a transition to Close should reset DB prefs to the default state.

ronys:
Testing this remotely uncovered an unexpected side effect: Looks like VNC relies on screen capture, since I can't view the opening password prompt on my VNC viewer.

Thoughts?

AshleyT3:
> Testing this remotely uncovered an unexpected side effect: Looks like VNC relies on screen capture, since I can't view the opening password prompt on my VNC viewer.
> 
> Thoughts?

My vote is to default ExcludeFromScreenCapture to FALSE/off, let the user opt-in from the options page if desired, and whenever opting in, present a message box warning "are you sure? Some apps such as VNC..." prompt before accepting the user's check to opt-in. (When notifying users of new pwsafe in emails or release notes, perhaps mention new opt.)

There could be a post-install/first-run prompt "Do you want to enable" (with same warnings)... but this can be annoying... safest/cleanest for initial release is former without this.

The ExcludeFromScreenCapture has long _not_ been in pwsafe, so given it may cause issues with desired screen capture apps (VNC or otherwise), it really seems safest to keep it off by default, let the user opt-in if desired. This allows for a no harm/foul release, and we can always change things later.

Related to prior... it's important to consider what it protects against... I do not see it helping for the case of foothold as admin, for example. It seems most beneficial for the case of a silent (non-visible) attack as a non-admin user, where visual access to pwsafe can act as reconnaissance for potential escalation (i.e., a mini PowerShell streaming server to attacker).

My guess is you considered the above already...  and you know your pwsafe users... What are your thoughts on this?

ronys:
Well, I'm not happy with leaving a preference with a default value that's less secure, but (a) it's the current state (b) the failure mode for VNC is nasty.

Is there a way to detect VNC programmatically? That could help. If not, I propose:
1. Setting the default to "false", as you wrote
2. Adding a blue (?) that provides a tooltip with a descriptive text
3. Adding a command-line option '--allow-screen-capture=1|0' to override the preference (including master password prompt!), for the case where a user didn't RTFM and found herself locked out... (hoping that they'll contact support)

AshleyT3:
> Well, I'm not happy with leaving a preference with a default value that's less secure, but (a) it's the current state (b) the failure mode for VNC is nasty.
> 
> Is there a way to detect VNC programmatically? That could help. If not, I propose:
> 
> 1. Setting the default to "false", as you wrote
> 2. Adding a blue (?) that provides a tooltip with a descriptive text
> 3. Adding a command-line option '--allow-screen-capture=1|0' to override the preference (including master password prompt!), for the case where a user didn't RTFM and found herself locked out... (hoping that they'll contact support)

I considered VNC detection but it ultimately felt silly after considering a number of things, maybe you can check me on this...

Generally, it seems not advisable to conditionally silently auto-enable/disable pwsafe options affecting security, but especially not in cases where only imperfect detection methods are available to arrive at guiding condition(s).

Pwsafe users may or may not want ExcludeFromScreenCapture (EFSC) enabled/disabled when VNC is in use by design or unexpectedly. Detection of the condition of VNC being present is one thing, the user's desire choice another which cannot be predicted.

For example, a pwsafe user may wish to have EFSC disabled when they install VNC, while some other pwsafe user may enable EFSC and want it to block VNC should it ever be running (whether the user is aware of VNC or not... but perhaps more so if the user is unaware).

It therefore seems VNC detection should, if anything at all, only guide enable/disable decisions through firm user awareness (i.e., IF vnc_present THEN prompt pwsafe user for explicit opt-in/out from an otherwise known concrete default).

Perhaps if detection around this were more relaible and a user's desire based on any detections were firmly predictable, potentials would be differnt... but it seems VNC detection will not be certain, and a pwsafe user's desire based on such would not be certain.

Probably TMI but VNC detection adds the following burdens to the pwsafe project...

1. It puts pwsafe in the business of relying on, maintaining detection methods which can change over time, creating a form of cross-project dependency.

2. We would have to research detection approaches viable from pwsafe, a non-elevated user land app. Offhand (research required), pwsafe's detection methods would be based on queries of registry, process info, and/or file-presence, all of which would be subject to permission issues making its detection method imperfect with some room for error. This would be in addition to imperfections induced by #&NoBreak;1 above.

3. VNC is just one app... there may be other apps with user-desired/undesired screen capture.

Regarding your suggestions... 

1. Default to false makes sense.

2. A tooltip sounds fine... but you do not want a Yes/No prompt if changing that option's state? User should not be changing this often, so a prompt won't be a pain, may certainly prevent problem usage.

3. A command line option... are you certain on that? Correct me if wrong, but doesn't that make it easy for an attacker with non-elevated access to restart pwsafe with the feature disabled? By keeping a pwsafe user's choice in the DB, it is only made active upon unlock, not overridable from command line. Since a typical non-admin non-elevated attacker does not have the SeDebugPrivilege ("Debug programs") privilege, and cannot debug pwsafe memory, it seems best to disallow potential non-admin non-elevated EFSC disable overrides.

Let me bounce your direction that we consider defaulting to false (as you say), tooltip (as you say), yes/no prompt (+1 your direction, costs little, is very clear), and no command line option (if user enables what is disabled by default, and has tooltip, and is prompted, they have had the right chance, no? What do you think?).

AshleyT3:
(addendum to prior)
Another possibility for you to consider... provide the command line override you specified but implement it to require elevation, else return access denied.

While I generally would not recommend using pwsafe elevated, it could be used to allow the user to get out of a misconfiguration bind so long as they have access to an account allowing elevation. After the user fixes the misconfiguration, they can restart pwsafe normally (without elevation).

This would only allow command line override through the use of an account where excluding capture has far less if any value.

ronys:
Answering briefly to move forward:
1. We (reluctantly) agree to default to false.
2. A descriptive tooltip shall be added
3. I'm averse to confirmation dialogs. Experience shows that those who need them most are least likely to read them.
4. Agree with your reservations re command line switch. The only reason I still see a need is because a user with a db where EFSC is true is 100% locked out if/when she only has VNC access. Perhaps display a warning dialog box when this is set (yes, I know I just contradicted myself...)?

AshleyT3:
> Answering briefly to move forward:
> 
> 1. We (reluctantly) agree to default to false.
> 2. A descriptive tooltip shall be added
> 3. I'm averse to confirmation dialogs. Experience shows that those who need them most are least likely to read them.
> 4. Agree with your reservations re command line switch. The only reason I still see a need is because a user with a db where EFSC is true is 100% locked out if/when she only has VNC access. Perhaps display a warning dialog box when this is set (yes, I know I just contradicted myself...)?

To confirm... while you still see a need for a command line override, you agree we should not fulfill that need, but to instead warn with confirmation dialog upon EFSC enable?

AshleyT3:
...and you do not want command line switch override only if user elevated? I assume not but checking again.

nobugshere:
I don't have any experience in this area, but if VNC can't display the EFSC windows, might it also be a problem for screen reader software?

ronys:
Correct re not bothering with elevated users. That would be a first use of this distinction in pwsafe, don't think it's justified.
Probably true re screen-readers - good thought.

AshleyT3:
> Correct re not bothering with elevated users. That would be a first use of this distinction in pwsafe, don't think it's justified.

Thanks.

> I don't have any experience in this area, but if VNC can't display the EFSC windows, might it also be a problem for screen reader software?

> Probably true re screen-readers - good thought.

Ditto, good thought...

I believe screen reader software uses a combination of UI text extraction and OCR depending on the software and what it is reading. I would gather the OCR portion would be impeded with EFSC enabled, not the UI text extraction-based reading. Given Windows has a focus on accessibility, it may have features supporting OCR-based reading when EFSC is enabled. Of course, a test with any given screen reading software is always in order to understand any limitations in relation to EFSC.

src/ui/Windows/WZPropertySheet.cpp
```diff
@@ -113,6 +114,13 @@ INT_PTR CWZPropertySheet::DoModal()
   return rc;
 }
 
+BOOL CWZPropertySheet::OnInitDialog()
```
dskrvk, line 117: I'm wondering if we could define a base class for all pwsafe dialogs and perform common init actions there rather than repeat them in every dialog subclass.

ronys: IIRC, CDialog, propsheets and "wizard" propsheets don't share a useful common ancestor for this.

ronys:
The feature works nicely. However:
- The initial password dialog now is never protected.
- I'm unhappy about the warning dialog box. Consider:

A user who uses pwsafe on a Windows VM via VNC can lock herself out inadvertently and irreversibly by setting EFSC. Relying on a tooltip and/or warning dbox when the preference is set isn't enough, IMOH.
A command line "emergency override", therefore, seems in place.
To address the valid concern that the command line may be "hidden" by an adversary, e.g., via the Windows shortcut properties, I suggest that a warning dialog box be displayed _upon application startup_ *when the command line flag is set*. Something like "Screen capture protection temporarily disabled. Clear the checkbox in Options->Security to make this permanent". This should alert the user that something's amiss.

AshleyT3:
> The feature works nicely. However:
> 
> * The initial password dialog now is never protected.
> * I'm unhappy about the warning dialog box. Consider:
> 
> A user who uses pwsafe on a Windows VM via VNC can lock herself out inadvertently and irreversibly by setting EFSC. Relying on a tooltip and/or warning dbox when the preference is set isn't enough, IMOH. A command line "emergency override", therefore, seems in place. To address the valid concern that the command line may be "hidden" by an adversary, e.g., via the Windows shortcut properties, I suggest that a warning dialog box be displayed _upon application startup_ _when the command line flag is set_. Something like "Screen capture protection temporarily disabled. Clear the checkbox in Options->Security to make this permanent". This should alert the user that something's amiss.

Your concerns make sense, let me bounce the following off you to double-check, consider... 

Somewhat in reverse order of your list... a command line override will defeat what seems to be the only reasonable scenario for having the feature, which is to guard against screen capture by a non-admin account in user land which has a foothold or access at some point in time prior to or during legit pwsafe usage.

If attacker is elevated admin, generally speaking, it is game over so we agree we are not talking about that. 

Requiring the user to click a button to proceed forward with the override is not sufficient given an attacker can simply automate the click (i.e., as part of the rogue shortcut)... unless...

If the override prompt dialog you desire required the user to enter a captcha not readable by another (non-admin) process, such might be sufficient.

Would you be against a captcha in that override prompt dialog?

So the options as I see them might be as follows...
- Requiring an elevated admin to use the command line override.  (You nixed this already.)
- Requiring user to enter a captcha in a command line override prompt dialog. (Pending your review.) The captcha does not have to be complex... just not readable from a non-admin querying window text, as one example. 

If not the above, maybe you have further thoughts.

The following are some brainstorming points which can be taken lightly, skimmed... just wondering if they raise any good considerations...

- Do not worry about the initial Master Password Entry and the like. The rule could be, a user must log into their DB to use the EFSC feature, that it largely pertains to a best-effort against screenshots of post-DB-auth password group/entry information.
- For command line override with simple prompt box (without captcha) that you would like to see, instead of requiring captcha or elevated admin, present a simple prompt the user as you recommended but afterwards, have some visual cue in pwsafe that the command line override is active such as a different color frame or blinking red block in the status bar, something like that. Note, I think frame colorizing may require a bit of churn (i.e., WM_NCPAINT) if no APIs for supported OSes ease such... this is why I mentioned a block on the status bar... that's simple. 
- If VNC is your primary use case of concern with user lockout, perhaps we should revisit the notion of VNC detection, where we implement such in a manner to allow us to include additional applications at a later date as needed. This way, we can prevent use of override unless VNC is in use, and for such cases, perhaps have a visual cue or captcha as discussed above.

AshleyT3:
...an addendum... something to consider, maybe obvious, but re capthca... it can be screen captured in the scenario outlined. To be useful, it seems it would require either an active attacker or some sort of OCR, all of which would have to act quickly to get past the dialog without the user seeing... doubtful that would be likely... so perhaps captcha is still viable... and we can add visual cue when pwsafe has been started w/common line override. (?)

ronys:
A captcha seems excessive, both in implementation and cumberness to the end user.
A second or two delay between displaying the dialog and enabling the close/OK button seems to me a reasonable compromise. This seems sufficient to me.

AshleyT3:
> A second or two delay between displaying the dialog and enabling the close/OK button seems to me a reasonable compromise. This seems sufficient to me.

What about the case where attacker uses these features before a user returns to their PC/laptop. The user sees an already running pwsafe as usual, opens it, unlocks DB, user thinks EFSC is enabled (screen capture disabled).

It seems there should not be a 100% quiet command line override of EFSC.

I'm wondering, if in addition to any visual cue while override is active, if the EFSC command line override warning dialog should appear both upon start of pwsafe as well as upon (just after) _any_ master password unlock. (?) Yes, redundant but seems required for case of attacker relaunching pwsafe with override while legit user absent... we do not want user to unlock DB without knowing override is active.

Recap on above:
- Employ command line override.
- When override active, pwsafe start prompts user with warning that cannot be dismissed for a couple of seconds.
- When override active, pwsafe DB unlock prompts user with same warning.
- When override active, pwsafe shows visual cue somewhere (i.e., status bar).
- Reverse current EFSC default from disabled back to enabled (because the override and warning prompts being discussed are about doing just that, allowing for a default of EFSC enabled). 
- The legit user can can disable EFSC if desired.
- Upon DB unlock of any DB having EFSC disabled (in DB prefs), any command line override becomes inert, skipping the post-DB-unlock warning prompt (in this case, only one EFSC warning prompt is presented upon pwsafe launch w/command line override active).

This seems to cover the following use cases:
- **Today's traditional desktop Windows pwsafe user** who wants OS screen capture blocking always on. They do not care about EFSC, so will never use override, things just work (flipping EFSC default back to 'true' sort of does this alone... what we first had... probably the most common use case [assumptive]).
- **Desktop user with VNC who forgets about EFSC**... this user re-lauches pwsafe with override, deals with prompts ... but such is acceptable since VNC is an an exceptional case for this user, or at least one where the prompts are fine.
- 100% VNC user always overrides pwsafe launch, sets any DBs used via VNC to have EFSC disabled. They only see the EFSC disabled warning prompt upon launch.

What do you think on this?

(forgive if I'm overthinking anything here but these offhand seem worthwhile to align on.)

I'm not sure if it helps to state this, but all of this is about making use of the OS EFSC feature to mitigate non-admin user capture of sensitive information, perhaps info allowing that very non-admin weaker foothold to escalate in privileges toward/to admin. It's probably the most likely case so is worthwhile despite not being comprehensive. That said, generally speaking, due to supply-chain monitor attacks, other ways to capture the display stream, high-value targets such as political leaders, activists, etc., should never be making passwords visible to the display display / display stream regardless of EFSC. I all this is obvious but mentioning fwiw.

AshleyT3:
...an addendum: For the case of the command line override being active during DB unlock which leads to a second post-DB-unlock warning prompt, we can skip that second DB unlock prompt in cases where the pwsafe startup w/override prompt was just dismissed just seconds or a minute prior.

ronys:
I tend to agree with most of your recap. Here's how I see it:
- The override flag is not persistent, meaning it doesn't affect the stored preference. This makes it useful for users in a hybrid scenario, who work both via VNC and directly.
- Visual cue in the status bar: Should have a tooltip "Screen capture protection overridden". Come to think of it, perhaps we should add a permanent tri-state indicator for screen capture protection: on/off/overridden?
- When opening a subsequent DB with the flag set, consider that the new DB might have the preference off. No need to prompt in this case.

AshleyT3:
> I tend to agree with most of your recap. Here's how I see it:
> 
> * The override flag is not persistent, meaning it doesn't affect the stored preference. This makes it useful for users in a hybrid scenario, who work both via VNC and directly.
> * Visual cue in the status bar: Should have a tooltip "Screen capture protection overridden". Come to think of it, perhaps we should add a permanent tri-state indicator for screen capture protection: on/off/overridden?
> * When opening a subsequent DB with the flag set, consider that the new DB might have the preference off. No need to prompt in this case.

@ronys, Some decisions:

- Completed: ExcludeFromScreenCapture defaults to enabled/true.
- Completed: Command line added: --allow-screen-capture
- TO DO: Status bar indicator.
- TO DO: --allow-screen-capture startup and post-DB-unlock timed warning dialog.

The status bar indicator is tri-state:

- Screen capture disallowed: normal/default.
- Screen capture allowed: normal user pref change to disable ExcludeFromScreenCapture.
- Screen capture force allowed: exceptional command-line user override.

Questions:
1) Which status bar location would you like for this indicator? I have started with the following location, between num ent and filter...
```... SB_NUM_ENT, SB_SCR_CAP, SB_FILTER...```
2) Would you prefer to see those states represented on the status bar as strings like `"R/W", "R-O"`, or 3 different bitmaps?

|                     State                                     |          String ID       | String                       |Bitmap(s)|
|----------------------------------------------|---------------------|-----------------------|----------|
| capture disallowed                                    | IDS_NO_SCR_CAP  | "✔️No Cap"   |  PC monitor with green X across the display (see mock below)|
|capture allowed via user DB pref setting   | IDS_SCR_CAP         | "☐Capture"    |  PC monitor filled with mild yellow (see mock below).|
|capture force allowed (cmd line override) |IDS_FRC_SCR_CAP   | "❌Capture"  | PC monitor filled with strong yellow on red background, perhaps alternating w/steady cursor-like blink. | 
* See mock below but, for textual indicator route, emojis used appear as black-and-white, not colorful like above table, assuming your browser shows colors. This seems fine as the indicators are supposed to show "generally OK," something "different but maybe not bad", and "warning warning."
* For text route, we can blink ❌Capture with a red background. For bitmap route, we can blink bitmap... see mock notes below.
* For bitmap route, we can start with dev bitmaps lest you know of some, where we can always replace w/artist's if desired later... IOW, we can move forward without the best art so long as its reasonable. 

The following screenshot shows 3 pwsafe mock dialogs, each with a different status bar capture string in the following order: No Cap, Capture, XCapture....
image


The following screenshot shows 3 pwsafe mock dialogs, each with a different status bar capture icon in the following order: No Cap, Capture, XCapture....
image
* The bitmaps shown are examples for mocks discussed here but not for our use for feature as they are derived from emoji renderings which I do not think we can simply derive from... so if you want bitmaps, I can create placeholder, if they look bad, we can replace with artist's bitmaps... but the placeholders will allow feature impl to move forward.
* Re prior bullet, if you are wondering, instead of bitmaps, why not use that desktop computer emoji directly, perhaps with green, yellow, red backgrounds or some variation like that... reason is the emjoi is "desktop computer" and not "computer monitor" so no guarantee it will look like a monitor as shown... but it did look like one for me so good for one-off mock for this discussion.
* The concern of prior bullet does not apply to the emojis used in textual indicators because those emojis should look something as shown as they are defined as check, empty ballot box, and red X. 

What do you think is right here?

I sort of like the bitmaps... but text is easier and it actually does not look bad, and may well be clearer. Curious what your thoughts are.

**Comment by ronys** [comment]
Bitmaps are fine, especially since they save us some translation effort.
As to the images, I'd go for something small and minimalistic, such as https://icons8.com/icon/XJZDcGJaksIF/full-screen

**Comment by AshleyT3** [comment]
@ronys latest change ([d6ca4a9](https://github.com/pwsafe/pwsafe/pull/1023/commits/d6ca4a965245567447dbcdc2b25436a834ddb287)) has changes less startup timed dialog, post-unlock DB timed dialog, and password entry indicator, all in-progress now.

**Comment by AshleyT3** [comment]
@ronys commit [85b1539](https://github.com/pwsafe/pwsafe/pull/1023/commits/85b1539cd75a8011afec58626dffe42fca66c624) adds timed '--allow-screen-capture' warning during pwsafe startup. Changes to date stand alone. Remaining is master password indicator.

**Comment by ronys** [comment]
Suggest changing the flashing red indicator to flash for, say, 10 seconds after DB is opened, then remain a steady red.

**Comment by ronys** [comment]
Nice. One nit:
In the Change Master Password dialog box, the EFSC indicator overlaps with the keyboard (see below)
image

Other that that, and text revision, I think we're ready to merge.

File: src/ui/Windows/StateBitmapControl.cpp [hunk 3]
```diff
@@ -0,0 +1,220 @@
+/*
+* Copyright (c) 2003-2023 Rony Shapiro <ronys@pwsafe.org>.
+* All rights reserved. Use of the code is allowed under the
+* Artistic License 2.0 terms, as specified in the LICENSE file
+* distributed with this code, or available from
+* http://www.opensource.org/licenses/artistic-license-2.0.php
+*/
+
+#pragma once
```
ronys, line 9: "#pragma once" doesn't really make sense for .cpp files:
ExcludeCaptureStateBitmapManager.cpp
ScreenCaptureStateControl.cpp
StateBitmapControl.cpp
StateBitmapManager.cpp

AshleyT3:
> Nice. One nit: In the Change Master Password dialog box, the EFSC indicator overlaps with the keyboard (see below) image
> 
> Other than that, and text revision, I think we're ready to merge.

@ronys, virtual keyboard high scaling overlap issue addressed in commit [9422415](https://github.com/pwsafe/pwsafe/pull/1023/commits/94224153065d1ffedc1a35a1732406eb429b9980)

src/ui/Windows/StateBitmapControl.cpp
AshleyT3: > "#pragma once" doesn't really make sense for .cpp files: ExcludeCaptureStateBitmapManager.cpp ScreenCaptureStateControl.cpp StateBitmapControl.cpp StateBitmapManager.cpp

So true... so so true! 💡

ronys:
Merging into master, documentation will be added separately.

Issue #882, opened 2022-09-29
version 1.4/1.5 stopped working after recent Mint 21 update


Sirjeff2000:
I recently did a fresh install on an HP laptop (AMD processor) and installed version 1.14 of pwsafe. It worked fine until I applied the most recent updates. I started having a problem that after launching pwsafe, I could no longer type in my password and PWsafe would freeze. Out of desperation, I installed the windows version through Playonix so that I cold access my passwords. I have no idea what caused the problem but I have anpother HP laptop with an i7 processor that works flawlessly even after the latest update. Any suggestions would be greatly appreciated.

ronys:
Is your other laptop running the same version of Mint?
Do the system logs (/var/log/messages, /var/log/syslog, etc.) show any errors?

peterkelley519:
Hi - I've been having a similar (or same) problem. I'm trying to piece together what happened. I think I had version 1.14 of pwsafe and mid-September updated to 1.15.  I'm running Linux Pop!OS 22.04 . The problem started around the same time, mid-September, but sometimes (only occasionally) behaves normally. 
I open the PasswordSafe, and it seems to be frozen, but I've found it's actually extremely slow to take the combination in, or any other mouseclicks etc. (30-90 second delay, sometimes indefinite).
No other applications are behaving this way, so I'm trying to figure out if it's got to do with my system or something particular that's happening to PwSafe.
Is there anything in particualr to look for in the syslog, or another log/setting?
Best Regards!

peterkelley519:
Not two minutes after I posted last, I was looking at /var/log/syslog and saw messages concerning my USB Hub, and though I didn't think it was likely related to my issue. But I figured I should try disconnecting the USB Hub. I opened up PWSafe and to my surprise it worked normally! closed PwSafe and then plugged the Hub in again, reopened PwSafe and it was stuck again. Repeated, because...science. Results confirmed - it appears something about my usb hub is messing with the input of pwsafe.
I have no idea what the significance of the messages are (they look like err msgs). I'll attach them.
Also - I have found no other applications that are affected by this.
@Sirjeff2000  - any chance it's something similar? 
and 
@ronys - Thank you for all the work you do with this! Any insight regarding the error msg?
Pete
[var_log_syslog_segment.log](https://github.com/pwsafe/pwsafe/files/9815645/var_log_syslog_segment.log)

peterkelley519:
Getting closer to the [possible] cause of the problem. 
[...]

PR #1024, opened 2023-09-06
Make YubiKey polling interval configurable

rafaelx:
This is an untested attempt to make the Yubi Key polling interval configurable via command line argument "-y".
This PR is related to the suggestion in #882 and has not been tested due to the lack of a Yubi Key.

1. Should the command line argument be "-y, --yubi_polling_interval" or "-p, --polling"?
2. Should an environment variable be used rather than a command line option?

### DIFF
--- a/src/ui/wxWidgets/DbSelectionPanel.cpp
+++ b/src/ui/wxWidgets/DbSelectionPanel.cpp
@@ -114,8 +114,10 @@ DbSelectionPanel::DbSelectionPanel(wxWindow* parent,
 #ifndef NO_YUBI
   SetupMixin(FindWindow(ID_YUBIBTN), FindWindow(ID_YUBISTATUS));
   Bind(wxEVT_TIMER, &DbSelectionPanel::OnPollingTimer, this);
-  m_pollingTimer = new wxTimer(this, POLLING_TIMER_ID);
-  m_pollingTimer->Start(YubiMixin::POLLING_INTERVAL);
+  if (YubiMixin::IsPollingEnabled()) {
+    m_pollingTimer = new wxTimer(this, POLLING_TIMER_ID);
+    m_pollingTimer->Start(YubiMixin::GetPollingInterval());
+  }
 #endif
   
   //The parent window must call our TransferDataToWindow and TransferDataFromWindow
@@ -124,7 +126,10 @@ DbSelectionPanel::DbSelectionPanel(wxWindow* parent,
 
 DbSelectionPanel::~DbSelectionPanel()
 {
-  delete m_pollingTimer;
+  if (m_pollingTimer != nullptr) {
+    delete m_pollingTimer;
+    m_pollingTimer = nullptr;
+  }
 }
 
 void DbSelectionPanel::SelectCombinationText()
--- a/src/ui/wxWidgets/DbSelectionPanel.cpp
+++ b/src/ui/wxWidgets/DbSelectionPanel.cpp
@@ -126,10 +126,7 @@ DbSelectionPanel::DbSelectionPanel(wxWindow* parent,
 
 DbSelectionPanel::~DbSelectionPanel()
 {
-  if (m_pollingTimer != nullptr) {
-    delete m_pollingTimer;
-    m_pollingTimer = nullptr;
-  }
+  delete m_pollingTimer;
 }
 
 void DbSelectionPanel::SelectCombinationText()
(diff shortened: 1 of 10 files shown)


ronys:
How's the first commit related? Was it added to this branch by mistake?

I'd start by having this only as a command-line argument, no env var, if for no other reason than it's more clearly documented (via --help).
For consistency, I'd call it '--yubi-poll-interval' ('-', not '_'). If possible, I'd also disable the "short" option ("-y"), as it's not going to be used very often.

src/ui/wxWidgets/SafeCombinationSetupDlg.cpp
```diff
@@ -84,7 +84,9 @@ SafeCombinationSetupDlg::SafeCombinationSetupDlg(wxWindow *parent, wxWindowID id
 #ifndef NO_YUBI
   SetupMixin(FindWindow(ID_YUBIBTN), FindWindow(ID_YUBISTATUS));
   m_pollingTimer = new wxTimer(this, POLLING_TIMER_ID);
```
dskrvk, line 86: Do we even need to initialize the timer if polling isn't enabled?

src/ui/wxWidgets/YubiMixin.cpp
```diff
@@ -81,6 +84,25 @@ void YubiMixin::HandlePollingTimer()
...
+    s_pollingInterval = YubiMixin::POLLING_INTERVAL_MIN; // limit the polling interval to a minimum
+    s_pollingEnabled = true;
+  }
+  else if (value > YubiMixin::POLLING_INTERVAL_MAX /* ms */) {
+    s_pollingInterval = YubiMixin::POLLING_INTERVAL_MAX; // limit the polling interval to a maximum
+    s_pollingEnabled = true;
+  }
+  else {
+    s_pollingInterval = value;
+    s_pollingEnabled = true;
```
dskrvk, line 102: Nit: if the only case when polling is disabled is the < 0 case, why not move this repeating line out of the if branches? E.g.
```c++
  s_pollingEnabled = true;
  if (value <= 0) {
    s_pollingInterval = YubiMixin::POLLING_INTERVAL;
    s_pollingEnabled = false;
  }
```

src/ui/wxWidgets/YubiMixin.h
```diff
@@ -25,7 +25,7 @@ class wxWindow;
 class YubiMixin
 {
  public:
-  enum {POLLING_INTERVAL = 500}; // mSec
+  enum {POLLING_INTERVAL_MIN = 100, POLLING_INTERVAL = 500, POLLING_INTERVAL_MAX = 60000}; // mSec
```
dskrvk, line 28: Nit: maybe rename POLLING_INTERVAL to POLLING_INTERVAL_DEFAULT?

rafaelx:
> How's the first commit related? Was it added to this branch by mistake?
> 
> I'd start by having this only as a command-line argument, no env var, if for no other reason than it's more clearly documented (via --help). For consistency, I'd call it '--yubi-poll-interval' ('-', not '_'). If possible, I'd also disable the "short" option ("-y"), as it's not going to be used very often.

I accidentally didn't branch from Master. Corrected via a forced push to a new branch.
Removed the short option and renamed the long option as suggested.

src/ui/wxWidgets/SafeCombinationSetupDlg.cpp
rafaelx: Changed as suggested.

src/ui/wxWidgets/YubiMixin.cpp
rafaelx: Changed as suggested.

src/ui/wxWidgets/YubiMixin.h
rafaelx: Changed as suggested.

src/ui/wxWidgets/DbSelectionPanel.cpp
```diff
@@ -124,7 +126,10 @@ DbSelectionPanel::DbSelectionPanel(wxWindow* parent,
 
 DbSelectionPanel::~DbSelectionPanel()
 {
-  delete m_pollingTimer;
+  if (m_pollingTimer != nullptr) {
```
ronys, line 129: The if statement isn't needed, since the language guarantees that delete-ing a nullptr is safe.
Also, I don't think that setting m_pollingTimer to nullptr is needed, since when the destructor function is done, nothing should be able to access this member.
(This comment applies to all destructors)

PR #1036, opened 2023-09-28
Fix minor CStateBitmapManager resource leak.

AshleyT3:
GDI objects were being deleted, but CBitmap instances remained. Given current usage, leak is minor at ~96 bytes each time a password dialog is dismissed.


AshleyT3:
@ronys no urgency but fyi when you have a sec.

ronys:
Thanks!

PR #1043, opened 2023-10-20
Fix: password length not updated from the policy in the generator (wx)

dlex:
In the password generator of wx GUI, when a password policy is selected, password length did not get updated, resulting effective settings different from the ones of the selected policy.

Reproduce: 
* Have more than 1 policy configured in the open database
* Manage -> Generate Password
* "Use Database Policy" on
* Change the policy from "Default Policy" to a different one with "Password length" configured to a value different from the one in "Default Policy"

Observed:
* "Password length" does not change
* When a password is generated, it uses the length from the "Default Policy"

Expected:
* "Password length" is update to the value in the selected policy
* Length of generated passwords is as the selected policy says


ronys:
Thanks!

dlex:
> The minimum length information and its display in the corresponding controls is also affected. (See also PasswordPolicyDlg::UpdatePolicy() about the minimum length data.)
> 
> Strange that this was not taken into account.

Sorry I was not fast enough to add that here. A new PR created for that: #1044

PR #1044, opened 2023-10-20
Fix: minimum char class lengths not updated from the policy in the generator (wx)

dlex:
This is a follow-up to #1043, that also relates to minimum numbers of characters requires for different character classes. All descriptions of #1043 are applicable here by extension.


rafaelx (approved):
Great!

ronys:
Thanks!

Issue #1009, opened 2023-08-10
Mac: Unlock Safe from systray menu opens an empty window


nobugshere:
On macOS with wxWidgets 3.2.2.1
When I lock a safe then try to unlock it via the System Tray menu item "Unlock Safe", the window frame is opened but no content is displayed, not even the tool bars.  If I then minimize and restore the window, the content is refreshed as normal.  This is also mentioned in #933 and #1001.

On Windows and Linux, the same menu option only unlocks the safe, but does not restore the UI.  The discussion in #1001 indicates that this is the intended behavior on all platforms.  I have not been able to track down the root cause of the window frame displaying.  At the moment, I'm thinking it may be down in the wxWidgets library.

I intend to keep looking at it, but if nothing else, I have a patch I can offer as a workaround that would restore the entire UI instead of a non-functional window.

ronys:
Indeed "unlock" is supposed to keep the UI hidden. "Restore" is supposed to prompt for the master password and restore the UI.
The patch you're proposing would make the two equivalent, which isn't good UX.

nobugshere:
I give up.  I can't find a way to suppress the empty window frame, or even why it's appearing.  PR #1047 makes "unlock safe" the same as "restore".  That may not be good UI design, but it's better than a broken window :-)

ronys:
How about minimizing the window immediately after restoring it?

seamusdemora:
I've noticed this behavior, but didn't comment... for me, I just do not **'Lock Safe'**. I tried it just now, and I agree that the current behavior is BFU. I also agree that it should be fixed, and the proposed solution (unlock safe = restore) sounds as good as any. 

Thanks again for keeping pwsafe for Mac "on track"!

PR #1047, opened 2023-10-25
macOS: Restore window when safe is unlocked

nobugshere:
This is in reference to Issue #1009 
On macOS, the systray unlock safe menu option opens an empty window frame.  I couldn't find a way to suppress the empty window frame, so the workaround is to go ahead and fully restore the UI.


ronys:
This make "unlock" equivalent to "restore".
Is it possible to minimize the window after restoring it?

nobugshere:
Yes, I did try that, but I didn't think it looked very good to have a window appear and immediately disappear.

Issue #1010, opened 2023-08-10
wx: Open/cancel an entry with multiple history entries gets a spurious change warning


nobugshere:
On macOS and Linux, when I open an entry that has multiple old passwords in the history list, then immediately click cancel, I get the warning pop-up that something has changed.  This appears to be the result of a difference in the way AddEditPropSheetDlg::PreparePasswordHistory() sorts the list from the edit dialog vs. how it's stored in the actual DB.  One ill fated attempt to correct this was in #1001.  A better way would be to re-work AddEditPropSheetDlg::PreparePasswordHistory() to match the way the DB records the list.

nobugshere:
Proposed fix: PR #1048
Factored-out some duplicated code from three different places.  The new routine should result in the same ordering as before, but now the sorting and trimming will be more consistent.

nobugshere:
After some more work and thought, I submitted PR #1055 as an alternative to PR #1048.  Some of the problems I encountered:
1. False mis-compare when canceling the edit dialog due to AddEditPropSheetDlg::PreparePasswordHistory sorting newest first, whereas the DB entry is nominally oldest first.  (This is the original problem.)  The easy fix would have been to just sort both lists before comparing, but wait, there's more!
2. If the max limit is ever reduced, the list would be sorted newest first, trimmed from the end, and stored newest first.
3. Next time the PW is changed, CItemData::UpdatePasswordHistory would use push_back to add to the end of the list then trim from the *front* without re-sorting, thus removing the wrong entry if the list had already been stored newest first upon reaching the limit.
4. Need to handle existing DB entries that may already be in an undefined order.

I like the new approach better because it encapsulates and standardizes the sorting and formatting of the string that is stored in the safe.  It also eliminates the duplication of the code that formats the string for storage.

nobugshere:
Closed after PR #1055 merge.

seamusdemora:
Thanks for all of your good work here. Any projections as to when this code might be released/available for d/l?

PR #1048, opened 2023-10-25
wx: Password history check always indicates changes

nobugshere:
Fix for issue GH #1010.  The PW history check failed if there is more than one entry due to inconsistent sorting.
This was tested on macOS 14 and Fedora 38 ARM.  Since some core routines were changed, it still needs testing on Windows.


nobugshere:
Added a line to the V3 and V4 format documents to specify the history sort order.

src/core/PWHistory.cpp
```diff
@@ -174,3 +174,40 @@ StringX MakePWHistoryHeader(bool status, size_t pwh_max, size_t pwh_num)
      << setw(2) << pwh_max << setw(2) << pwh_num;
   return os.str().c_str();
 }
+
+// This code was factored from CItemData::UpdatePasswordHistory and AddEditPropSheetDlg::PreparePasswordHistory
+// and is called from both places.
+// Sort the list, trim it to the maximum length (if necessarry), and format as a StringX.
```
ronys, line 180: s/necessarry/necessary/

src/core/PWHistory.h
```diff
@@ -101,6 +101,8 @@ StringX GetPreviousPassword(const StringX &pwh_str);
 
 StringX MakePWHistoryHeader(bool status, size_t pwh_max, size_t pwh_num);
 
+StringX PWHistoryToStringX(PWHistList &pwhistlist, bool keep_history, size_t pwh_max);
```
ronys, line 104: define as a cast operator?

src/core/PWHistory.cpp
nobugshere: Oops.  Thanks.

src/core/PWHistory.h
nobugshere: Any more feedback on this?  I realized (a bit late) that since I changed core routines, the Windows MFC code may also need an update.  Then again, Windows didn't have the bug in the first place.  I don't have a Windows build environment set-up at the moment to check it.  (I'm hoping someone else can cover that part. ;-)  In any case, I'm working on making PWHistList a full fledged class, the wx side is almost done and looks a bit cleaner, but it will certainly need Windows changes.  If you are interested in that, I think I can submit it as a new PR that would include this one, or as a separate follow-on after this is merged.

ronys: Since this PR already touches core, and therefore affects the Windows build, I'd like to do the merge once, so please add the changes here.

nobugshere: If you want a look at the version with PWHistList as a class, you can see it [here](https://github.com/nobugshere/pwsafe/tree/pwHistClass).  The wx side is done except for coretest.  If you decide you want it, I'll either post a new PR or merge it with this one.  If you don't want to deal with the class changes, I'll add a unit test to this PR.

ronys:
Superceded

PR #1050, opened 2023-10-31
[i18n] Minor corrections to german translation in pwsafe_de.po

rafaelx:
Based on the correction suggestions in 780f98c.


ronys:
Danke!

PR #1055, opened 2023-11-16
wx and Windows: Re-work of PWHistList into a class

nobugshere:
Submitted for your review: This PR is a fix for issue #1010 (and some related sorting problems) and a complete alternative to PR #1048.  It converts PWHistList from a typedef into a full class.  The idea is to consolidate and encapsulate the code related to sorting and formatting of the history list field.  Other code is free to sort and modify the list, however, a specific order and format will always be imposed on the stored string (oldest to newest.)  It should correctly accommodate existing safe entries that are already "out of order"; they will only be re-sorted when they require an update, but with more consistent handling they should not cause false change detections. 

Because this changes core routines, it affects all builds.  It has been built and received at least some testing on the following:
macOS 14 Sonoma M1Max ARM, Xcode 15 (had the most testing and daily use)
Fedora 38 ARM
Windows 10 x86_64, VS2022


ronys:
Merged manually into master. Thanks!

PR #1061, opened 2023-12-02
Maintain modal dialog state across lock/unlock.

AshleyT3:


ronys:
Can you describe a scenario that can demonstrate the effect of this PR? I understand what it does conceptually, but want to see the actual effect on the UI.

AshleyT3:
> Can you describe a scenario that can demonstrate the effect of this PR? I understand what it does conceptually, but want to see the actual effect on the UI.

@ronys The reasons are not theoretical. Here are steps to observe the issue...

1. Install pwsafe release 3.64 or your favorite official Windows release.
2. Run pwsafe.
3. Create a new DB named p1.psafe3.
4. Choose menu Manage/Options/Security, check "After" and set to 1 minute for brief duration to lock timeout.
5. Add a single entry with 'a' as the group, title, username, password, save the entry.
6. Right-click on the entry to see the context menu, choose "Display Subset of Password." Do not close the dialog, let it remain open.
7. Wait for DB timeout.
8. Unlock the DB.
9. Observe the Display Subset of Password modal dialog is now modeless because pwsafe modified window state breaking the dialog's configuration essentially owned/expected by MFC at the time of lock.
10. To accentuate the issue, do not dismiss the new modeless dialog created in step 9 and follow steps 6 through 9 again in order to convert any other pwsafe modal dialog into a modeless dialog, including any dialog you've already converted (i.e., open multiple View/Edit, Manage/Options, etc.). 

Some screenshots from install of release V3.64...

Following is example of many modal dialogs converted to modeless...

image

Following is example of deleting an entry while a couple of View/Edit property sheets are open at once...

image

Following is example of all entries deleted while a couple of VIew/Edit property sheets are open...

image

ronys (approved):
Good catch, nice fix. Thanks!

PR #1072, opened 2023-12-24
macOS: Update Xcode build to C++17

nobugshere:
Hi @ronys,
It looks like you tried to update the macOS build to C++17, but ran into a problem.  The only problem was a compile error trying to cast to a protected base class.  Changing it to public compiled successfully.  It's actually casting to a public base class of the protected base class (wxEvtHandler), so I guess C++14 wasn't as strict about that.


ronys:
Nice! Looking at the error messages at time, I thought the problem was in wxWidgets, which is why I chose to keep c++14 here.
Odd that the problem only popped up in the MacOS build.
In any case, since there's no user-visible change, I don't think this justifies a minor version bump. Can you revert the change to version.wx?

nobugshere:
Done.  After submitting the PR, I tried it on Fedora 39.  The CMakeLists.txt file is set-up to use 14 for any non-windows build.  If I bump that to 17 and rerun Cmake I get the same error.  But, I don't think I would have figured it out as quickly from the GCC error messages as I did from the Clang/Xcode presentation.  I'm not a Cmake guru, but that file has code for building wxWidgets and comments that 3.2.1 doesn't compile with C++17, so I didn't change it.

ronys:
Thanks!

PR #1099, opened 2024-02-11
wx: Add wxWidgets version number to the About dlg

nobugshere:
This adds the wxWidgets version number to the About dialog:
About with version number

One side effect on macOS is the string "(wxWidgets)" is removed from the app name in the main menu (see before and after screenshots) , but I think it looks better this way :-)

Menu before

Menu after

The file version.in has a separate "#ifdef linux" branch, but it doesn't seem to be used i.e. this change also works on Linux.  If it's obsolete I can remove it.


ronys:
I like the idea, but the first line now has too much information.

Can it be changed so that the first line is just "PasswordSafe vX.Y.Z (git info) [debug] macOS
and the wx info moved to a new line, immediately below, beginning with "Built using ", e.g., "Built using wxWidgets 3.2.4"?

nobugshere:
Ha! That's how I did it at first, then I changed it.  I should have stuck with my gut :-)

Any thoughts on the unused "#ifdef linux" in version.in?

Screenshot 2024-02-11 at 1 57 57 PM

ronys:
Thanks! I've made a small tweak (removed the colon after "using" and implemented your cleanup suggestion re "ifdef linux"

Issue #1085, opened 2024-01-27
Safe Combination vs. Master Password wording between Linux and Windows versions


igor2x:
In Windows 10 using Password Safe v3.65, menu Manage  | Master Password
![image](https://github.com/pwsafe/pwsafe/assets/10895030/7adc7440-cf81-40fc-9d13-f407a3f27b31)

On Ubuntu 23.10 using Password Safe v1.18.3, menu Manage | Change Safe Combination.
![image](https://github.com/pwsafe/pwsafe/assets/10895030/b98a562c-3f03-417b-b903-7c22641be631)

Two different wordings. If we follow the "inclusive language" both wordings show be the "Safe Combination".

P.S. By the way in my humble opinion checkbox name "Show Combination" should be changed to "Show Safe Combination", to make it clear what this checkbox does or even better "Visible".

P.S.S Also at first I was confused the logic of "Confirmation". Maybe whole wording could be better. Like:
![image](https://github.com/pwsafe/pwsafe/assets/10895030/a31f4c9a-87ad-46d0-a12b-9e6f644dc3c9)

igor2x:
Similar issues:
- First window after program starts (left side)
- Toolbar "Lock" and then unlock (right side)

On top Windows, on bottom Linux.

![image](https://github.com/pwsafe/pwsafe/assets/10895030/ffca57f0-2090-446a-8a04-a0563d41d8eb)

igor2x:
Similar "master password" vs. "safe combination" issues:

Left side images: First window after program starts.
Right site images: Toolbar "Lock" and then unlock.
On top Windows, on bottom Linux.

![image](https://github.com/pwsafe/pwsafe/assets/10895030/5ad1df50-d717-416c-b2e2-c223bd2af894)

Current wording:
"Please enter the Master Password for this password database."

Suggestion:
"Enter the Master Password for unlock password database:"

igor2x:
Similar issues. Print screen:
- First window after program starts (left side)
- Toolbar "Lock" and then unlock (right side)
- On top Windows, on bottom Linux.

Looking at above print-screens:
1. Color icon of Password Safe in Linux is black. It should be dark red (like on Windows).
2. Wording "Open Password Database" may be strange working. Why "open"? What does it mean? "open as noun" or "open as verb". It can also be wrongly understood as "been open". To simplify I suggest removing word "Open" from the string.
3. Current wording: "Please enter the Master Password for this password database."
[...]

PR #1111, opened 2024-02-20
Change master password dialog

igor2x:
Intention is to fix a bug https://github.com/pwsafe/pwsafe/issues/1085

Unify wording between Windows and non-Windows for Change Master Password dialog. I have also removed the word "Please" from the string on Windows and non-Windows version, because it is redundant. Simplification.

In the process of changing strings for Change Master Password dialog, I have also changed the same named strings on other dialogs (first dialog with master password entry and File | Export) , but I did separate commits for each one.


ronys:
Nice work. One minor request:

In the "Change Master Password" dialog, the capitalization of "Master Password" is inconsistent: "Show Master Password" vs. "Old master password" and "New master password". All of these instances should be capitalized, as in the Windows version. The text in the explanatory sentence should remain uncapitalized, though.
wxWidgets:
image
Windows:
image

Thanks,

Rony

PR #1114, opened 2024-02-21
If incorrect master password entered, replace display more friendly message

igor2x:
If Master Password is incorrectly typed in, bellow dialog is displayed, which message is little bit confusing, What is "passkey"?, and options are too crowded.
image

This pull request changes dialog to:
image

### DIFF
--- a/src/ui/Windows/res/PasswordSafe3.rc2
+++ b/src/ui/Windows/res/PasswordSafe3.rc2
@@ -237,7 +237,7 @@ BEGIN
   IDS_SYNCHRONIZED        "%d entries synchronized"
   IDS_NOCURRENTSAFE       "[No current database]"
   IDS_CANNOTBEBLANK       "The master password cannot be blank."
-  IDS_INCORRECTKEY        "Most commont issues:\n - incorrect master password,\n - not a Password Safe database,\n - a corrupt database, or \n - backup database has been opened instead of original, ending with '~' character."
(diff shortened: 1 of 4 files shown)


igor2x:
Above additional commits and now dialog looks like:
image

What I still don't like, but don't know how to fix:
- title bar, to be just "error"
- left "!" icon to be aligned on top, it should be on the middle
- last word broke to new line

nobugshere:
I think the last bullet item, about the backup database, is not meant to be listed as a possible cause of the problem.  It's a hint to the user that a backup DB might be available if the original is corrupted.  However, all of my backups end with .ibak.  Is ~ still in use? (wx version)

igor2x:
@nobugshere, you are correct, there is no "~" file anymore. It is now .ibak file.

I have checked dialog on Password v3.65 on Windows:
image

On Windows it doesn't look as bad as on non-Windows.

Maybe I should:
- revert back changes on Windows like there were before
- make non-Windows message the same as on Windows

igor2x:
In above commit:
- revert back changes on Windows like there were before (except adding space between "PasswordSafe" word)
- made changes on non-Windows the same like on Windows

If there are only three options, it is nicer to have a sentence. If there was four or more, it would be better each option in its own row.

igor2x:
It now looks like:
image

igor2x:
In above commit, I have changed title bar text from "Error" to "Can't open a password database" for:
- incorrect master password dialog
- 3-times incorrect master password
- lock database and then try to unlock database with  wrong password

Dialogs look now like this:

image

@ronys, I think now it is ready for review.

PR #1116, opened 2024-02-22
Remove word Safe from Lock Safe and Unlock Safe

igor2x:
Remove word "Safe" from "Lock Safe" and "Unlock Safe" strings.

Before this PR for example:
- on Linux File | Lock **Safe**
- on Windows File | Safe

Make wording the same on Windows and non-Windows builds


igor2x:
@ronys, now I see build for mac-pwsafe failed, but error message is: "Install language files and create dmg"
Did my changes really made this error or is this error from some other source?

nobugshere:
I don't know what happened here.  It looks like the build went okay, then create-dmg created the .dmg disk image, but was not able to unmount it because it was busy for some reason.  All I can suggest is to try it again.  If you go to the Actions tab, in your own repo, and select the mac-pwsafe workflow you should see a button to trigger the workflow on a selected branch.

igor2x:
@nobugshere, on Action tab on my repo it was message that Workflow was disabled in my repo I clicked on enable and now I have empty workflow.
image
This is going to be useful for future build.

igor2x:
I just pushed another simple change and everything has built successfully.
Now PR is ready for review.

ronys:
Re the mac-pwsafe build: I've seen this job fail occasionally on the unmount operation. Looks like a race condition in the build environment. Retrying works for me.

Issue #1020, opened 2023-09-04
Yes/No buttons: Use Delete/Cancel buttons instead of Yes/No before executing dangerous actions like Delete Group


igor2x:
Today I have have pretty big disaster with Password Safe v3.63 on Windows 10, I have permanently lost 30+ passwords, because of two circumstances:
- my not the best setting in pwsafe,
- me as not reading Yes/No question with potential dangerous lost of data

What happened?
1. First my settings: menu Manage | Options | Backups
   - "Save database immediately after any change": checked
   - "Backup Name | Max=3"
2. I have changed some passwords, but because every change results in database save (first setting) and only having 3 backup history files I quickly run out of useful backup, the trap I have gotten into.
3. Then I notice one of my groups does not have the best name and I decided to rename the group. Right mouse click "Rename Group".
4. Question popped up and I clicked on Yes button. Yes, yes, yes, don't bother me I want to rename. But.... to my surprise whole group of passwords were deleted. What I most probably did is miss the "Rename Group" and click one selection bellow "Delete Group", but because I was convinced I was renaming I confirmed Yes without reading the message, but the message was most probably delete confirmation.
5. This should not be a problem. I closed down pwsafe and lets rename backup... But all backups are few seconds old. Upsss, I lost my passwords.

I don't blame the software, I have had dangerous settings and did not read the notification.

Now I did:
- unchecked "Save database immediately after any change" setting
- changed "Backup Name | Max=50"

But my recommendation for feature request: Please never ever use Yes/No questions when something potentially very dangerous fallows like deleting data. In my humble opinion Yes/No is pretty bad practice, because it forces user to actually read and understands the warning and then anxiously click and hope correct Yes or No is selected. Yes/No are dangerous because users are so annoyed with this kind of confirmations that they always click on Yes, because yet one more nag.

My recommendation is, when something dangerous follows name buttons so it is clear from the buttons itself (without reading actual warning) that something dangerous is potentially following. Instead of Yes and No buttons please use "Delete" and "Cancel". When I see Delete/Cancel combination I immediately know clicking on Cancel is something that gets me out of danger. And... if "Delete" buttons appears, it is immediately obvious that what follows is not "rename group" and in this case I would for sure read the notification massage.

---
I don't know if there are any other Yes/No questions, if they are, can you please consider changing to ACTION/Cancel variant.

ronys:
[...]

PR #1119, opened 2024-02-22
Rename Yes_No buttons to Delete_Cancel in Delete Group and Delete Entry on non-Windows

igor2x:
Attempt to fix [issue 1020](https://github.com/pwsafe/pwsafe/issues/1020)

---
1. **DONE:** This commit is for non-Windows only.
image
---
2. **TODO:** Now searching for Windows strings and I get:
> src/ui/Windows/res/PasswordSafe3.rc2:  IDS_DELGRP              "Are you sure you want to delete the selected group?"
> src/ui/Windows/res/PasswordSafe3.rc2:  IDS_DELENT              "Are you sure you want to delete the selected entry?"

@ronys, can you confirm I should change following two strings in the code? I don't have Windows to build Password Safe Windows build.
image

---
3. **TODO:** @rony, it would be nice this questions have a "exclamation point" icon on the left side. Any tip how to add it?
image


ronys:
For now we'll just change the texts per your suggestions. To add an exclamation mark icon, I'd like to refactor this to use the standard message box, which should be in a separate commit/PR.

PR #1129, opened 2024-02-23
Dialog with wrong master password. Remove dialog icon and make dialog less wide

igor2x:
I made some additional improvements to "incorrect master password" dialog on top of [PR 1114](https://github.com/pwsafe/pwsafe/pull/1114) that was already merged into master.

Now I see that design changes in PR 1114 was not the best choice, after using program in day-to-day activity. Now I see how distracting "explanation mark" icon is and how wide is actually dialog and so trying for eyes to move so long from end of first line to second line.

- Text is the same.
- Removed icon
- Added line brake to make dialog shorter

Comparisons:

image

### DIFF
--- a/src/ui/Windows/res/PasswordSafe3.rc2
+++ b/src/ui/Windows/res/PasswordSafe3.rc2
@@ -237,7 +237,7 @@ BEGIN
   IDS_SYNCHRONIZED        "%d entries synchronized"
   IDS_NOCURRENTSAFE       "[No current database]"
   IDS_CANNOTBEBLANK       "The master password cannot be blank."
-  IDS_INCORRECTKEY        "Incorrect master password, not a Password Safe database, or a corrupt database."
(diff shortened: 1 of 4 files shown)


igor2x:
I have discussed with @ronys and for now he wants to preserve icons left of dialogs, so I have put them back as there were original. So this PR is now only by adding new line to prevent super wide dialog. Tiny improvement, but I still think it is better then nothing.

By the way, I was struggling to understand how did I change dialog icon if there was no icon before. Now I have realized that I have been testing "before" change with Password Safe v1.18.2 **flatpak version**. I have read some xwWidgets documentation and it is written that platform on which program runs, can decide by its own, if icons are displayed or not. Flatpak does not display dialog icons, so that is why in first picture in this PR it is clear that flatpak version does not have icon, then I have only changed text and magically icon appears, but second image is texting Linux build (not flatpak one), where icons are displayed by default.

@ronys, can you review. The idea is not to have super long dialogs, because they look ugly, so braking lines at predictable words is way better then letting wxWidgets to do it when some width is exceeded.

image

PR #1130, opened 2024-02-23
Master password setup dialog

igor2x:
- Change title
- Change text "Verify" with "Confirmation" to have the same terminology with "Change Master Password" (second dialog on print-screen)
image


igor2x:
**Note:** Can you please make sure I changed the correct string for Windows build  in  "Change text Verify with Confirmation" commit above. For such a short string I am not 100% I changed correct string.

---

Now after above commit dialog looks like:
image

TODO: Add missing semicolon at the end "Confirmation" word.

ronys:
The text in the Master Password Setup dialog box still refers to the "combination" instead of "master password".
Suggest re-using the Windows text:
"A new password database will be created.\nThe Master Password you enter will be used to encrypt the password database file. The Master Password can use any keyboard character and is case-sensitive."

Issue #1120, opened 2024-02-22
Yes/No buttons: Create new database 


igor2x:
One additional Yes/No. Create new database and type in simple password and bellow dialog appears:

![image](https://github.com/pwsafe/pwsafe/assets/10895030/6ed1be6e-d4c5-4757-8f71-ca0cb5018c60)

1. Yes/No suggestion:
- change **No** with **Cancel** button
- change **Yes**  with **Use anyway** button

2. Different terminologies
- when new database is selected to be created there is "Safe Combination" terminology
- in dialog above there are two terminologies: "passphrase" and "Password"
I suggest to use only one terminology consistently to simplify wording.

igor2x:
Exactly the same problem is selecting Manage | Change Safe Combination and selecting new weak password.

Dialogs are exactly the same.

rafaelx:
Resolved by #1131

PR #1131, opened 2024-02-24
Make Weak master password dialog more pretty.

igor2x:
Attempt to fix issue https://github.com/pwsafe/pwsafe/issues/1120

Fixes for two dialogs (that have exactly the same appearing):
- "weak password" dialog if at "change master password" dialog weak password is typed in
- "weak password" dialog if at "create new database" dialog weak password is typed in

Fixes:
1. Exact problem in title instead of old "Error".
2. Replace fatal error icon with explanation point (because fatal icon should be when something is over, can't be continued). By the way on picture bellow on left site is flatpak version that does not uses icons.
3. Nitpick: add full stop and the end of error message.
4. Changed button labels from Yes_No to "action" and Cancel.
5. Set Cancel button as default button (this way already done at new database dialog, but was missing at change master password). If no default is set then default is "Yes", which is what we want to tell user that this is not the best option.

Before and current PR changes:

image

### DIFF
--- a/src/core/core.rc2
+++ b/src/core/core.rc2
@@ -17,7 +17,7 @@
 
 STRINGTABLE
 BEGIN
-  IDSC_PASSWORDTOOSHORT   "Password is too short"
+  IDSC_PASSWORDTOOSHORT   "Password is too short."
   IDSC_PASSWORDPOOR       "Password should be mixed case, with at least one digit or punctuation character"
 END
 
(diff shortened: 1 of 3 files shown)


ronys:
Interesting. Just tried this on Debian, no icon is shown.
Found this, which explains the logic, along the lines that you've mentioned elsewhere.

Issue #1122, opened 2024-02-22
Yes/No buttons: Save changes dialog


igor2x:
One additional Yes/No. Do some changes and then click on "Close" button from toolbar and bellow dialog appears:

![image](https://github.com/pwsafe/pwsafe/assets/10895030/ec075fed-7f7d-48c5-a4af-cf217524ffef)

1. Yes/No suggestion:
- change **No** with **Discard**
- change **Yes** with **Save**
- **Cancel** button is fine just like it is

2. Order of buttons. Standard behavior of buttons are (on Linux): Cancel / Discard / Save.

3. Maybe add new line before database path.

Issue #1132, opened 2024-02-24
Cancel button Save changes dialog closes down Password Safe instead of cancels the action


igor2x:
Password Safe v1.18.2 installed as flatpak on Ubuntu 23.10.

1. Manage | Options | Backups tab | make sure option "Save database immediately after any change" is unchecked.
2. Right click on entry or folder and delete it and when asked confirm delete action.
3. Click on "Close" button from toolbar.
4. Dialog appears with three buttons:
- No - if selected, database is closed down and changes are discarded. Works as expected.
- Cancel - PROBLEM. I expect this option would close down the dialog and do nothing. But two weird things happen. First the same dialog reappears (first problem), click on Cancel again and whole Password Safe is closed down (second problem). Positive is when I reopen the database data are not lost.
- Yes - if selected, database is closed down and changes are saved to database. Works as expected.

igor2x:
Interesting if in step 2 click on "New" button the same dialog appears in step 4, but Cancel button works as expected, dialog is closed down without any change to existing database.

ronys:
Fixed in 50815abc5

PR #1133, opened 2024-02-24
Unsaved database close down or new database opened

igor2x:
Attempt to fix https://github.com/pwsafe/pwsafe/issues/1122

How to test:
1. Open password database and make sure Manage | Options | Backup tab | "Save database immediately after any change" is unchecked.
2. Delete some entry or group and confirm delete action. So change is made that is currently not saved.
3. Click on "New" or "Close" button from toolbar and dialog appears that we have unsaved changes (this PR).

image

Above image:
- top and bottom left Password Save 1.18.2 flatpak version
- top and bottom right changes from this PR

- first row what happen if in step 3: New button is clicked
- second row what happens if in step 3: Close button is clicked

### DIFF
--- a/src/ui/wxWidgets/MenuFileHandlers.cpp
+++ b/src/ui/wxWidgets/MenuFileHandlers.cpp
@@ -88,9 +88,11 @@ int PasswordSafeFrame::New()
   int rc, rc2;
 
   if (!m_core.IsReadOnly() && m_core.HasDBChanged()) {
-    wxString msg(_("Do you want to save changes to the password database: "));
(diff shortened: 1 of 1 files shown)


igor2x:
NOTE: There is weird bug if "Close" is used in step 3 and Cancel button is clicked. This was issue before I have made a changes and I have reported issue in [bug report 1132.](https://github.com/pwsafe/pwsafe/issues/1132), just in case if you test and wonder if changes in this PR produced the new issue, it isn't.

Issue 1132 is independent from the issue I am trying to solve in this PR.

@ronys, can you please review.

PR #1137, opened 2024-02-25
Change wording Database Policy to Password Policy

igor2x:
In Manage menu there is "Password Policy" entry. If we select that option, "Manage Password Policy" dialog opens.

Select Manage | Generate Password and bellow dialog appears with checkbox "Use Database Policy", which points to password policy named Default Policy.

This PR changes name of "Database Policy" to "Password Policy". Left side: current, right side: this PR.

image


ronys:
The current texts aren't clear, but I don't think that the proposed changes help.

Here's the idea:
In the Manage Password Policies, the user can create, edit and delete **named** policies that are stored in the database.

Wherever a password can be generated, the user can choose to use either a named policy ("default policy", or one created by the user), or an "ad-hoc" policy that the user can define in that window (generate password or Add/Edit Entry).

I think it would be best to use the wording that's in the Windows Generate Password dialog box: "Use Named Policy":
image

This should also replace the "Use Database Policy" radio button in the Add/Edit Password Policy Tab in Windows:
image

ronys (changes requested):
Suggested alternate wording, for Windows as well, in comment.

Issue #1128, opened 2024-02-23
Deleting empty group returns incorrect error like deleting entry


igor2x:
I am testing deleting entries and deleting group in Password Safe v1.18.2.

My test database looks like:
![image](https://github.com/pwsafe/pwsafe/assets/10895030/3ffe4b38-3a44-4320-b1db-68dfa9e23b26)

1. Delete group.
![image](https://github.com/pwsafe/pwsafe/assets/10895030/e8e8e974-b71c-4187-a14d-7769393c6c9e)
2. And correct message is returned. Click No.
![image](https://github.com/pwsafe/pwsafe/assets/10895030/e4e27b47-d829-48fc-9b3f-377a895538ed)
3. Now delete entry. Click Yes.
![image](https://github.com/pwsafe/pwsafe/assets/10895030/bf2c2827-6c0a-4c4d-9703-4060755546e3)
4. And correct message is returned.
![image](https://github.com/pwsafe/pwsafe/assets/10895030/fbbbc35b-8b73-4122-8994-59780995f12f)
5. Now delete empty group.
![image](https://github.com/pwsafe/pwsafe/assets/10895030/05f4b626-d7ba-4a96-8128-bb2e46fffb94)
6. And now problem appears. I expect to get "delete group" confirmation dialog, just like in step 2, but I get "delete entry" message.
![image](https://github.com/pwsafe/pwsafe/assets/10895030/a2aaf11e-81e1-48ee-ad82-c787abbda0ae)

If click on Yes, empty group is delete correctly. Just little bit of confusion, because it looks like Password Safe treats "empty group" as "entry".

The problem is, because I ask myself, did I really clicked delete on entry? Which one? But then I see I have tried deleting group.

Probably some logic behind recognizing empty group is false.

igor2x:
Looks like code for this issue is at: src/ui/wxWidgets/DeleteConfirmationDlg.cpp

![image](https://github.com/pwsafe/pwsafe/assets/10895030/9037ba04-f37b-4417-ae51-cf2ada0ce242)

nobugshere:
Taking a quick look, it looks like the actual number of children being passed into the constructor isn't being used by the dialog class, so I would suggest re-working it into an "isGroup" flag.  (Note that it can only be non-zero when using tree-view.)

nobugshere:
Submitted PR #1141 to fix the test for the group/entry message.

PR #1141, opened 2024-02-26
Display the correct "Are you sure" message when deleting an empty group

nobugshere:
This is for issue #1128 .  It converts the unused num_chidren into an isGroup bool.


ronys:
When this was originally written, empty groups weren't a thing. Thanks!

PR #1164, opened 2024-03-06
Rename Safe List to Password Database in File | Clear Recent Safe List

igor2x:
Word "Safe" is confusing and we are replacing it to "Password Database".
- Current: File | Clear Recent **Safe** List
- This PR: File | Clear Recent **Password Database** List

image


ronys (changes requested):
1. There's a typo in the Windows text: "Databse" instead of "Database"
2. The text is a bit long in English, could be even worse in other languages. Suggest changing this to "Clear Recently Opened List":

- Original text was 22 characters
- "Clear Recent Password Database List" is 35 characters
- "Clear Recently Opened List" is 26 characters

PR #1169, opened 2024-03-07
Simplify File Properties title dialogs

igor2x:
File | Properties. Simplify title.

image

### DIFF
--- a/src/ui/Windows/res/PasswordSafe3.rc2
+++ b/src/ui/Windows/res/PasswordSafe3.rc2
@@ -926,7 +926,7 @@ END
 STRINGTABLE
 BEGIN
   IDS_NUMGROUPS_E              "%s (%s empty)"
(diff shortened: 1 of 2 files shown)


src/ui/Windows/res/PasswordSafe3.rc2
```diff
@@ -926,8 +926,8 @@ END
 STRINGTABLE
 BEGIN
   IDS_NUMGROUPS_E              "%s (%s empty)"
-  IDS_ENTER_NEW_DB_NAME        "Please enter the new database name"
-  IDS_ENTER_NEW_DB_DESCRIPTION "Please enter the new database description"
+  IDS_ENTER_NEW_DB_NAME        "Short name""
```
ronys, line 929: 1. Extra double quote broke the build
2. "Short" is irrelevant here. For consistency, the title should be the same as the field: "Name"

ronys (changes requested):

PR #1182, opened 2024-03-09
[wx] Fix labeling of Show/Hide button in Add/Edit dialog for Alias

### LINKED ISSUE #1178: For an alias, the Show button is not renamed to Hide in the Add/Edit dialog box
When the Add/Edit dialog box for an alias appears and the Show button is clicked to display the password, the button is not renamed to Hide.

rafaelx:
Resolves #1178.


src/ui/wxWidgets/AddEditPropSheetDlg.cpp
```diff
@@ -1453,6 +1454,7 @@ void AddEditPropSheetDlg::ItemFieldsToPropSheet()
     } else {
       HidePassword();
     }
+    UpdateShowHideButton();
```
dskrvk, line 1457: This line is in both if/else branches - maybe move it out?

dskrvk (approved):

PR #1183, opened 2024-03-10
[wx] Update text control style flags in wxGtk build without replacing text control

### LINKED ISSUE #1148: Show button to display password shrinks Password field

1. Right click and Add Entry.
2. Click on Show button and Password field for some reason shrinks (problem). I expect the size of Password field to stay the same before and after click on Show button.

image

rafaelx:
Resolves #1148.

### DIFF
--- a/src/ui/wxWidgets/AddEditPropSheetDlg.cpp
+++ b/src/ui/wxWidgets/AddEditPropSheetDlg.cpp
@@ -1703,6 +1703,13 @@ void AddEditPropSheetDlg::DoAliasButtonClick()
 
 void AddEditPropSheetDlg::UpdatePasswordTextCtrl(wxTextCtrl* &textCtrl, const wxString value, wxTextCtrl* before, const int id, const int style)
 {
+  ASSERT(textCtrl);
+#if defined(__WXGTK__)
+  // Since this function is called with only a single style flag such as "0", "wxTE_PASSWORD" or "wxTE_READONLY",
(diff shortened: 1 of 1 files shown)


nobugshere:
Does that really need to be ifdef'ed?  If "Silvia's suggestion" is still known to be needed on some platforms, I'd suggest adding it to the comments.  Or, if WXGTK is the only variant you can test, that's worth a comment too :-)

ronys:
According to https://docs.wxwidgets.org/3.0/classwx_text_ctrl.html:
"wxTE_READONLY, wxTE_PASSWORD and wrapping styles can be dynamically changed under wxGTK but not wxMSW."

I'd like to keep the ability to compile wx under Windows (broken now, on my todo list...).

nobugshere:
Just a note: The new wxGTK path does _**not**_ work on wxOSX either.  The older code still works fine.
(Tested on macOS 14.4, wx 3.2.4)

rafaelx:
Thanks for the feedback. I don't have a macOS based system to run tests on.
Is wxOSX also based on GTK or on Cocoa? In order to activate the old code for macOS as well, the defines ([Preprocessor Symbols](https://forum.wxwidgets.org/latest/page_cppconst.html)) used might have to be adjusted. Would make sense after the merge of PR #1185.

nobugshere:
It's based on Cocoa.  I think the __WXGTK__ define is correct, any other platform will use the original path.

PR #1185, opened 2024-03-12
[wx] Merge 'AddEditPropSheetDlg::UpdatePasswordTextCtrl' with 'ShowHideText' from wxUtilities to prevent redundant code

rafaelx:
- Directly/Dynamically changing the style flags of a password text control also improves the layout behavior of YubiCfgDlg.
- The ShowHideText code in wxUtilities was pretty much identical to that of AddEditPropSheetDlg::UpdatePasswordTextCtrl.

### DIFF
--- a/src/ui/wxWidgets/wxUtilities.h
+++ b/src/ui/wxWidgets/wxUtilities.h
@@ -176,9 +176,24 @@ inline const wxChar* ToStr(bool b) {
   return b? wxT("True"): wxT("False");
 }
 
-// Workaround for wxTE_PASSWORD being immutable:
-void ShowHideText(wxTextCtrl *&txtCtrl, const wxString &text,
-                  wxSizer *sizer, bool show);
+/**
+ * @brief The purpose of this function is to update or remove the style flags "wxTE_PASSWORD" or "wxTE_READONLY" of a wxTextCtrl.
+ *
+ * According to the documentation of wxTextCtrl the style flags "wxTE_PASSWORD" and "wxTE_READONLY" can be changed during
+ * runtime under wxGTK but not wxMSW. This circumstance is taken into account by this function.
+ *
+ * In the current version, under wxGTK the style flag is applied dynamically to the text control and on other platforms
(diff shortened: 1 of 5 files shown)


nobugshere:
I pulled from rafaelx:UpdatePasswordTextCtrl and verified the changes on a Mac, all looks good.

macOS M1max 14.4, wx 3.2.4, Xcode 15.3, with and without Rosetta.

igor2x:
@rafaelx, I really like you have taken time to write good comments inside source code.

For this:
"According to the documentation of wxTextCtrl the style flags "wxTE_PASSWORD" and "wxTE_READONLY" can be changed during runtime under wxGTK but not wxMSW."

Just a humble suggestion. Can you inside commend add URL like, where you have found this restriction. Someone may easily verify the claim, specially important over time (years) when restriction may be removed and in this case documentation probably also changes to indicate removal of restriction.

src/ui/wxWidgets/wxUtilities.h
```diff
@@ -176,9 +176,24 @@ inline const wxChar* ToStr(bool b) {
...
+ *
+ * @param sizer the sizer to which the text control belongs
+ * @param textCtrl the control whose style are to be changed
+ * @param text the text to update the text control with
+ * @param before the control element in the layout before "textCtrl" to respect the TAB order
+ * @param style the new style flag for "textCtrl" (previous flags will not be preserved)
+ *
+ * @see https://docs.wxwidgets.org/stable/classwx_text_ctrl.html
+ */
+void UpdatePasswordTextCtrl(wxSizer *sizer, wxTextCtrl* &textCtrl, const wxString text, wxTextCtrl* before, const int style);
```
dskrvk, line 196: Is it really necessary to put the comment in both the header and the source file?

rafaelx:
@nobugshere Great!
@igor2x I added a link in the notes section.
@dskrvk I removed the comment in the .cpp file because the .h file serves as an interface declaration.

Thank you everyone for the feedback.

PR #1186, opened 2024-03-13
[wx] Minor improvement on layout of YubiCfgDlg

rafaelx:
Makes the text control for “YubiKey Serial Number” stretchable and may fix what the following comment in YubiCfgDlg::CreateControls() points out.

> With wx:3.2.2.1 this text wxTextCtrl doesn't resize, it stays one character wide. The initial string of 0s forces it.


nobugshere:
Yes, that does fix my test case.  I think you can remove the comment and string of 0's.

nobugshere:
Looks good here.
MacOS 14.4 M1max, Xcode 15.3, wx 3.2.2.1 and 3.2.4

PR #1188, opened 2024-03-16
[wx] Fix re-editing issue of group name in tree view

### LINKED ISSUE #1145: When duplicate group appears text can't be typed in anymore

1. Create group with name: group
2. On the same root level create another group with the same name: group
   Error is displayed: "Double group name". Click on OK button.
3. Cursor gets back to "New Group" text as expected.
 
Now problem appears. Start typing and nothing happens. I expect typed text to appear in input field. Work-around is to click with mouse in input field and then text is accepted.

Print-screen after step 3:
image

rafaelx:
- Resolves #1145 
- Also seems to fix a crash when ESC cancels editing of a group name.
- Would be great if someone could also check the behavior on a Mac.

### DIFF
--- a/src/ui/wxWidgets/TreeCtrl.cpp
+++ b/src/ui/wxWidgets/TreeCtrl.cpp
@@ -971,7 +971,7 @@ void TreeCtrl::PreferencesChanged()
   ;
 }
 
-void EditTreeLabel(wxTreeCtrl* tree, const wxTreeItemId& id)
+void TreeCtrl::EditTreeLabel(wxTreeCtrl* tree, const wxTreeItemId& id)
 {
   if (!id) return;
   wxTextCtrl* edit = tree->EditLabel(id);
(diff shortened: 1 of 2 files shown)


nobugshere:
I'm able to reproduce the problem from issue #1145 on my Mac and this PR does fix it.  I could not reproduce the crash.
macOS 14.4, M1max, Xcode 15.3, wx 3.2.4

rafaelx:
For reproducing the crash.

- Add one group
- Add a second group
- Start renaming the second group by giving the name of the first group
- Close the message box that appears indicating the naming conflict
- Press ESC key -> pwsafe is gone

nobugshere:
Still no crash.  There is an option in the Miscellaneous tab: "Escape key closes application" that causes the app to close regardless of the above procedure, similar to clicking the corner button.  I've tried it both ways and with or without the system tray menu enabled, on both the Mac and a Fedora VM.

rafaelx:
I suspect I was wrong about this. The crash I remembered was probably caused by my experimental code while analyzing the problem. You are right, pressing ESC will close the application normally. Sorry for causing any inconvenience. :-|

PR #1193, opened 2024-03-21
[wx] Do not exit after entering the wrong password several times

### LINKED ISSUE #1115: Three times incorrect password attemts, differences between Windows and non-Windows
In first dialog after Pasword Safe is started.

Type in incorrect password for 3 times and on Windows:
image

The same on Linux:
image

Also on Windows OK button just closes current dialog and next try of master password can be entered.
On Linux it closes down Password Safe program.

I like Windows behavior more:
- error message is more descriptive
- why should we close down program - master password should be complicated, user probably typed in wrong password 3-times in a role (we should fight brute force with some other mechanism like 30 seconds password can't be entered or similar)

My suggestion is, make on Linux the same as on Windows:
- make the same wording
- don't close down program after 3-times incorrect master password

rafaelx:
Resolve #1115

### DIFF
--- a/src/ui/wxWidgets/SafeCombinationEntryDlg.cpp
+++ b/src/ui/wxWidgets/SafeCombinationEntryDlg.cpp
@@ -389,6 +389,7 @@ void SafeCombinationEntryDlg::OnOk( wxCommandEvent& )
 
 void SafeCombinationEntryDlg::ProcessPhrase()
 {
+  static unsigned tries = 0;
   int status = m_core.CheckPasskey(tostringx(m_filename), m_password);
   wxString errmess;
   switch (status) {
@@ -439,10 +440,16 @@ void SafeCombinationEntryDlg::ProcessPhrase()
     break;
   case PWScore::WRONG_PASSWORD:
   default:
-    if (m_tries++ >= 2) {
-      errmess = _("Too many retries - exiting");
+    if (++tries > 2) {
+      errmess = wxString::Format(_("The master password has been entered %d times without success:\n"), tries);
+      errmess += _("- Is Caps Lock off?\n");
+      errmess += _("- Is the language correct (if multilingual)?\n");
+      errmess += _("- Is this the correct database?\n");
+      errmess += _("- Perhaps the database was damaged. Try opening a backup copy.");
     } else {
-      errmess = _("Incorrect master password, not a Password Safe database,\nor a corrupt database.");
+      errmess =  _("Incorrect master password,\n");
+      errmess += _("not a Password Safe database,\n");
+      errmess += _("or a corrupt database.");
     }
     break;
   } // switch (status)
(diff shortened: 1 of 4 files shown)


dskrvk:
What is the motivation for this change? Are there security implications? E.g. it would become somewhat easier to brute-force the combination.

igor2x:
@rafaelx thank you for creating fix.

First I have compiled Password Safe according to [Linux developers instructions](https://github.com/pwsafe/pwsafe/blob/master/README.LINUX.DEVELOPERS.md) and program compiled without an error and pwsafe starts up without an issue on my Ubuntu 23.10 computer.

Then I tried to compile your fix with:

```
mkdir pwsafe_password_three_times
cd pwsafe_password_three_times/
git clone https://github.com/rafaelx/pwsafe.git
cd pwsafe
sudo sh Misc/setup-linux-dev-env.sh
```

but the last command returns error:

> Reading package lists...
> Building dependency tree...
> Reading state information...
> E: Unable to locate package libwxgtk3.0-gtk3-dev
> E: Couldn't find any package by glob 'libwxgtk3.0-gtk3-dev'
> E: Couldn't find any package by regex 'libwxgtk3.0-gtk3-dev'

Have you maybe added some new dependency or similar?

igor2x:
I have investigated and searched for package on packages.ubuntu.com and got the following packages:
image

In Ubuntu 23.10 there is no more package `libwxgtk3.0-gtk3-dev` (red rectangle), but are packages `libwxgtk3.2-1` and `libwxgtk3.2-dev`, so I tried to install them instead:

`sudo apt install libwxgtk3.2-1 libwxgtk3.2-dev`

and packages are already installed:

> Reading package lists... Done
> Building dependency tree... Done
> Reading state information... Done
> libwxgtk3.2-1 is already the newest version (3.2.2+dfsg-4).
> libwxgtk3.2-dev is already the newest version (3.2.2+dfsg-4).
> 0 upgraded, 0 newly installed, 0 to remove and 0 not upgraded.

@rafaelx, have you maybe added some dependency to some old version of wxgtk package?

rafaelx:
Hi @igor2x . I haven't made any changes to the dependencies. To the compile problem. You don't need to clone my repository. What may have happened is that you used the outdated "master" branch of my repository to compile pwsafe. However, I follow the pwsafe's "master" branch, on which the pull requests are based.

I suggest following the steps below.

- Go to the pwsafe directory you created when you clone the pwsafe repository.
- If necessary, run the script "setup-linux-dev-env.sh" again to make sure that the development environment is set up according to the latest version of the script.
- Get the pull request.

```
git fetch upstream pull/1193/head:pwsafe_password_three_times
```

- **upstream** refers to the pwsafe repository on GitHub (the name could also be origin; check with 'git remote -v')
- **pull/1193/head** refers to the pull request id
- **pwsafe_password_three_times** is the name of the branch that will be created locally. You can use any name you like.

Now, you should be able to switch to the branch which contains the pull request.

```
git checkout pwsafe_password_three_times
```

Next, go to the build directory and start the compilation process.

```
make -j 4 && ./pwsafe
```

- Use the **-j** option to spread the build process across multiple cores and thus speed it up.

Sit back and enjoy the fruits of your labor. ;-)

igor2x:
@rafaelx, thanks a lot for taking time and explaining how I should get the PR in new branch and compile.

I tested your code and it works excellent, just like expected!!! Thanks for fixing this. This was pretty annoying in my case when I have pretty complicated master password and I can easily mistype three times.

PR #1198, opened 2024-03-27
wx: Compare dialog tables are hard to read in dark mode

nobugshere:
On Linux and Mac, the Compare dialog tables are hard to read when the system is using a dark theme.  The screenshots were taken on a Mac, but my Fedora system was similar.  Also, changed the fixed font used on Mac to wxSYS_OEM_FIXED_FONT because it's easier to read than the one that was being selected. (The Linux font is good as is.)
Test systems:
macOS 14.4.1 M1Max, wx 3.2.4, Xcode 15.3
Fedora 39 ARM VM, wx 3.2.4, xfce desktop

Before - Dark mode - note the white text on white background
before dark

After - Dark mode
After dark

After - Light mode
after light

### DIFF
--- a/src/ui/wxWidgets/ComparisonGridTable.h
+++ b/src/ui/wxWidgets/ComparisonGridTable.h
@@ -20,11 +20,12 @@
 
 // #define CurrentBackgroundColor    *wxWHITE
 #if wxVERSION_NUMBER >= 3103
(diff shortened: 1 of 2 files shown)


src/ui/wxWidgets/ComparisonGridTable.h
```diff
@@ -20,11 +20,12 @@
 
 // #define CurrentBackgroundColor    *wxWHITE
```
dskrvk, line 21: I think this line can be deleted.

PR #1211, opened 2024-03-29
Change dialog title, remove word Sorry and unify the same error with the same message

igor2x:
1. Change Error title to real problem "Duplicate selection".
2. Remove word "Sorry", because it is unusual wording and redundant - simplification.
3. Two pairs of the same problem but order of selections are reversed. Unify the error message.

### DIFF
--- a/src/ui/wxWidgets/AddEditPropSheetDlg.cpp
+++ b/src/ui/wxWidgets/AddEditPropSheetDlg.cpp
@@ -3031,7 +3031,7 @@ void AddEditPropSheetDlg::OnEasyReadCBClick(wxCommandEvent& evt)
     // Check if pronounceable is also set - forbid both
     if (m_PasswordPolicyUsePronounceableCtrl->GetValue()) {
       m_PasswordPolicyUseEasyCtrl->SetValue(false);
-      wxMessageBox(_("\"Easy-to-read\" and \"Pronounceable\" cannot be both selected."),
+      wxMessageBox(_("\"Easy-to-read\" and \"pronounceable\" cannot be both selected."),
                    _("Unsupported selection"), wxOK|wxICON_ERROR, this);
       return;
     }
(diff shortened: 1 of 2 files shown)


src/ui/wxWidgets/AddEditPropSheetDlg.cpp
```diff
@@ -3031,8 +3031,8 @@ void AddEditPropSheetDlg::OnEasyReadCBClick(wxCommandEvent& evt)
     // Check if pronounceable is also set - forbid both
     if (m_PasswordPolicyUsePronounceableCtrl->GetValue()) {
       m_PasswordPolicyUseEasyCtrl->SetValue(false);
-      wxMessageBox(_("Sorry, \"easy-to-read\" and \"pronounceable\" cannot be both selected"),
-                   _("Error"), wxOK|wxICON_ERROR, this);
+      wxMessageBox(_("\"Pronounceable\" and \"easy-to-read\" cannot be both selected."),
```
dskrvk, line 3034: Nit: the order of options has changed. I think the original intent was for both dialogs to say "<This option> and <other option> cannot be both selected".

src/ui/wxWidgets/AddEditPropSheetDlg.cpp
```diff
@@ -3058,8 +3058,8 @@ void AddEditPropSheetDlg::OnPronouceableCBClick( wxCommandEvent& evt)
     // Check if ezread is also set - forbid both
     if (m_PasswordPolicyUseEasyCtrl->GetValue()) {
       m_PasswordPolicyUsePronounceableCtrl->SetValue(false);
-      wxMessageBox(_("Sorry, \"pronounceable\" and \"easy-to-read\" cannot be both selected"),
-                   _("Error"), wxOK|wxICON_ERROR, this);
+      wxMessageBox(_("\"Pronounceable\" and \"easy-to-read\" cannot be both selected."),
```
dskrvk, line 3061: Same here

ronys:
It's not a "duplicate" selection in any sense, it's really an unsupported combination, but this would be confusing.
I think "Unsupported selection" is better.

igor2x:
Now looking at "Files changed" Github feature I noticed:
-  That order of strings in both files were not the same as I assumed when reverted the strings in "Revert words order" commit.
- There was missing hypen in "easy to-read", so adding first hypen "easy-to-read".

dskrvk (approved):

Issue #1197, opened 2024-03-24
[wx] Modernization of Password Safe Dialogs


rafaelx:
## Introduction

There have been some suggestions recently to improve the visual appearance of the application. In most cases, initiated by @igor2x.

This issue should serve as a discussion about a possible modernization of the initial password safe dialog and the password input fields of the application, but can also be extended to other dialogs.

## Virtual Keyboard

In some dialogs, next to the password input field, there is a button to start the virtual keyboard.
I wonder if this keyboard is really needed and offers some added value to the user.

The reason for this consideration is that the button to start the virtual keyboard could provide a different functionality. The button could be given an eye symbol instead of the keyboard symbol and be used to show and hide (by replacing the text by asterisks) the password in the input field. It would thus replace the existing checkbox for this purpose and thus a control element would be omitted.

Ideally, the password input field would have already integrated such a button on the far right side, but wxWidgets unfortunately does not offer such a control element.

## Text Input Fields

In web / mobile applications, text input fields sometimes do not have a label, but they do have a hint text directly in the input field.

The concept of hint text could also be used for Password Safe's text input fields to give the application a more modern and simplified look. However, the hint text is hidden as soon as the input field becomes focus or there is text. In order to still provide information to the user about the purpose of the input field, a tooltip could be used.

## Dialogs that prompt the user to take action

In each or only certain dialogs, there could be a banner at the top. The banner could consist of a background image, a gradient, or just a specific background color. The banner would contain a title in large, bold font and text in normal font addressed to the user. See also [wxBannerWindow](https://docs.wxwidgets.org/3.2/classwx_banner_window.html) for the idea.

## Draft
A first version of the initial dialog serves here as a draft to be able to imagine these ideas. The design was done partly by changes to the source code, partly by a painting program to keep the effort low.

The orange text might appear in a different color and the font could be larger. Removing the keyboard button would leave more room for the other buttons and both text entry fields would have the same length.

![grafik](https://github.com/pwsafe/pwsafe/assets/4042917/28ff582f-8e18-48ee-9757-4bfc1bd91825)

igor2x:
@rafaelx, thanks for starting this topic.

Yes, I think some of the UI could be improved or modernized. For example toolbar icons are colorful, which is "old" design. This "colors" design get us in trouble specially in dark mode. Dark mode, despite I don't use it much, is standard now days and apps that do not support it are some kind of second class apps.

[...]

Issue #1134, opened 2024-02-24
On Exit button click data lost appears


igor2x:
Password Safe v1.18.2 on Ubuntu 23.10.

WARNING: Be careful this can be dangerous to test, because data lost appears. I suggest to create some dummy database and test with it.
1. Open password database.
2. Manage | Options | Backup tab | "Save database immediately after any change.
3. Now delete one entry and confirm delete action. Remember entry name that was deleted.
4. Lock button from toolbar.
5. File | Unlock Safe.
6. Dialog appears. Click on Exit button. Password Safe closes down.
7. Open Password Safe and open database and see member from step 3 was gone. Because option from step 2 is set, I should explicitly save any change, but in this case Password Safe just saved the changes.

I know this can be fixed with some additional checking if option from step 2 is checked/unchecked, but in my humble opinion Exit button is:
a) Unusual in programs in the first place. User can always click on "x" from top-right from dialog and then "x" from main program, if user really wants to exit.
b) There may be some other dangerous combinations of settings that Exit button may explicitly ignore. This is just dangerous to have in the first place.

My suggestion completely remove Exit button from unlock dialog.

igor2x:
I have found another very similar data lost situation.
1-4 steps the same as above.
5. Click to "x" button on top-right and program is closed down without asking to save or discard the changes and changes are just saved.

It looks like there are issues with Manage | Options | Backup tab | "Save database immediately after any change" set to unchecked.

Maybe there are even some other way to trick Password Safe into quitting without saving.

igor2x:
Current issue is not relevant anymore, because [pull request 1231](https://github.com/pwsafe/pwsafe/pull/1231) removed Exit button. I am closing this issue down.

Issue #1104, opened 2024-02-15
Small "nitpicking" to make program simpler and more consistent


igor2x:
Password Safe v1.18.2 on Ubuntu 23.10. I know to some of people this small issues may look trivial, but getting through them all over and over again it would be nice if this could be fixed. You know beauty is feature.  :-)

**1. Branding**
On Password Safe on Windows the icon is always red, that is official icon of program. In Linux this official icon is black. I see no need for that and the main branding of product is lost. I suggest to replace black icon with red icon.
![image](https://github.com/pwsafe/pwsafe/assets/10895030/9c1b640c-2d65-4ae3-8731-1cffaa479ac8)

**2. Main window several small issues**
a) At top left corner there is small black icon. There should be no icon at all. Looking at other Linux programs and there is no icon in any of the program. Icon somehow indicates that you can click on it and some action will follow, but nothing happens clicking on icon. I suggest to remove the icon, because it is redundant.
b) In the middle of title bar: In the programs I have look at (e.g. text editors, image editors,...) the order of current opened file and the program itself is first opened document and then program name so: "pwsafe.psafe3 - Password Safe". And another one is this name concatenation "PasswordSafe" instead of "Password Safe". It looks like a bug, someone forgot to add a space, also the name of the program it should be clear what it is. The branding of program should be clear what it is and used consistently, to reduce confusion. Smaller confusion, simpler to use a program.
c) bottom right the URL is something very unusual and distracting. I suggest to remove it. Less distractions simpler to use a program. This kind of info belongs to Help menu.

![image](https://github.com/pwsafe/pwsafe/assets/10895030/f7ba8965-2f69-4659-9c16-5138e9e231df)

**3. Remove non-essential elements**
Remove anything that is distraction, to simplify the usage and reduce confusion. I see no need for icon there inside in the middle of the program. For other wording to make program simpler I have already written suggestions in [Yes/No buttons issue.](https://github.com/pwsafe/pwsafe/issues/1020)

![image](https://github.com/pwsafe/pwsafe/assets/10895030/395ef8a2-0286-46ba-8b29-42a0852259b4)

**4. More love to the About page**
Like it is now Help | About is little bit crowded. Left side of image is current image, right site is my suggestion.
a) At the top there is the most important part. The name of the program is just too small and the "branding" name is lacking space between Password and Safe.
b) All of the lines it would look better if they are aligned to the middle vs. current left.
[...]

PR #1215, opened 2024-03-29
[wx] Tweaks to the layout of SafeCombinationPromptDlg

### LINKED ISSUE #1124: Yes/No buttons: Unlock dialog and similar issues
One additional this time OK/Cancel.

Lock the database and unlock it and bellow dialog appears.

image

1. OK/Cancel combination is also not user friendly:
- change **OK** with **Unlock**
- **Cancel** button is fine like it is now

The main mantra should be, from the button names it should be absolutely clear what ACTION will follow. OK doesn't really tell anything useful. To understand how design of dialog should be made, imagine you clicked on some ACTION from main application window with blind folded and now you take off blind fold and only think you see is input box and confirmation buttons like this:
image

From above "Cancel" we can assume that dialog will disappear and we will be moved logically one action into the past. But what does "OK" button do? We have no idea. Now it makes more sense it should be "Unlock" button.

Why is this important. Working with program should be quick. Password Safe is not the program I spend hours a day, but password manager is mainly distraction of regular work. Like opening some program I am focused and working in, I just need to enter the userid and password I don't remember, so distraction switch to password manager and get the userid and password. If I am forced to read text, this distraction gets tiring.

2. Exit button.
In my humble opinion **Exit** button is by user unusual and unexpected feature and most probably very rarely used. What users expect to exit a program? Click on Cancel button and then in main program click on top right "X" to close a program.

3. Icon on the left.
What is the point of the Password Safe icon on the left. It doesn't tell user anything useful, it is actually only a distraction. General guidance should be minimalism, use only the UI components that are essential for solving the problem.

rafaelx:
Resolves #1124 

Fixes an assertion regarding layout constraints with YubiKey control elements

> ./src/common/wincmn.cpp(1068): assert ""(minW == wxDefaultCoord || maxW == wxDefaultCoord || minW <= maxW) && (minH == wxDefaultCoord || maxH == wxDefaultCoord || minH <= maxH)"" failed in DoSetSizeHints(): min width/height must be less than max width/height!

Current Version of Unlock Dialog without YubiKey
UnlockDialog

New Version of Unlock Dialog without YubiKey
UnlockDialogUpdate

Current Version of Unlock Dialog with YubiKey
UnlockDialogYubi

New Version of Unlock Dialog with YubiKey
UnlockDialogYubiUpdate

Note
After studying some documentation on creating UI designs, the common recommendation seems to be to position the label above the text input field. Visually, I like this representation better, especially when there are several text input fields with labels of different lengths. The length of the text input fields remains independent of the label length and its translation. If there are no objections, then I would like to apply this recommendation to other dialogs as a first step in regards to #1197.

A margin of 5 pixels was used between the label and the text input field and a margin of 12 pixels was used for all other controls.


igor2x:
@rafaelx, excellent that you have started working on UI and UX.

**In general**
Yes UI guidelines suggest to add text above input box. I like this idea. The first issue that usually appears is spacing, incorrect spacing (or no spacing at all) and input boxes looks crowded. There should be text+input_box+spacing+text+input_box. Another problem that can happen is spaghettification - short wide dialogs with very long length dialogs. In my humble opinion we should do the design of New Entry to see what it looks like when there are multiple input boxes. In my humble opinion we should only have few of the must have input boxes and all other hidden inside the "> Additional options" (see File | Compare on Linux).

**Unlock dialog**
Good design is make is simple, make only the options that are vital to fix the problem all other must be removed or hidden inside some "Advance" button or similar.

I will comment only on Yubikey dialog, which is actually ordinary dialog plus additional things.

image

The first thing we need to ask is, from minimalistic point of view is hide everything except caption, input box and buttons like this (the blurry look, like many users gets when UI gets too much crowd). 
image
What is this dialog all about? It is about unlocking database, isn't it?
[2] The title bar from current "Enter Master Password" should be "Unlock password database".
[11] Cancel button is fine, we know what it does, it closes dialog without any changes. Name of OK button is not good, because from above blurred image user does not know what OK button will do. Change "OK" label with "Unlock". By the way, I also don't like icons on buttons. Remove them.
[1] Icon in top-left is useless, specially in Linux that does not do anything. Consider removing it.
[3] What is purpose of those buttons? I don't know what they actually do. Aren't they just something that should be removed on simple dialog as this one, that is asking for password only? If not, please explain what is purpose of this buttons?
[4] Now we have text "Enter the Master..."+spacing+actual_path_to_database+spacing+Master Password+input box. How many groups of information do we have? It looks many, but actually spacing [4] should just be removed.
[5] What is the purpose of "Open as read-only". User can't select this in any way. It is not dependent on any other option. This should be removed.
[6] Can we make size of keyboard icon the same as the width of input box?
[7] Show Master Password should get to the left where is not location of "Open as read-only" or even better would be in eye icon at the right side of input box, but I don't know if current GUI widgets support this.
[8] Yubi key icon (if we really need it), should be put to the left. Maybe instead of icon itself, just color text "Insert your Yubikey" in green.
[9] Do we really need "Help" button? Is this dialog really so complicated that user needs some help to read? If yes, then we are doing something very wrong. Help should be offered only when we except that there is high probability user will actually use it. In this case it is just clutter. Remove "Help" button.
[10] Very unusual an potential dangerous action is performed by "Exit" button. In all of other programs, user just needs to click on "x" on top-right to close dialog and then another "x" on top-right to close a program. I assume "Exit" is exceptionally rarely used. It is clutter, remove the "Exit" button.

I imagine the dialog could look like this:
image

P.S. I don't know how much of an use is Keyboard icon. It does not work on my Linux+Wayland at all. Maybe consider removing this "keyboard" feature. Is it really needed?

igor2x:
Actually if we look little bit deeper, additional optimization:
- remove keyboard icon (this removes one of the features; decision should be made)
- Actually we can remove whole text "Enter the Master Password to unlock the password database" text. Why? Because now we already have "Unlock password database" in the title bar, so no need to repeat. And what actually follows bellow this text is "Password Database".
image

This is actually great example, how we can strip down dialogs from fully packed rarely used if at all features to bare bone, to the point what dialog is all about.

Note: Maybe some width/height can be adjusted.

igor2x:
Thinking little bit further:
- How to get rid of color after "Password Database:"?
- How to make database path more beautiful (greyed out input box).

image

rafaelx:
> [1] Icon in top-left is useless, specially in Linux that does not do anything. Consider removing it.

As I see it, at least with Kubuntu (KDE/Plasma), all applications place the application icon in their dialogs on the left side in the title bar.

> [3] What is purpose of those buttons? I don't know what they actually do. Aren't they just something that should be removed on simple dialog as this one, that is asking for password only? If not, please explain what is purpose of this buttons?

These are the typical buttons for minimizing and maximizing a dialog. I also see these buttons in dialogs of other applications.

> [5] What is the purpose of "Open as read-only". User can't select this in any way. It is not dependent on any other option. This should be removed.

Good question. It appears that the implementation of this has not been completed. Should probably correspond to the functionality in the Windows version.

> [6] Can we make size of keyboard icon the same as the width of input box?

Such a version can be seen in the screenshot in #1197. To keep the functionality of the virtual keyboard, I placed the button at the bottom right of the unlock dialog. This button even appears three times in the password change dialog. To start the virtual keyboard, only one such button is needed. The password change dialog also needs to be revised.

> [7] Show Master Password should get to the left where is not location of "Open as read-only" or even better would be in eye icon at the right side of input box, but I don't know if current GUI widgets support this.

The "Open as read-only" checkbox was replaced by an eye icon, as also already shown in #1197. Unfortunately, it seems that the wx framework does not support having an eye icon directly in the text entry field.

> [8] Yubi key icon (if we really need it), should be put to the left. Maybe instead of icon itself, just color text "Insert your Yubikey" in green.

It is an button which is needed.

> [10] Very unusual an potential dangerous action is performed by "Exit" button. In all of other programs, user just needs to click on "x" on top-right to close dialog and then another "x" on top-right to close a program. I assume "Exit" is exceptionally rarely used. It is clutter, remove the "Exit" button.

If there is an exit button, I see an advantage for the user. One use case could be when the user has finished working on the computer and now wants to shut it down quickly. Password Safe may be minimized and locked after a long period of non-use. The user does not have to enter their super secure and highly complicated password to unlock in order to exit the application.

Here is an intermediate status of the unlock dialog.
UnlockDialogUpdate2

UnlockDialogYubiUpdate2

Eye-Icon using a dark theme.
UnlockDialogYubiUpdate2DarkTheme

igor2x:
@rafaelx, excellent, it looks much better (simpler).

It looks like we have opposite philosophy, you would like to keep as much as possible (to have multiple features), I would like to remove as much as possible (to have simplicity). I think it is fine, we have different perspective, so we can discus and make good compromise and result one good "template" how to change all of the dialogs.

a) Password Safe Icon on the top left. On Gnome no application is using icon on top-left titlebar. I don't know for other Linux distribution and MacOS. On Windows I see a point of this icon, because icon is not just for  decorative purpose, but clicking on it and small menu appears with Minimize, Close etc menu options. Double click on this icon closes the window (just like "x" on the top-right title). But on Gnome icon has no effect. Because I am used to Windows, I double click on icon and nothing happens. I looked at the Windows again, on my work computer I see all of the programs use icon on top-left title bar except for Microsoft products. It looks like Microsoft changed its mind about left-side title bar icon. I see no icon in Microsoft Files, Microsoft Office, Microsoft Edge etc.
b) Top-right "up" and" "down" icon, I see this are Min and Max from some OS I am not using (I use Windows and Linux). I see a point for Minimize icon now, just click on Minimize and whole program (not just a dialog, but background program) is minimized. For Maximize, I don't know if it is really needed, I think only "Min and Close" is enough.
c) I don't have Yubikey, so I didn't understand Yubikey is a button. Actually from design (grey button background) I haven't figure it out it is actually a button or at least it is clickable button. If it is a button, why doesn't it look like a button? The same color for: border, font text and background color like other buttons? If I understand correctly Yubikey is feature instead of Master Password.
d) Is it necessary for "Yubikey" button to have a text on icon? I think the users that are using Yubikey already know the icon itself. We don't use "Keyboard" text next to the keyboard icon. In my humble opinion Keyboard, Eye and Yubikey should have the same design, icon instead of text. The same color border and the same background color.
e) Text "Insert your Yubikey": why is there a word "your"? What if I don't insert "my" Yubikey, but someones else or shared key, like husband and wife key, it is "our" key. I think word "your" is safe to be removed and just use "Insert Yubikey".
f) If we are keeping Keyboard icon, it should be in the same line as input box, because it is related to the input box. See for example Manage | Change Master Password. Each field has keyboard icon, because when using keyboard icon, keyboard is started and then after password is entered using virtual keyboard text gets inserted into current input box. In my humble opinion keyboard should be removed as it is probably very rarely used or at least to make an option in Options to enable it if advanced users want to use it.
g) I restarted Windows and Windows did not ask me to close Password Safe program. Like I understand  Windows sends request to close a program to every running program and tries to make sure each of the program does the "clean exit program" routine, like saving some text, removing lock file or similar. It looks to me that Password Safe closes down without an issue, so Windows do not need to ask for force close confirmation before Windows restarting. I know I have used "Exit" button just once and I used it, because I mistakenly clicked on it instead of Cancel button and was surprised that Password Safe "crashed". I think we should remove Exit button, my humble opinion. If we really really really want to keep this button then we need to separate them from other much more used Cancel and Unlock buttons.

New proposal:
image

h) About Yubikey, it still looks to me cluttered, but I don't use Yubikey and don't know what is purpose of "Insert Yubikey" text? I don't see this text on #1197. If text is not required, then "Yubikey" icon next to the keyboard, as suggested in #1197 would probably be better looking.

nobugshere:
I would like to add some (hopefully helpful) observations to the discussion:
[1][3] On Linux, the layout of the title bar does not follow a single convention, as it is does Windows or macOS.  It can depend on the desktop environment, window manager, and theme selected by the user.  wxWidgets just tries to "fit-in".  For instance on Fedora 39, with the Xfce desktop, it looks like this:
Screenshot_2024-04-01_09-27-57
Note the top-left does not display the icon, but is a system provided menu.  That appears to be typical in this environment.

This particular dialog is used to unlock a safe that has already been opened, as opposed to the initial dialog used when the program is launched.
[5] The "Open as read-only" text and checkbox are greyed-out because the user cannot change it.  However, it does reflect the status used when the safe was first opened.  In the example above, the safe was opened read-only and will continue to be after it is unlocked.  It's a useful indication, but perhaps there is a better way to show it.

[8] The "Insert Yubikey" text, and its associated button, are only displayed if a Yubikey was present when the safe was opened, but has been removed.  When it's inserted, or if it is already present when the dialog is displayed, the text changes to: "Click the button labeled 'Yubikey' in green".

rafaelx:
@igor2x I think we're not too far apart in our opinions. I really appreciate your suggestions and find the exchange of ideas very motivating. As you have probably noticed, the discussion takes place between very few people. The question that arises here is whether this small group of people can adequately represent the entire user base. I don't know the entire history of Password Safe and I think that some people who contributed to its development have already thought about certain functionalities and dialogs. Just because I don't need an Exit button, a virtual keyboard, or a Yubi key, I don't want to limit or remove that functionality just because I can. So I have a more conservative view on changes.

I agree that if we find a good compromise, we will lay a foundation for improving other dialogs and will certainly receive Rony's approval for this.

We have three buttons, each of which has a good reason for being positioned to the right of the text input field. The eye button, the keyboard button and the Yubi button. Even if we removed the text from the Yubi button, there would still be quite a few buttons which would limit the space available for the text entry field. The three buttons in a row almost have the character of a toolbar.
As a non-representative Password Safe user, I assume that the virtual keyboard is not used frequently. Entering text using a physical keyboard is certainly much more convenient. I also don't think it makes sense to have several keyboard buttons on one dialog (see password change dialog) if the purpose of the button is only to start the virtual keyboard. If the new password was entered using the virtual keyboard in the password change dialog, the keyboard button would not be pressed again in the password confirmation input field. My suggestion is therefore to have only one keyboard button per dialog, positioned in the bottom right corner. This functionality is therefore still available to users who want to use it, and the row of buttons is reduced in favor of simplicity.

The Yubi controls follow the approach of descriptive text first, then the control to operate. As with the text input fields. Due to the vertical arrangement of text (Insert Yubikey) and button, both pieces of information are close together and not scattered across the entire width of a dialog. The Yubi relevant text and associated button both have the character of a separate notification ("Hey user, give me a Yubi key and then press the Yubi button."), which usually appear in the application as message boxes and prompt the user to make a decision. Hence the centered positioning of the two elements.

The exit button and keyboard button are slightly separated from the main buttons.

Here is my current favorite.
UnlockDialogYubiUpdate3

UnlockDialogYubiUpdate4

igor2x:
@nobugshere, thanks for jumping in. The more the merrier.
[1] Great that wxWidgets just fits-in with buttons and icons to look more natural in each of system.
[5] Checked or unchecked on greyed out "Open as read-ony" has some information, but in my humble opinion if user can't do anything at the moment of unlock, this information is irrelevant at this moment and even distracting. User will deal with "read-only" or "read-write" after unlock is performed. In [issue 1104](https://github.com/pwsafe/pwsafe/issues/1104#issuecomment-1948195379) I have suggested how R-O vs. R/W could be way more visible.

@rafaelx, OK, lets leave buttons on the screen, removing them is way simpler then adding them back (with code included).

Sorry I don't have Yubikey, so my understanding is not from practical experience, but instead from Youtube videos or from Password Safe documentation.

**[1] Yubikey text**
In general, dialog should tell user what user has to do and what action will follow. For example: "Enter password to unlock the database."
What user has to do: "Enter password..."
What action will follow: "... to unlock the database."

I suggest instead of "Click the button labeled 'Yubikey' in green" text, display text "Activate Yubikey". From the button itself it is obvious button must be clicked (it got from greyed out to clickable state).

**[2] Yubikey button and text location** 
Somehow I think there is too much empty space in your previous post on left or right of text & YubiKey button. Because Yubikey is related to Master Password, that is why I have added it (in my previous post) to the bottom of Master Password field.

**[3] Green color button**
What is the purpose of green color of Yubikey? In GUI design anything should have its purpose, green color, what purpose? Is it just for branding?

ronys:
Excellent discussion, thanks to all participants.

My input:
1. The "open as read-only" checkbox is needed, as there are scenarios where the user will definitely want this from the start (as opposed to switching to read-only after opening).
2. Fields may appear too spaced out because in another language the label is much longer. Any suggestion on how to handle this more elegantly welcome.
3. The Yubikey button is green when a Yubikey is connected. In more detail:Yubikey controls are shown only if a Yubikey is detected. If it's detected, the button is green. If, after being detected, the user removes the Yubikey, the button turns grey/disabled, until the yubikey is re-inserted.

igor2x:
@ronys, thanks for joining discussion.

[1] In this PR we are discussing about dialog when user Locks the database and then Unlocks database, so "Unlock database" dialog. In this case "Open read-only" checkbox is always greyed out and user can't change this setting. In my humble opinion, if user can't do anything to make a change, then this option is just a clutter and simplification is possible. 

"Unlock database" dialog is very similar to "Master Password Entry", the first dialog when Password Safe is opened. But at this first dialog it is useful for user to have an option to have "Open read-only" checkbox.

@rafaelx, now I have checked the first dialog when Password Safe is started up. This dialog is very similar to "Unlock database" and this first dialog does not have Exit button. In my humble option Exit button should be removed everywhere in the application, not only because it is clutter, but only because we need to make care we don't get into the lose data problem. I have reported [issue 1134](https://github.com/pwsafe/pwsafe/issues/1134) about data lost when Exit button is used. So, if we have such a functionality, we need to properly take care of not losing data. But in case of Exit button and if we add additional question to user save or discard the data change further complicates dialog design.

ronys:
Technically, there's no reason to disable the "Open read-only" checkbox in the unlock database. I'd rather allow the user to do this than remove the checkbox entirely.

rafaelx:
Convinced by @igor2x's arguments, my votes are as follows.
1b*, 2b*, 3b, 4a, 5c**, 6a

*The unlock dialog should focus on unlocking the database and should follow the principle KISS - Keep it Stupid and Simple or just simplicity. Further functionality is available to the user in the main window.
**Since I'm not a Yubikey user, I don't know the exact procedure for using a Yubikey. Maybe it would make sense to have the Yubikey button to the right of the password input field, if the Yubikey button has to be pressed after entering the password. The Yubikey button may already be located to the right of the input field in the initial dialog for exactly this reason.

Here's a new suggestion.
UnlockDialogYubiUpdate5

ronys:
Hi,

Starting with @igor2x 's last point: In the Windows implementation, the code is indeed common, and the runtime display and behavior is controlled by a few enums:

```
// Index values for which dialog to show during GetAndCheckPassword
enum {
      // Normal dialogs
      GCP_FIRST         = 0,   // At startup of PWS
      GCP_NORMAL        = 1,   // Only OK, CANCEL & HELP buttons
      GCP_RESTORE       = 2,   // Only OK, CANCEL & HELP buttons
      GCP_WITHEXIT      = 3,   // OK, CANCEL, EXIT & HELP buttons
      GCP_CHANGEMODE    = 4,   // Only OK, CANCEL & HELP buttons

      NUM_PER_ENVIRONMENT = 5,
};

// GCP read only flags - tested via AND, set via OR, must be power of 2.
enum {GCP_READONLY = 1,
      GCP_FORCEREADONLY = 2,
      GCP_HIDEREADONLY = 4,
      GCP_APP_WINDOW = 8};
```
If I were to refactor, I'd get rid of this, as the code is ugly and painful to maintain. So for wx I'm happy to keep these classes separate.

As to the decisions regarding the unlock dialog:
1. Agree with 'b' - removing the Open read-only checkbox
2. Exit button: I'd rather keep it, adding logic to prevent unsaved data loss. Rationale: Give users a way to quickly exit the program when the database is locked.
3. Yubikey device removed: 'b', since I find the sudden appearance of buttons more annoying than their being enabled/disabled.
4. 'b' -  Green enabled button without "Yubikey" text on button and text next to it "Activate Yubikey".
5. Note that when the Yubikey button is clicked a 'progress bar' appears to show the user how long we're waiting for the button on the Yubikey device to be clicked. I think that 'a' works best with this.
6. 'a', as discussed above.

rafaelx:
A new version based on the last statements.

UnlockDialog1

UnlockDialog2

UnlockDialog3

UnlockDialog4

Have all suggestions been taken into account? Any change requests?

igor2x:
@rafaelx. This looks exactly as we have discussed and also take in account what @ronys has suggested.

---

Just wondering, will this change only apply to wxWidgets or also for Windows? It would be nice if both gets "modernized look".
I see no "Exit" button on Windows. Assuming wxWidgets version was created according to Windows dialogs, it looks like "Exit" button was already removed from Windows version, but is still not wxWidgets.
image

If we want to have the same look and feel on wxWidgets and Windows:
1. Remove "Exit" button on wxWidgets or
2. Add "Exit" button on Windows (including logic to prevent data lost).

rafaelx:
@igor2x I only contribute to the Wx version from time to time.

igor2x:
@rafaelx, thanks for explaining. No worries, just implement the wxWidgets version. I can't wait to start playing with new dialogs.

My suggestion for Windows version: For each dialog in wx that is modernized, we just open an new Github issue, for reminder that the same change is performed in future in Windows version. In my humble opinion if good and modern design is implemented in wxWidgets, then there is like half of job done for Windows, because Windows developer just copy/paste the design.

I think discussion in this PR was hugely beneficial for all the dialogs we can modernize in the same way. @rafaelx, fantastic, that you have started this journey and @ronys for vital insider perspective.

ronys:
Regarding the "Exit" button in unlock on Windows:
Currently this is shown in one case only: When the unlock dialog is shown, and the system tray icon is **not** in use.

It's arguable if this is still needed, since users can right-click on the icon in the taskbar and select "Close Window", as with any Windows app.

igor2x:
@ronys, @rafaelx I have opened new [issue 1222](https://github.com/pwsafe/pwsafe/issues/1222) for Exit button discussion, because in my humble opinion, we are slowing down implementation of this PR, for trivial decision. We are also trying to solve too many issues with single PR.

There are two stages of discussion:
- Early stage: a lot of ideas and solutions, solutions are pretty quickly narrowed down to almost final decision. In this stage discussion is vibrant, exciting and user involved.
- Late stage: a lot of discussion for trivial decisions, a lot of time spend, every one gets bored, user actions get stalled. In this stage someone must cut the discussion and create separate topics, with each separate issue in separate topic. In larger system, every individual smaller issue can be discussed and solved in separate team.

In my  humble proper path should be:
1. Implement solution just like final proposal was made in [@rafalex post above]( https://github.com/pwsafe/pwsafe/pull/1215#issuecomment-2048541502) (with Exit button included) and only this should be topic of this PR.
2. Exit button discussion is continued in separate [issue 1222.](https://github.com/pwsafe/pwsafe/issues/1222) Topic: do we want to keep the button or not. We can still remove the button in separate PR, if we decide to. Removing Exit button is trivial from application developer point of view, so no need to slow down actions in this PR.
3. Taking care of data lost Exit button issue in separate PR, that tackles the [issue 1134](https://github.com/pwsafe/pwsafe/issues/1134).

rafaelx:
The focus of this pull request is on the unlock dialog, but it also impacts other dialogs that request a password from the user. This is because the button to launch the virtual keyboard has been replaced with a button with an eye icon that can be used to show or hide the password in the text input field. This means that other dialogs now have the existing checkbox for showing or hiding the password in addition to the eye icon button.

In other dialogs the keyboard button is no longer present and needs to be added again.
Another side effect concerns the YubiKey button. YubiKey related button text has been removed.

In preparation for the upcoming changes to the other dialogs, the functions 'CreateLabeledSafeCombinationCtrl' and 'CreateYubiKeyControls' were created in wxUtilities.cpp to achieve uniform placement in terms of alignment and spacing of the controls. These functions were added to a namespace called wxUtilities to better identify the origin of these functions in the source code. Some of the existing functions could also be added to this namespace. (separate task beside the dialog changes)

The update of the affected dialogs is being tracked via a separate ticket/issue on GitHub.

nobugshere (commented):
Due to the changes in wxUtilities.h, Xcode wants an additional include in media.cpp:
```
#include <wx/stattext.h>

```

Why Xcode and not gcc?  I dunno .

rafaelx:
Does adding '#include <wx/stattext.h>' to 'wxUtilities.h' fix the problem?

nobugshere:
Yes, it does.  That makes more sense.

nobugshere:
On the Mac, the new text boxes seem oddly proportioned, compared to the font size, and the text is top aligned instead of vertically centered.  Note the dropdown list in the first box hasn't changed.  Does it have something to do with the show text button?

Screenshot 2024-04-14 at 2 41 34 PM
Screenshot 2024-04-14 at 2 42 05 PM

ronys:
The changes to wxUtilities.h broke the mac-pwsafe build. Looks like some refactoring is needed:

In file included from /Users/runner/work/pwsafe/pwsafe/src/os/mac/media.cpp:15:
/Users/runner/work/pwsafe/pwsafe/src/ui/wxWidgets/wxUtilities.h:227:31: error: use of undeclared identifier 'wxStaticText'
  std::tuple<wxBitmapButton*, wxStaticText*> CreateYubiKeyControls(wxWindow *parent, wxWindowID buttonId, wxWindowID statusTextId);
                              ^
/Users/runner/work/pwsafe/pwsafe/src/ui/wxWidgets/wxUtilities.h:227:44: error: expected expression
  std::tuple<wxBitmapButton*, wxStaticText*> CreateYubiKeyControls(wxWindow *parent, wxWindowID buttonId, wxWindowID statusTextId);
                                           ^
/Users/runner/work/pwsafe/pwsafe/src/ui/wxWidgets/wxUtilities.h:235:71: error: use of undeclared identifier 'wxStaticText'
  wxBitmapButton* GetYubiKeyButtonControl(std::tuple<wxBitmapButton*, wxStaticText*>& controls);
                                                                      ^
/Users/runner/work/pwsafe/pwsafe/src/ui/wxWidgets/wxUtilities.h:235:84: error: expected expression
  wxBitmapButton* GetYubiKeyButtonControl(std::tuple<wxBitmapButton*, wxStaticText*>& controls);
                                                                                   ^
/Users/runner/work/pwsafe/pwsafe/src/ui/wxWidgets/wxUtilities.h:243:3: error: unknown type name 'wxStaticText'; did you mean 'wxStaticBox'?
  wxStaticText* GetYubiKeyStatusControl(std::tuple<wxBitmapButton*, wxStaticText*>& controls);
  ^~~~~~~~~~~~
  wxStaticBox
In file included from /Users/runner/work/pwsafe/pwsafe/src/os/mac/media.cpp:15:
In file included from /Users/runner/work/pwsafe/pwsafe/src/ui/wxWidgets/wxUtilities.h:29:
In file included from /Users/runner/work/pwsafe/pwsafe/src/ui/wxWidgets/SafeCombinationCtrl.h:16:
/Users/runner/wxbuild/include/wx-3.2/wx/sizer.h:1064:28: note: 'wxStaticBox' declared here
class WXDLLIMPEXP_FWD_CORE wxStaticBox;

rafaelx:
@nobugshere Interesting, the control elements on the Mac seem to be lower in height. Could you please determine the height of one of the buttons and a text input field?

`pws_os::Trace(
    L"Button Size: w=%d h=%d",
    (unlockButton->GetSize()).GetWidth(),
    (unlockButton->GetSize()).GetHeight()
 );`

The size of the eye icon is 27 x 27 pixels, which may be a bit too large. The keyboard icon appears to be the appropriate size at 31x16 pixels. Maybe the eye icon should also be reduced to a height of 16 pixels?!

rafaelx:
I experimented a bit with the size of the icons and reduced the height of the eye icon to 16 pixels. Under Linux the icon looks much too small. Since the text input field is higher, there is unused space above and below the icon. Another effect is a significant loss of quality due to the strong downscaling. Maybe separate icons should be created for the Mac version.

UnlockDialogReducedIconSize

ronys:
Quite an improvement! Following are some minor adjustments that will make this ready for merging into master (breaking up into several comments, one per dialog)
image
1. After clicking once on the eye button, there's a thin rectangular border around it that doesn't go away. If possible, this should be cleaned up
2. The Exit button should be removed, not disabled.

ronys:
image

Very odd: When the Yubikey is inserted *after* the dialog is displayed, the button and text are incorrectly located. If, however, the Yubikey is already inserted when pwsafe is started, everything looks fine:
image

ronys:
The initial dialog is a bit problematic:
image
1. The "Show Master Password" checkbox is redundant, and should be removed
2. The "New..." button is too small. I think it should also have more space around it to indicate that it's not related to the checkbox
3. The controls should be more aligned. For example, the "Open as read-only" checkbox should be aligned with the left edge of the Master Password edit control, and the right edge of the edit control should be aligned with the right edge of the Password Database drop-down.

rafaelx:
> image
> 
> Very odd: When the Yubikey is inserted _after_ the dialog is displayed, the button and text are incorrectly located. If, however, the Yubikey is already inserted when pwsafe is started, everything looks fine: image

I'm assuming the pwsafe version you're using is based on wx 3.0.5 (Debian?), which results in the "Exit" button being disabled. This would mean that the exit button would also be disabled in a build based on master. See also PasswordSafeFrame::CanCloseDialogs and its use in PasswordSafeFrame::UnlockSafe when creating the unlock dialog.

The rectangular frame around the eye icon is the typical button border, which should be disabled by the style flag used. Could also be an effect of wx 3.0.5 that does not occur in newer versions of wx. What's strange is that this frame no longer appears in the screenshot with the YubiKey button.

The differences between the wx versions and the platforms don't necessarily make it any easier. :-/
Would it be possible to change the Debian build to a more recent wx version?

igor2x:
Suggestion:
1. Completely remove "New" button and add icon instead (benefit we have no idea how wide should button be for non-English languages, it may be large)
2. Currently is "New" icon is at "password" field and from design it looks like this is something related to password like new password. I suggest to move "new database" icon at the place where database field is.
3. From best looking it would be nice if icons are the same size.
Note: I just added some new icon from web to demonstrate idea, maybe we can find better one.

image

rafaelx:
> Suggestion:
> 
>     1. Completely remove "New" button and add icon instead (benefit we have no idea how wide should button be for non-English languages, it may be large)
> 
>     2. Currently is "New" icon is at "password" field and from design it looks like this is something related to password like new password. I suggest to move "new database" icon at the place where database field is.
> 
>     3. From best looking it would be nice if icons are the same size.
>        Note: I just added some new icon from web to demonstrate idea, maybe we can find better one.
> 
> 
> image

I also think that the "New" button should be placed to the right of the "Browse" button due to its relation to the database, as seen in the prototype in #1197.

However, we should limit the discussion here to the Unlock dialog, otherwise the topics will be too mixed up here. As has already been suggested, a separate ticket/issue per dialog for discussion.

igor2x:
This PR is related to Unlock Database and I suggest to not go beyond this issue frame.
I think additional issues in another dialogs should be tackled in separate issue/PR.

I have opened new [issue 1225](https://github.com/pwsafe/pwsafe/issues/1225) related to my suggestion how to improve "Master Password Database". May I invite you to discuss this in separate issue.

nobugshere:
@rafaelx,
I seem to be falling behind :-). Here are some sizes based on the same version as the screenshot I posted yesterday.  
macOS 14.4.1, Xcode 15.3, wx 3.2.4
rafaelx/GH1124 commit [255b099](https://github.com/rafaelx/pwsafe/commit/255b099d9db3b9cffe18d970b9779b0641216cab)

Unlock Button Size: w=84 h=20
textCtrlFilename Size: w=64 h=35
m_YubiBtn Size: w=35 h=35

rafaelx:
> @rafaelx, I seem to be falling behind :-). Here are some sizes based on the same version as the screenshot I posted yesterday. macOS 14.4.1, Xcode 15.3, wx 3.2.4 rafaelx/GH1124 commit [255b099](https://github.com/rafaelx/pwsafe/commit/255b099d9db3b9cffe18d970b9779b0641216cab)
> 
> Unlock Button Size: w=84 h=20 textCtrlFilename Size: w=64 h=35 m_YubiBtn Size: w=35 h=35

Thank you for providing this information.
Interesting, the buttons are getting a height of 20 pixels on the Mac.
I'm very curious to see how it looks like on the Mac with the reduced eye icons.  8^)

nobugshere:
@rafaelx, 
Here you go.  These are from rafaelx/GH1124 commit [760b0d3](https://github.com/rafaelx/pwsafe/commit/760b0d34b7c4e59c93a4894af8dbdae590458669).   Looks much better to me though, as you noted, the eye icon looks a little rough.  I should note these are rendered for a 3440x1440 display, not a HiDPI display.  Also, the Help (?) button looks a little lonely.
Screenshot 2024-04-16 at 11 07 20 AM
Screenshot 2024-04-16 at 11 07 43 AM

nobugshere:
Same as above with a Yubikey inserted before launch.
Screenshot 2024-04-16 at 11 25 20 AM
Screenshot 2024-04-16 at 11 25 38 AM

rafaelx:
Many thanks for the screenshots. It's really helpful to see what it looks like on the Mac.
I think that then two sets of icons with different sizes would be a solution. A set of ugly icons for Mac and a set of beautiful icons for Linux. I'm just kidding. The icons can certainly be optimized in terms of appearance.

nobugshere:
Looks okay, but for some reason, the text aligns to the top and I don't see a way to change that.  But, at this size, it's not bad.
Screenshot 2024-04-16 at 3 00 49 PM
Screenshot 2024-04-16 at 3 42 49 PM

Issue #1226, opened 2024-04-17
[wx] Modernization of Change Password Dialog


rafaelx:
This ticket/issue was created to track and discuss the changes to the password change dialog.
![ChangePasswordDialog](https://github.com/pwsafe/pwsafe/assets/4042917/c45cd7c4-e554-411a-aaa4-95cee3dd21f9)
![ChangePasswordDialogYubiInserted](https://github.com/pwsafe/pwsafe/assets/4042917/a9b663d2-e247-494d-afd7-2ce09ce8a95b)
Related to #1197.

igor2x:
1. Looking at text: "Enter the current master password followed by a new one. Type the new one again to confirm it." it looks to me redundant. From three fields design isn't it obvious all three fields should be filled in? In my humble opinion, this text doesn't tell anything that is not really obvious.
I see two options:
a) Remove above text and that is (if we want to go ultra simplistic).
b) But... in my humble opinion, because this is potentially dangerous action and if new master password is not remembered it can lead to critically bad consequences, there should be some text, but explaining what will this change do "Setting new master password and it will be from now on used to unlock Password Safe database." Maybe something similar, I am not a native English speaker, so text probably can be improved.

2. Obviously "Show Master Password" checkbox should be removed, because now we have eye icons. I checked what does Windows use when pressed Alt+Ctrl+Delete to change password. Windows use eye icon at each of fields, just like we have. Selecting "Show Master Password" checkbox it displays all passwords, that may be convenient, but on the other hand how frequent do we change master password? Not really frequent, so having eye icons are fine in my opinion.

3. OK button should be replace with: "Replace" or probably even "Change".

4. Remove icons on buttons.

---

For second Yubikey dialog: I don't have Yubikey... I don't want to comment too much, because my assumptions are probably wrong. Just wondering how to get rid of text bellow or how to simplify the text. Specially "top Yubikey" is what I don't like. Why is bottom Yubikey not greyed out if it should not be clicked?

nobugshere:
I just found a couple of bugs in this dialog.  If the safe only uses a Yubikey (no password):
1. After activating the key as usual, it grays-out the second YK button, expecting you to insert a different key, so you cannot add a PW and keep the same YK.  You can accomplish this, but it takes two passes.  First, change it to PW only, then a second invocation to add the YK back.
[...]

PR #1232, opened 2024-04-20
[wx] Fix layout issue when showing Yubi controls dynamically

rafaelx:
This PR refers to https://github.com/pwsafe/pwsafe/pull/1215#issuecomment-2059550069 and was fixed by updating the dialog's sizer hints so that the dialog is resized to the new required size.

By manually changing the dialog size, the bottom buttons can no longer be reduced below their required minimum size and thus retain their full size.

Another change is that the layout will only be updated once when the Yubi controls are displayed and no longer periodically with the poll rate (0.5 seconds), in the hope that it has a positive effect on performance.

It would be great if a YubiKey owner could test these changes.

### DIFF
--- a/src/ui/wxWidgets/YubiMixin.cpp
+++ b/src/ui/wxWidgets/YubiMixin.cpp
@@ -73,7 +73,7 @@ bool YubiMixin::IsYubiInserted() const
 void YubiMixin::HandlePollingTimer()
 {
   // Show Yubi controls when inserted first time:
-  if (yubiExists() && !m_btn->IsShown() && !m_status->IsShown()) {
+  if (yubiExists() && (!m_btn->IsShown() || !m_status->IsShown())) {
     m_btn->Show(true);
     m_status->Show(true);
 
--- a/src/ui/wxWidgets/YubiMixin.cpp
+++ b/src/ui/wxWidgets/YubiMixin.cpp
@@ -34,12 +34,14 @@ void YubiMixin::SetupMixin(wxEvtHandler *eventHandler, wxWindow *btn, wxWindow *
   m_btn = btn;
   m_status = status;
   m_present = !IsYubiInserted(); // lie to trigger correct actions in timer even
-  // Hide Yubi controls if user doesn't have one:
-  if (m_btn != nullptr) m_btn->Show(yubiExists());
-  if (m_status != nullptr) m_status->Show(yubiExists());
-  if (IsPollingEnabled()) {
-    m_pollingTimer = new wxTimer(eventHandler, timerId);
-    m_pollingTimer->Start(GetPollingInterval());
+  if ((m_btn != nullptr) && (m_status != nullptr)) {
+    // Hide Yubi controls if user doesn't have one:
+    m_btn->Show(yubiExists());
+    m_status->Show(yubiExists());
+    if (IsPollingEnabled()) {
+      m_pollingTimer = new wxTimer(eventHandler, timerId);
+      m_pollingTimer->Start(GetPollingInterval());
+    }
   }
 }
 
@@ -71,11 +73,11 @@ bool YubiMixin::IsYubiInserted() const
 void YubiMixin::HandlePollingTimer()
 {
   // Show Yubi controls when inserted first time:
-  if (yubiExists()) {
+  if (yubiExists() && !m_btn->IsShown() && !m_status->IsShown()) {
     wxWindow *parent = nullptr; // assume both have same parent
-    if (m_btn != nullptr) {m_btn->Show(true); parent = m_btn->GetParent();}
-    if (m_status != nullptr) {m_status->Show(true); parent = m_btn->GetParent();}
-    if (parent != nullptr) parent->Layout();
+    m_btn->Show(true); parent = m_btn->GetParent();
+    m_status->Show(true); parent = m_btn->GetParent();
+    if (parent != nullptr) parent->GetSizer()->SetSizeHints(parent);
   }
 
   // Currently hmac check is blocking (ugh), so no need to check here
--- a/src/ui/wxWidgets/YubiMixin.cpp
+++ b/src/ui/wxWidgets/YubiMixin.cpp
@@ -74,10 +74,15 @@ void YubiMixin::HandlePollingTimer()
 {
   // Show Yubi controls when inserted first time:
   if (yubiExists() && !m_btn->IsShown() && !m_status->IsShown()) {
-    wxWindow *parent = nullptr; // assume both have same parent
-    m_btn->Show(true); parent = m_btn->GetParent();
-    m_status->Show(true); parent = m_btn->GetParent();
-    if (parent != nullptr) parent->GetSizer()->SetSizeHints(parent);
+    m_btn->Show(true);
+    m_status->Show(true);
+
+    // Update the dialog's size hints to honor the required minimum
+    // control sizes as additional controls have appeared.
+    auto *parent = m_btn->GetParent();
+    if (parent != nullptr) {
+      parent->GetSizer()->SetSizeHints(parent);
+    }
   }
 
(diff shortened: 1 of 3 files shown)


src/ui/wxWidgets/YubiMixin.cpp
```diff
@@ -71,11 +73,11 @@ bool YubiMixin::IsYubiInserted() const
...
 {
   // Show Yubi controls when inserted first time:
-  if (yubiExists()) {
+  if (yubiExists() && !m_btn->IsShown() && !m_status->IsShown()) {
     wxWindow *parent = nullptr; // assume both have same parent
-    if (m_btn != nullptr) {m_btn->Show(true); parent = m_btn->GetParent();}
-    if (m_status != nullptr) {m_status->Show(true); parent = m_btn->GetParent();}
-    if (parent != nullptr) parent->Layout();
+    m_btn->Show(true); parent = m_btn->GetParent();
+    m_status->Show(true); parent = m_btn->GetParent();
```
dskrvk, line 79: Unless `Show` modifies the value of `m_btn->GetParent` the second `parent = m_btn->GetParent()` is now pointless.

src/ui/wxWidgets/YubiMixin.cpp
```diff
@@ -71,11 +73,11 @@ bool YubiMixin::IsYubiInserted() const
 void YubiMixin::HandlePollingTimer()
 {
   // Show Yubi controls when inserted first time:
-  if (yubiExists()) {
+  if (yubiExists() && !m_btn->IsShown() && !m_status->IsShown()) {
     wxWindow *parent = nullptr; // assume both have same parent
-    if (m_btn != nullptr) {m_btn->Show(true); parent = m_btn->GetParent();}
-    if (m_status != nullptr) {m_status->Show(true); parent = m_btn->GetParent();}
-    if (parent != nullptr) parent->Layout();
+    m_btn->Show(true); parent = m_btn->GetParent();
```
dskrvk, line 78: It would be better to keep the two statements on separate lines.

ronys:
Unlock Password Database dialog looks great:
image

The Master Password Entry dialog is a bit confusing, as the YubiKey text ("Activate YubiKey", etc.) appears far from the button and unrelated:
image

 Perhaps putting the button and related text on a separate row immediately under the password edit control, and moving the "New..." button to where the yubikey button is currently placed?
I also think the now redundant "Show Master Password" should be removed. Something like:
image
The password edit control can be lengthened a bit, as well as resizing the "New..." button.

What say?

rafaelx:
@ronys 

Have you also tested the Password Change dialog? Is everything OK with the layout there too?

I agree with the suggestions for the Master Password Entry dialog. I would also create a separate issue for this, like I did for the Password Change dialog.

ronys:
On Debian 12, debug build, YubiKey inserted, selecting  Change Master Password in the menu gives the following error message:
image

Here's the backtrace:
```
ASSERT INFO:
./src/common/wincmn.cpp(1068): assert ""(minW == wxDefaultCoord || maxW == wxDefaultCoord || minW <= maxW) && (minH == wxDefaultCoord || maxH == wxDefaultCoord || minH <= maxH)"" failed in DoSetSizeHints(): min width/height must be less than max width/height!

BACKTRACE:

[1] wxWindowBase::DoSetSizeHints(int, int, int, int, int, int)
[2] wxTopLevelWindowGTK::DoSetSizeHints(int, int, int, int, int, int)
[3] wxTopLevelWindowBase::SetMinSize(wxSize const&)
[4] wxTopLevelWindowGTK::SetMinSize(wxSize const&)
[5] wxWindowBase::WXSetInitialFittingClientSize(int, wxSizer*)
[6] wxTopLevelWindowGTK::GTKUpdateClientSizeIfNecessary()
[7] wxTopLevelWindowGTK::Show(bool)
[8] wxDialog::Show(bool)
[9] wxDialog::ShowModal()
[10] wxEvtHandler::TryHereOnly(wxEvent&)
[11] wxEvtHandler::ProcessEventLocally(wxEvent&)
[12] wxEvtHandler::ProcessEvent(wxEvent&)
[13] wxEvtHandler::ProcessPendingEvents()
[14] wxAppConsoleBase::ProcessPendingEvents()
[15] wxApp::DoIdle()
[16] g_main_context_dispatch
[17] g_main_loop_run
[18] gtk_main
[19] wxGUIEventLoop::DoRun()
[20] wxEventLoopBase::Run()
[21] wxAppConsoleBase::MainLoop()
[22] wxEntry(int&, wchar_t**)
[23] __libc_start_main
```
And here's what's displayed when I click on 'Continue':
image
After entering the Old Master Password and clicking on the YubiKey, the YubiKey is no longer recognized.
image
Closing the dialog and reopening it, or closing the database and opening another, fixes this, so it's likely a problem with the logic and not a low-level issue.

nobugshere:
Trouble on the Mac.  If the Yubikey is inserted before I start pwsafe, the dialog looks like the screenshot above (as expected.)  However, if I start pwsafe then insert the key, it looks like this: (Note the overlapped buttons to the right of the Master Password box.)  This does not happen on the current master HEAD.
Screenshot 2024-04-20 at 11 22 38 AM

On the plus side, I could not reproduce the assert error.

macOS 14.4.1, Xcode 15.3, wx 3.2.4
pwsafe: rafaelx:YubiControlsLayout commit c4e254c

nobugshere:
I also think it would be better to keep the Yubi related button and text on the same row. 
I wonder if the "New..." button should be on the Password Database selection row?

rafaelx:
Thanks for the feedback so far.

I added the missing handler for the second timer (second YubiKey button) to the password change dialog.
In addition, the restriction that the dialog cannot grow in height has been removed.

Further changes to the password change dialog will be taken into account in #1226.

ronys:
There's a problem with the last commit: Now when I open the Change Master Password dialog with the Yubikey inserted, the dialog box is displayed but doesn't respond to any input: focus to edit field, button click, here's what I see:

image
Only option is to "Force Quit".

I'm guessing some kind of deadlock. Without YubiKey all works fine.

rafaelx:
@ronys Could you please test it again?

I assume that YubiKey processing/communication needs some time. While the YubiKey request started by the first timer (for the first YubiKey button) is still being processed, another status query is carried out in parallel by a second timer (for the second YubiKey button), which may be the reason for the behavior described.

rafaelx:
Thank you again for taking the time to test this and for providing the information.

For me, the backtrace shows the regular call chain from the timer event handler down to access to the YubiKey.
In the original implementation, a timer was created in the constructor of SafeCombinationChangeDlg and was started at twice the poll interval. According to the comment, the polling interval was increased by a factor of two due to the presence of two YubiKey buttons.
In the event handler of the timer SafeCombinationChangeDlg::OnPollingTimer, HandlePollingTimer of the first YubiMixin was called and then HandlePollingTimer of the second YubiMixin.
With the last commit I reverted to this original implementation, assuming this is the last working version. The only difference is that the polling interval was not doubled.
I wonder if the original version worked.
Without a YubiKey I can't reproduce the problem and can only modify the code based on guesswork.

ronys:
I've just checked, and the change master password dialog works fine with YubiKey in 1.18.2.

I'll look into this once I get the new Windows release process started, hopefully in time to make this part of the corresponding wx release.

ronys:
Is this PR still relevant?

rafaelx:
> Is this PR still relevant?

Yes, it should fix the layout issue if a YubiKey is inserted while a dialog requesting for password is shown.

ronys:
For the record, changing the Yubikey polling interval from 0.5 to 0.9 seconds resolved the "not responding" problem.

rafaelx:
> For the record, changing the Yubikey polling interval from 0.5 to 0.9 seconds resolved the "not responding" problem.

This was the only thing I didn't change to revert back to the original implementation. :(

PR #1240, opened 2024-05-01
[wx] Fix status update of policy dialog controls

### LINKED ISSUE #1236: Password Policy  unchecked by At least fields still enabled
Password Safe v1.9-pre from master branch on Ubuntu 23.10.

1. Manage | Password Policy.
2. Edit button on a policy.
3. Uncheck any of the "Use..." checkboxes. At right "At least" field gets greyed out.
4.  Click on OK button and click on OK button again.
5. Repeat steps 1 and 2. "Use..." checkbox is disabled as expected, but at right site  "At least" field in expected not greyed out and also "-" and "+" buttons are enabled. Clicking on e.g. "+" to increase the value then repeating step 4 and steps 1 and 2 and value clicked on "+" is lost.

In my humble opinion if "Use..." checkbox is unchecked "At least" field should get greyed out and only if "Use..." checkbox is enabled then "At least" field should get enabled again.

rafaelx:
Resolves #1236

### DIFF
--- a/src/ui/wxWidgets/PasswordPolicyDlg.cpp
+++ b/src/ui/wxWidgets/PasswordPolicyDlg.cpp
@@ -79,7 +79,6 @@ PasswordPolicyDlg::PasswordPolicyDlg(wxWindow *parent, PWScore &core,
   }
   
 ////@begin PasswordPolicyDlg creation
-  m_DialogType = type;
   SetExtraStyle(wxWS_EX_BLOCK_EVENTS);
   wxDialog::Create( parent, id, caption, pos, size, style );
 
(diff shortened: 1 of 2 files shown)


igor2x:
@rafaelx, I have downloaded the build you have created from Github Artifacts for Ubuntu and installed it.
Tested this fix and looks problem is fixed. Thanks.

rafaelx:
@igor2x Please note that the password generator dialog is also affected.

rafaelx:
@igor2x Thanks for pointing this out. I was so focused on the spinner controls that I lost sight of the main functionality of the password generator dialog.

src/ui/wxWidgets/PasswordPolicyDlg.cpp
```diff
@@ -763,6 +775,8 @@ void PasswordPolicyDlg::OnUseNamedPolicy( wxCommandEvent& event )
     m_pwpUCSpin->Enable(m_pwpUseUpperCtrl->GetValue());
     m_pwpDigSpin->Enable(m_pwpUseDigitsCtrl->GetValue());
     m_pwpSymSpin->Enable(m_pwpSymCtrl->GetValue());
+    m_OwnSymbols->Enable(m_pwpSymCtrl->GetValue());
+    FindWindow(ID_RESET_SYMBOLS)->Enable(m_pwpSymCtrl->GetValue());
```
dskrvk, line 779: Is it possible for `FindWindow` to return null in some cases?

ronys: Good question. Since the window (button) associated with this id is created unconditionally along with the rest of the dialog's control's, I'd say no. Or, more accurately, if FindWindow() won't find this control, we have bigger problems...

PR #1260, opened 2024-05-06
[wx] Rename 'Yes/No' buttons to 'Discard/Cancel' in Password Policy dialog

### LINKED ISSUE #1257: Yes/No buttons in cancel Password Policy
Password Safe v1.19-pre from latest master on Ubuntu 24.04.

1. Manage | Password Policies.
2. Edit button at "Default Policy".
3. Click some "Use.." checkbox.
4. Click on Cancel button and bellow dialog appears.
image

Suggestion:
- Replace text "Are you sure you wish to cancel?" with "Do you want to discard the changes."
- Instead of **Yes** button use "Discard"
- Instead of **No** button use "Cancel"

P.S. Logic should be the same as in currently already fixed "Cancel" button behavior on "Edit Entry" dialog.

rafaelx:
Resolves #1257


dskrvk (approved):

ronys:
Renaming the No button to "Cancel" may look a bit odd.
For example, under Windows, the message box won't have a close button (x) without wxCancel.

PR #1266, opened 2024-05-08
 [i18n] Updates in Russian Help/.po and minor fixes

pm-kan:
Besides l10n updates this PR contains few fixes for issues spotted while checking translations:
- have to extend 'Key' label in MFC forms
- one of `CGeneralMsgBox::MessageBox` cases (currently unused) had missed `break`
- `CAddEdit_Additional::OnInitDialog` had two `InitToolTip` calls and registration for "old-style tooltips" were lost
- warnings about skipped entries in XML import weren't included into report


src/ui/Windows/GeneralMsgBox.cpp
```diff
@@ -267,6 +267,7 @@ INT_PTR CGeneralMsgBox::MessageBox(LPCWSTR lpText, LPCWSTR lpCaption, UINT uiFla
       tuples.push_back(std::make_tuple(IDCANCEL, IDS_CANCEL));
       tuples.push_back(std::make_tuple(IDTRYAGAIN, IDS_TRYAGAIN));
       tuples.push_back(std::make_tuple(IDCONTINUE, IDS_CONTINUE));
+      break;
```
ronys, line 270: Good catch - thanks!

PR #1271, opened 2024-05-19
[wx] Fix group handling

### LINKED ISSUE #1142: wx: When a group is deleted, it's not always removed from view
When using the Tree View, and the option "Save database immediately after any change" is _**not**_ selected, and a non-empty group is deleted, the leaf entries are removed but the group itself remains in the tree view until the save button is clicked, then it is removed.  Empty groups are removed immediately from view.  I'm fairly certain this is just a graphical update issue.

1. Clear the "save database immediately..." checkbox, you may have to restart pwsafe for that to take effect.
2. select Nested Tree view
3. Select a group that contains entries
4. right-click and select Delete Group, the entries are deleted but the group name remains
5. Click the Save button on the toolbar, or File|Save menu, and the deleted group name disappears.

Tested on:
macOS 14.3.1, pwsafe 1.18.2 release (wxWidgets 3.2.2.1) and local builds with wxWidgets 3.2.4
Fedora 39 ARM, local build from master with wxWidgets 3.2.4

### LINKED ISSUE #1233: File | Properties if password is saved to root level it is counted as empty group
This is not super important issue, just small inconsistency.

Password Safe v1.19-pre from master branch on Ubuntu 23.10.

1. I created test database with some entries and then File | Properties.
image

What does: **1 (1 empty)** means? I have no empty groups.

2. If I add another group and "empty" gets increased by 1.
image

3. If I add an entry to empty group. I still have 2 empty groups.
image

Suggestion:
1. Fix proper counting of empty groups.
2. Remove this "empty groups" information. I don't know if this is really relevant.

rafaelx:
Fixes #1142
Fixes #1233

### DIFF
--- a/src/ui/wxWidgets/AddEditPropSheetDlg.cpp
+++ b/src/ui/wxWidgets/AddEditPropSheetDlg.cpp
@@ -1983,9 +1983,17 @@ Command* AddEditPropSheetDlg::NewAddEntryCommand(bool bNewCTime)
   /////////////////////////////////////////////////////////////////////////////
 
   m_Item.SetStatus(CItemData::ES_ADDED);
-    
+
+  auto *commands = MultiCommands::Create(&m_Core);
+  auto itemGroup = m_Item.GetGroup();
+
+  if (m_Core.IsEmptyGroup(itemGroup)) { // The group is no longer empty if a new item is added
+    commands->Add(
+      DBEmptyGroupsCommand::Create(&m_Core, itemGroup, DBEmptyGroupsCommand::EG_DELETE)
+    );
+  }
+
   if(m_Item.IsAlias()) { // If alias is pointing to shortcut base the shortcuts must be removed before converting to alias base
-    auto commands = MultiCommands::Create(&m_Core);
     const CItemData *pbci = m_Core.GetBaseEntry(&m_Item);
     ASSERT(pbci);
     if (pbci && pbci->IsShortcutBase()) {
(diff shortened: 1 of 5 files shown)


rafaelx:
During testing, I noticed that when deleting a base entry that has a shortcut to it in a separate group, not only the shortcut but the entire group is deleted. Same behavior as the Windows version. Personally, I would expect that only the shortcuts would be deleted, leaving it up to the user what to do with the remaining empty group. Currently, such resulting empty group is deleted through an "automatic cleanup mechanism".
On the one hand, this seems a bit user-unfriendly and patronizing, but on the other hand, it avoids empty groups that would remain. In the case of a large database, this may also be an advantage. Just a hint.

Another note to testers of these changes. Please also consider the undo and redo functionality when creating and deleting items (password elements, shortcuts, nested groups and combinations of these).

nobugshere:
Testing on a Mac, this seems to fix #1142, but I still have trouble with #1233.  There seems to be many variations, but a simple sequence to reproduce the issue:
1. create a new db.
2. Add an entry - ent1
3. Add a group - grp1 - Properties now reports Groups: 1 (1 empty)
4. Move ent1 into grp1 - Properties still reports Groups: 1 (1 empty)

macOS 14.4.1, Xcode 15.3, wx 3.2.4, pulled from rafaelx:GH1142

igor2x:
I have downloaded Github/Actions for this PR on Ubuntu 22.04 and tested @nobugshere reported issue and it looks like last fix has solved the issue. I think this PR is ready to be merged.

nobugshere:
I concur.  It looks good on Mac and Fedora.

PR #1283, opened 2024-05-29
[wx] Remove misplaced Yubi status text control

rafaelx:
While testing the last merges I noticed that the Yubi status text appears to the left of the help button on the entry dialog. This is probably code that was forgotten to be removed or result of the merges?!

PwsafeEntryDlg


ronys:
Probably my mistake while merging stuff. Thanks!

PR #1287, opened 2024-06-01
Mark copied password as secret for Linux

### LINKED ISSUE #1276: Copied passwords not marked as secrets on Linux
Hi, I'm using Kubuntu 24.04, Password Safe 1.18.2. I'm seeing that copied passwords are saved by the built-in clipboard manager. Password Safe is also not able to clear the passwords from the clipboard manager. 

This doesn't happen with KeePassXC, where copied passwords don't show up in the clipboard manager at all. Looks like this is what KeePassXC is doing to mark the copied passwords as secrets: 

https://github.com/keepassxreboot/keepassxc/blob/2a62000b28c7db843a95519a20f3f34ab9e1a45c/src/gui/Clipboard.cpp#L60

hxu1:
Fixes #1276 

image

### DIFF
--- a/src/ui/wxWidgets/Clipboard.cpp
+++ b/src/ui/wxWidgets/Clipboard.cpp
@@ -47,16 +47,16 @@
 #include "wxUtilities.h"
 
 #if defined(__UNIX__) && !defined(__WXOSX__)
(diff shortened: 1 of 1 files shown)


ronys:
Thanks!

PR #1294, opened 2024-06-02
Allow a blank password when using a Yubikey for a new DB

nobugshere:
When creating a new DB, the "Set Master Password" dialog does not allow a blank password with a Yubikey.  The unlock dialogs and the "Change Master Password" dialog all support using a Yubikey without a password.  This PR fixes the "Set Master Password" to allow this when creating a new DB.

(The big change to the Xcode project file is only because I sorted the file names in the UI.)


ronys:
Nice catch - thanks!

PR #1299, opened 2024-06-03
[wx] Modernization of the Change Password Dialog

### LINKED ISSUE #1226: [wx] Modernization of Change Password Dialog

rafaelx:
PasswordChangeDialog
PasswordChangeDialogYubi

Closes #1226


nobugshere:
A quick look on the Mac revealed a small regression.  In the current master, the "Master Password Entry" dialog appears with some empty space that is filled-in if a yubikey is inserted.  In a build from rafaelx:GH1226, the dialog initially appears with the extra space then, about a second later, the window shrinks.  (I speculate the delay is related to the Yubikey timeout.)  If a Yubikey is inserted, it re-expands and displays the button and text as normal.  From a user perspective, a dialog with some extra empty space is preferable to one that suddenly shrinks after it appears.  

Unfortunately (from a Mac user perspective :-)), this problem does *not* happen on my Fedora VM.  The window appears without the blank space, then expands when a key is inserted.

Fedora 40 ARM64, Xfce, gcc 14.1.1, wx 3.2.4, GTK 3.24.42

nobugshere:
When changing a PW using a Yubikey, the dialog opens with a minimum size.  Then, as each Yubikey message is displayed, the window expands horizontally to accommodate the longer messages, two expansions.  (See the sequence of screenshots below.)  Personally, I would prefer to have the window start large enough to accommodate the longest line rather than have it grow incrementally while using it.  I guess that applies to my previous post, as well; A little bit of empty space that may or may not be used later is preferable to repeated resizing.  (Maybe @igor2x could chime in here.)
Screenshot 2024-06-03 at 9 12 45 PM
Screenshot 2024-06-03 at 9 13 00 PM
Screenshot 2024-06-03 at 9 13 10 PM

igor2x:
"A little bit of empty space that may or may not be used later is preferable to repeated resizing."

The question is how much is "little", specially because we need to understand that our application is multi-languaged and line can be much larger in some non-English language.

I think dialog size should remain the same size during whole change password process. Also location of buttons or field should never change. Only the text should change and it can be broken into two lines:

Enter new master password (if any) and
click on bottom YubiKey button.

Above two lines also makes it more clear that two actions are required.

ronys:
I agree with your comments re sizing and spacing, but am a bit uncomfortable with the phrasing of "Enter the new master password (if any) and ..."
- It's not clear what's meant by "(if any)". I think "(optional)" is better.
- Although we support no password with Yubikey, I'm not sure this is the right way to expose this capability. I'm more comfortable leaving this as a note in the documentation.

igor2x:
@ronys, I agree "(if any)" wording is confusing and is candidate for simplification.
I see it this way:
- remove "(if any)" from text
- create new issue to add requirement for mandatory password (this should be in separate PR)

--- 

Also... there is second part of text that is little bit unusual: "click on top/bottom Yubikey button".
Like I see it when dialog is opened both Yubikey buttons are greyed out and text "Insert Yubikey" is displayed.
Now user inserts Yubikey. Like I understand (don't have Yubukey to test) now we enable both Yubikey buttons, isn't it. Why? Why don't we enable only only top Yubikey button and leave bottom Yubikey button greyed out. In this case text can be: "... and click on Yubikey button." Then when user clicks on top Yubikey, then this button gets greyed out and bottom Yubikey button is un-greyed and we have the same part of the message bellow: "... and click on Yubikey button". The logic is, it is obvious from interface which Yubikey to click, because the only one that is enabled.

----

Another option, but this looks more like a wizard approach. Using only single Yubikey button.
1. "Insert Yubikey" and Yubikey button is disabled.
2. "Enter old master password and click on Yubikey button" and Yubikey button is enabled. You know from the logic what logic should follow in source code when first time Yubikey button is clicked.
3. "Enter new master password and click on Yubikey button" and Yubikey button is also enabled, but this time when clicked logic for new master password is applied.

image

ronys:
With Yubikey, the change master password dialog needs to support 4 combinations:
1. No Yubikey in old, no yubikey in new (this is the same functionality as when there's no Yubikey detected)
2. From an old master password without Yubikey to a new master password with the Yubikey, that is, adding Yubikey protection to an existing database - in this case, the user must only click on the lower Yubikey button
3. From an old master password with Yubikey to a new one without, that is, removing Yubikey protection. Here the user need only click on the upper buttone.
4. From and old master password to a new one, both with Yubikey. Here both buttons must be activated.

I can't think of a simpler way to support the above functionality.

igor2x:
Sorry, it looks like I did it again. Start discussion in PR and making PR process way longer then necessary. In the future I must restrain on commenting Yubikey, because I am too much of the guessing... and my proposed solutions are probably bad one.

The last question before my Yubikey "smart" ideas was how to handle dialog resizing. I think this should be finalized and then PR merged. If possible make dialog in fixed size and break the text in two lines instead.

PR #1304, opened 2024-06-10
Fix logic issues in the Change Master Password dialog

### LINKED ISSUE #1230: wx: Change Master Password dialog: Yubikey vs password logic problems
I found a couple of problems in the "Change Master Password" dialog. If the safe only uses a Yubikey (no password):

1. After activating the key as usual, it grays-out the second YK button, expecting you to insert a different key, so you cannot add a PW and keep the same YK. You can accomplish this, but it takes two passes. First, change it to PW only, then a second invocation to add the YK back.
2. If I want to remove the YK and just use a PW, I can fill-in the new and confirmation boxes, but when I click OK, it says "The combination cannot be blank." I have to put something in the Old Master Password field; anything at all works.

Tested on macOS 14.4.1, wx 3.2.4.  Likely affects Linux as well.
v1.19.0-pre and 1.18.2

nobugshere:
Resolves #1230

Corrects some behavior when using a Yubikey:
- Allows the old password to be blank
- Requires that the new and confirmation fields match
- After validating the old password/Yubikey combo, advance the focus to the New PW field

Tested on:
Fedora 40 ARM64, Xfce/X11, gcc 14.1.1, wx 3.2.4, GTK 3.24.42

### DIFF
--- a/docs/ReleaseNotesWX.md
+++ b/docs/ReleaseNotesWX.md
@@ -7,6 +7,20 @@ https://pwsafe.org/. Details about changes to older releases may be found in the
 
 In the following, SFxxxx refers to Bug Reports, Feature Requests and Service Requests in PasswordSafe SourceForge Project tickets, and GHxxxx refers to issues in the PasswordSafe GitHub project.
 
+PasswordSafe 1.20 Release ??? 2024
+=====================================
+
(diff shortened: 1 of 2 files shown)


ronys:
Thanks!

PR #1306, opened 2024-06-11
[wx] Use the password strength check with a Yubikey

### LINKED ISSUE #1300: [wx] Using strong passwords with a Yubikey
Related to the discussion in #1299 

@ronys, 
If one *does* enter a PW with a Yubikey, do you think it should go through the same strength test(s), as for a PW only entry, and the user alerted if it doesn't pass?  (Currently, that is not the case.)  I think the main reason for using a PW with a Yubikey is in case someone steals your key.  So, if you set one, it should be a good one.  But, it is optional anyway, so I keep going back and forth on this :-)

This is in reference to the WX version; I haven't check the behavior of the Windows version.

nobugshere:
Resolves #1300 for Mac and Linux.  Similar changes are still needed for Windows.

- Move some duplicated code to a new function: CheckPasswordStrengthAndWarn() in wxUtilities.cpp
- Add calls to the "Set Master Password" and the "Change Master Password" dialogs; with and without a Yubikey

### DIFF
--- a/src/ui/wxWidgets/wxUtilities.cpp
+++ b/src/ui/wxWidgets/wxUtilities.cpp
@@ -28,6 +28,7 @@
 #include <wx/versioninfo.h>
 
 #include "core/PWScore.h"
+#include "core/PWCharPool.h" // for CheckMasterPassword()
 #include "PWSafeApp.h"
 #include "SafeCombinationCtrl.h"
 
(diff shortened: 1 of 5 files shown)


src/ui/wxWidgets/wxUtilities.cpp
```diff
@@ -186,6 +187,36 @@ void UpdatePasswordTextCtrl(wxSizer *sizer, wxTextCtrl* &textCtrl, const wxStrin
...
+{
+  // Vox populi vox dei - folks want the ability to use a weak
+  // passphrase, best we can do is warn them...
+  // If someone want to build a version that insists on proper
+  // passphrases, then just define the preprocessor macro
+  // PWS_FORCE_STRONG_PASSPHRASE in the build properties/Makefile
+  StringX errmess;
+  if (!CPasswordCharPool::CheckMasterPassword(password, errmess)) {
+    wxString cs_msg;
+    cs_msg = errmess.c_str();
```
dskrvk, line 200: Can we combine the last two lines?

src/ui/wxWidgets/wxUtilities.h
```diff
@@ -201,6 +201,15 @@ inline const wxChar* ToStr(bool b) {
  */
 void UpdatePasswordTextCtrl(wxSizer *sizer, wxTextCtrl* &textCtrl, const wxString text, wxTextCtrl* before, const int style);
 
+/**
+ * Checi if the password passes the strength rules, prompt the user if it does not.
```
dskrvk, line 205:
```suggestion
 * Check whether the password passes the strength rules, prompt the user if it does not.
```

nobugshere: Thank you!

dskrvk (approved):

Issue #1288, opened 2024-06-01
[wx] Create new password database can be resized beyond minimal dialog size


igor2x:
I have downloaded latest Password Safe  from master branch and installed on Ubuntu 22.04. Dialog looks like this:
![image](https://github.com/pwsafe/pwsafe/assets/10895030/b0a29be1-53ba-4956-8e1a-33d94431443f)

Dialog can be resized beyond minimal, so buttons gets hidden and interesting last sentence from the text above fields gets also hidden like this:
![image](https://github.com/pwsafe/pwsafe/assets/10895030/adc06bf1-6849-44db-b05f-989270cb66d9)

I retested on v1.18.2/flatpak and dialog can't be resized bellow minimal. This issue is regression in v1.19-pre.

rafaelx:
I can't reproduce this under Ubuntu 23.10 (wx version 3.2.2 ; Wayland) and Kubuntu 24.04 (wx version 3.2.2 ; X11).

rafaelx:
Does this actually affect all password entry dialogs with a row of buttons at the bottom?
Which wxWidgets version is used on Ubuntu 22.04 for the Deb package and the Flatpack version?
(The version is printed on the command line when the application is started from there and the About dialog is called.)

igor2x:
**1. xwWidgets version**
[Flatpak version](https://github.com/flathub/org.pwsafe.pwsafe/blob/master/org.pwsafe.pwsafe.yml) is using xwWidgets v3.2.4.

Help | About:
![image](https://github.com/pwsafe/pwsafe/assets/10895030/63939355-24ac-4b25-9565-109dbd901fec)


**2. Dialogs affected**

Dialogs that can go beyond minimal (problem hiding buttons):
- First dialog where Master Password is entered
- File New dialog
- right click on entry and Delete button

Not affected (works fine)
- right click on entry and Edit
- Manage | Options
- Manage | Change Master Password

Issue #1247, opened 2024-05-04
Unlock dialog can be resized too much to hide buttons


igor2x:
Password Safe from master branch on Ubuntu 24.04.

Similar to [issue 1245](https://github.com/pwsafe/pwsafe/issues/1245) Unlock database dialog can be resized too much so that buttons are hidden.

![image](https://github.com/pwsafe/pwsafe/assets/10895030/1d0c5fb5-64cc-4267-bb70-8e9051f6cfaa)

P.S. Why is so much vertical empty space between Master Password input box and Cancel button? Is it some kind of placeholder for Yubikey? If yes, then it is OK, if not, then probably would be nice to reduce the spacing.

nobugshere:
Hi @igor2x ,
Is this the way the window appeared on its own, or are you saying that manual resizing allowed the buttons to be hidden?

Just a data point, I cannot resize the dialog box smaller than its original size.
Fedora 39 ARM, wx 3.2.4, GTK 3.24.41, Xfce

igor2x:
This: manual resizing allowed the buttons to be hidden.

rafaelx:
This should be fixed by #1309. Can you confirm this?

igor2x:
Tested: Password Safe v1.19.1 flatpak. Works fine. Dialog can't be resized too much to hide buttons. Fixed. Thanks.

Issue #1245, opened 2024-05-04
Main Password Safe dialog can get resizes to minimal dialog too much to hide buttons


igor2x:
I have [installed](https://github.com/pwsafe/pwsafe/actions/runs/8950267357) Password Safe latest version from [rafaelx's PR 1244 ](https://github.com/pwsafe/pwsafe/pull/1244)on Ubuntu 24.04.

I see dialog is improved a lot. The only issue is that dialog can get resized above minimal. In my humble opinion dialog should not get resized so much that buttons are hidden.

![image](https://github.com/pwsafe/pwsafe/assets/10895030/4091c38f-ed51-429f-acab-e3910cc526d2)

P.S. I have tested several Password Safe versions including latest master and all have this issue, that is why I am reporting  new issue instead of commenting in any existing PR.

rafaelx:
Should be fixed by #1232.

I'm surprised that the "Browse" and "New" button are so wide in the screenshots. In my case, the buttons look smaller, as shown in https://github.com/pwsafe/pwsafe/issues/1241#issuecomment-2091831307

rafaelx:
This should be fixed by #1309. Can you confirm this?

igor2x:
Tested: Password Safe v1.19.1 flatpak. Dialog can't be resized too much to hide dialogs. Fixed. Thanks.

PR #1309, opened 2024-06-12
[wx] Reserve space for Yubi controls to prevent dialog resizing

rafaelx:
Related to #1288, #1245, #1247

### DIFF
--- a/src/ui/wxWidgets/SafeCombinationChangeDlg.cpp
+++ b/src/ui/wxWidgets/SafeCombinationChangeDlg.cpp
@@ -175,12 +175,18 @@ void SafeCombinationChangeDlg::CreateControls()
 #endif
 
 #ifndef NO_YUBI
-  m_yubiStatusCtrl = new wxStaticText(this, ID_YUBISTATUS, _("Insert your YubiKey"), wxDefaultPosition, wxDefaultSize, 0);
-  mainSizer->Add(m_yubiStatusCtrl, 0, wxEXPAND|wxALL, 12);
+  auto* panel = new wxPanel(this, wxID_ANY, wxDefaultPosition, wxSize(-1,  35));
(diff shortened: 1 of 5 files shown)


nobugshere:
A *very minor* cosmetic nit-pick: On my Fedora VM, the text gets close to the edge of the window.  

Just for reference, on a newly installed Ubuntu 24.04 VM, the last word wraps and on Mac there is a little space.  Perhaps it's the result of slight differences in font size and shape. 

Fedora 40 ARM64, Xfce/X11, gcc 14.1.1, wx 3.2.4, GTK 3.24.42
Fedora Screenshot_2024-06-14_10-12-12

Ubuntu 24.04 ARM64, Gnome/Wayland, gcc 13.2.0, wx 3.2.4, GTK 3.24.41
Ubuntu Screenshot from 2024-06-14 10-23-12

Mac Screenshot 2024-06-14 at 10 14 19 AM

nobugshere:
That last commit doesn't seem to have made any visible difference at all.

rafaelx:
Thanks for testing on different platforms and sharing the results.

It seems that Wx does not respect the spacing requirement for the text element when performing line breaks. Therefore, it is possible that the text comes very close to the edge of the dialog, as shown in the following figure.

PasswordChangeWordWrap

I have increased the height of the available space for the status text so that there is enough space for two lines. The spacing is no longer based on the dialog width, but is limited by the additional panel (visualized by a black border in the screenshot below) on which the status text is placed.

PasswordChangeWordWrapBox

I hope that this will now show the same behavior across all platforms.

nobugshere:
It does look the same on all three platforms now.  The word "button" is initially wrapped to the second line.  When I resize the window, the text reflows as it should.  When shrinking the window, the last word wraps with the right amount of space between it and the edge.  This all looks good to me.  Thanks!

PR #1311, opened 2024-06-15
[wx] Clear old lock on open

nobugshere:
When a file is already open, and the user tries to open a different file via "File->Open..." or the Open button, the already open file is saved and closed, but its lock file is not removed.

### DIFF
--- a/src/ui/wxWidgets/PasswordSafeFrame.cpp
+++ b/src/ui/wxWidgets/PasswordSafeFrame.cpp
@@ -1389,10 +1389,8 @@ int PasswordSafeFrame::Open(const wxString &fname)
   if (rc != PWScore::SUCCESS)
     return rc;
 
-  // Clear the lock on the old file, if any.
-  // This has to be done after the old file is saved (see SaveIfChanged() above)
-  // and before a call to SetCurFile() (see below)
-  m_core.SafeUnlockCurFile();
+  // Save the current file name so we can unlock it later.
+  stringT oldfn = GetCurrentFile().c_str();
 
   // prompt for password, try to Load.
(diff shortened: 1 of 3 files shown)


src/ui/wxWidgets/PasswordSafeFrame.cpp
```diff
@@ -1389,6 +1389,9 @@ int PasswordSafeFrame::Open(const wxString &fname)
   if (rc != PWScore::SUCCESS)
     return rc;
 
+  // Clear the lock on the old file, if any.
+  m_core.SafeUnlockCurFile();
```
dskrvk, line 1393: I’m thinking maybe we should wait until after the password check - maybe put this around line 1444 (ClearAppData())?

nobugshere: SafeUnlockCurFile() calls GetCurFile() to get the name of the lock file, so it has to be called before SetCurFile() at line 1402 (setting the name of the new file to be opened.)  I'll add a comment to that effect.  The areas you mention, ClearAppData() and line 1444, are within a '#if 0' block.  I don't know the history of that code, but checking the PW seems to be done elsewhere.

dskrvk: What about the return condition on line 1412 - if we get there, the password didn't match, and seems like the current file should remain open and locked?

nobugshere: I see what you mean now, you're right.  I added a new version of the SafeUnlockFile function so I could save the file name and pass it later.  Please, have another look.  Thanks for checking this.

dskrvk (approved):

Issue #1315, opened 2024-06-26
Redesign Edit Entry dialogs


igor2x:
The first rule is: [Minimize the number of fields.](https://designlab.com/blog/form-ui-design-best-practices)

Suggestion:
1. Remove text above that is describing what fields should be filled. This is currently needed, because UI is not clear what is required to be filled in.
2. Remove Group field, that is actually pretty confusing to me. I have already clicked on one of the folders (or in Password Safe language "group") and now why are you asking me for the same info I have already provided in previous step?
3. On Basic tab, the name itself says it should only be basic info there. I would split current "Basic" to two tabs: "Basic" and "Details". Then on Basic tab only put the most essential info. I actually never need anything else then Tile, Username and Password (also Confirm).
4. Move all other less important setting to new tab with name Details.
5. At Username field mark this field is optional, to follow rule No. 2 from above link.
6. I suggest to rename Additional to Action. Because that actually it is, there are info for autotype and similar settings.
7. I currently do not have comment on "Alias To" button, because I don't know how fill this work. Is this "alias" functionality from Windows, that do not work on Linux? I think this "Alias To" belongs more to the "Username" field from user point of view. I think "Alias To" is not basic setting and should probably be put somewhere on "Details" tab. Also I think we are somehow abusing "Password" field for multiple information like "Alias". Don't know why? Are we trying to be compatible with current file format or something. I think Alias should not be saved in "Password" field.

![image](https://github.com/pwsafe/pwsafe/assets/10895030/46a7febc-965e-463e-8138-905756e73dd9)

I think we are doing it again... :) We are trying to do whole remake on one single simple issue. I am afraid current issue will have long discussion again.

rafaelx:
2. When using the tree view, the group membership may be obvious, but in the list view, the entries are not organized into groups. In the list view, the group membership of an entry cannot be changed simply by dragging and dropping, so the dropdown box for displaying and changing the group name is quite useful.

rafaelx:
Another version to get an idea of ​​what the dialog looks like without the descriptive text at the top, when some elements are simply rearranged, with input fields marked as optional and the new resulting size.

![image](https://github.com/pwsafe/pwsafe/assets/4042917/8cb5f3aa-4a48-45b4-8b15-5360f98d7934)

A new tab for some of the optional information seems appropriate here. Personally, I already like this version. :)

nobugshere:
[...]

Issue #1238, opened 2024-04-26
Add matching icon for Password and Confirmation


igor2x:
Password Safe v1.19-pre from master branch on Ubuntu 24.04.

Add new entry:
![image](https://github.com/pwsafe/pwsafe/assets/10895030/24b5607c-c54e-406d-bc62-d568a8343258)

Problem is I never know if Password and Confirmation match. I need to click Show or click OK button.

It would be nice if some green check box is displayed just like in installation process of installing Ubuntu. When passwords match right of both input fields green check icon is displayed.
![image](https://github.com/pwsafe/pwsafe/assets/10895030/8b9db57e-597f-45ba-a4ff-bb90dfb11afc)

But there should be no icon at all until passwords matches, it is just annoying to have icon that tells you the obvious, that passwords do not match.

rafaelx:
The screenshot shows a first suggestion of the dialog with password confirmation icons. In addition, the "Alias" button has been explicitly displayed so that the positioning of all possible controls is visible. The existing "Show" button has been replaced by an icon so that it takes up less space.

![AddEditConfirmationIcon](https://github.com/pwsafe/pwsafe/assets/4042917/2cbdbcc3-52f1-41fd-84ae-e6ceff545131)

If we follow the recent changes to the other dialogs, the labels could be above the text input fields and each password input field could have its own eye icon, which would be located directly to the right of the input field. The confirmation icons would then be positioned to the right of the eye icon. Hm, or should the order of the elements here be: input field, confirmation icon, eye icon?

rafaelx:
If the labels and input fields are aligned vertically, the dialog grows accordingly in height and no longer appears compact.

![AddEditConfirmationIcon2](https://github.com/pwsafe/pwsafe/assets/4042917/7ba2d0a2-b974-431e-ab86-ab3ec52e6b15)

igor2x:
@rafaelx, this looks good and is very good starting point.

"Compact" problem is, because[ we don't follow ](https://designlab.com/blog/form-ui-design-best-practices) UI design rules.

The first rule is: "Minimize the number of fields". I always though there are way too many field in Add Entry.
Suggestion:
1. Remove text above that is describing what fields should be filled. This is currently needed, because UI is not clear what is required to be filled in.
2. Remove Group field, that is actually pretty confusing to me. I have already clicked on one of the folders (or in Password Safe language "group") and now why are you asking me for the same info I have already provided in previous step?
[...]

PR #1333, opened 2024-07-08
[wx] Redesign of Basic tab of Add/Edit Entry Dialog

rafaelx:


rafaelx:
Closes #1315 
Resolves #1238

nobugshere (changes requested):
First look on the Mac, I notice a redundant horizontal scroll bar on the policy tab that the other tabs don't have.  I think you can remove it.

rafaelx:
@nobugshere Thanks for the feedback. Following your suggestion, I removed the horizontal scroll bar (which never appeared in my case) and increased the height of the notes field a bit.

nobugshere:
Thanks!  I haven't noticed anything else.

nobugshere (approved):

src/ui/wxWidgets/wxUtilities.cpp
```diff
@@ -269,6 +274,31 @@ wxBitmap wxUtilities::GetBitmapResource( const wxString& name )
     wxBitmap bitmap(cpane_xpm);
     return bitmap;
   }
+  else if (name == wxT("graphics/checkmark_placeholder.xpm"))
```
dskrvk, line 277: Can we use a switch statement instead?

src/ui/wxWidgets/wxUtilities.cpp
```diff
@@ -269,6 +274,31 @@ wxBitmap wxUtilities::GetBitmapResource( const wxString& name )
...
+  {
+    wxBitmap bitmap(checkmark_placeholder_xpm);
+    return bitmap;
```
dskrvk, line 280: Nit: we could simplify this to
```suggestion
    return wxBitmap(checkmark_placeholder_xpm);
```

src/ui/wxWidgets/AddEditPropSheetDlg.cpp
```diff
@@ -1707,19 +1744,57 @@ void AddEditPropSheetDlg::ShowPassword()
...
+
+  // There is nothing to do if there is no user input, but the content of
+  // the password input fields may have been changed by the hide/show functionality.
+  if (!m_BasicPasswordTextCtrl->IsModified() && !m_BasicPasswordConfirmationTextCtrl->IsModified()) {
+    return;
+  }
+  // If both passwords entered are the same, the green checkmark icons will appear to indicate the match.
+  if (m_BasicPasswordTextCtrl->GetValue() == m_BasicPasswordConfirmationTextCtrl->GetValue()) {
+    m_BasicPasswordBitmap->SetBitmap(wxUtilities::GetBitmapResource(wxT("graphics/checkmark_green.xpm")));
+    m_BasicPasswordConfirmationBitmap->SetBitmap(wxUtilities::GetBitmapResource(wxT("graphics/checkmark_green.xpm")));
```
dskrvk, line 1778: Not sure how expensive the `GetBitmapResource` call is, but in any case could we load the resource once, assign the result to a variable and use it for both bitmaps?

src/ui/wxWidgets/wxUtilities.cpp
rafaelx: I don't think C++ supports this for strings.

src/ui/wxWidgets/wxUtilities.cpp
rafaelx: Changed as proposed.

src/ui/wxWidgets/AddEditPropSheetDlg.cpp
rafaelx: Changed as proposed.

dskrvk (approved):

src/ui/wxWidgets/wxUtilities.cpp
dskrvk: That's a bummer. It works with enums though, so another option would be to change the `GetBitmapResource`'s argument to be an enum. That would also provide a nice compile-time check against unsupported resource names. Up to you.

rafaelx: While working on the redesign of various dialogs, I noticed that there is a method called GetBitmapResource in the source code of every dialog implementation. This expects a path to an image file as a string and returns an image object. When the eye icon was introduced, this method had to be extended in several dialogs to take the path of the eye icon into account. These dialogs also had to get the same include statement for the eye icon. To avoid this redundancy, GetBitmapResource was introduced in wxUtilities. This function still follows the old concept that was introduced by a UI designer for the individual dialogs in order to keep the changes to a minimum. I fully agree that this could now be solved better in terms of programming. However, I would suggest considering this as a separate task, as a concept that is common to all dialogs and image resources should be introduced. This PR should keep its focus on the redesign of the Add-Edit-Entry dialog.

ronys:
Hi,

A couple of minor tweaks and I think we'll be ready to merge:

1. On my Debian 12, the Password and Confirm edit controls are too narrow (see below).
2. Rather than put "(Optional)" on most fields, I'd rather use the convention of "*" for the two mandatory fields, that is "Title*" and "Password*". I'm also fine with no indication, as (a) most users will fill these anyway, and (b) the error messages are pretty clear.
3. I've never had any complaint about "URL" being unclear. Changing it to "Web address" here means it will need to be changed in all menus, and makes the autotype mnemonic ('\l') even less related to the value name. Please change this back to 'URL'
image

rafaelx:
> Hi,
> 
> A couple of minor tweaks and I think we'll be ready to merge:
> 
>     1. On my Debian 12, the Password and Confirm edit controls are too narrow (see below).
> 
>     2. Rather than put "(Optional)" on most fields, I'd rather use the convention of "_" for the two mandatory fields, that is "Title_" and "Password*". I'm also fine with no indication, as (a) most users will fill these anyway, and (b) the error messages are pretty clear.
> 
>     3. I've never had any complaint about "URL" being unclear. Changing it to "Web address" here means it will need to be changed in all menus, and makes the autotype mnemonic ('\l') even less related to the value name. Please change this back to 'URL'

The last commits are related to 2 and 3. The first issue still needs some attention.

EntryDialogBasicTab

ronys:
Much better, I think.
Only nit I have is with the Confirm field. It's required only if the password is hidden, so marking it as required when the password's shown is potentially confusing.
We can either add logic to show/hide the asterisk in Confirm depending on whether the password is hidden or shown, or remove it entirely from the Confirm label.

igor2x:
The idea of redesign dialog entry is to be more clear, simpler to use etc.
1. Using asterisks is in my humble opinion not the best design. Why? You can see it from dialog itself, it is though to require explanation at the top of dialog what is the purpose of asterisks, which somehow disqualifies the whole purpose of redesign, to be simplistic. The _Optional_ tag at the optional fields is [UX design step 2](https://designlab.com/blog/form-ui-design-best-practices) as _Clearly mark optional fields_. OK, somehow I am also fine with asterisks, because it is so common usage. But I would suggest to remove comment above _All fields marked with an asterisk (*) are required._ If we think that this comment is required, then we are clearly doing something wrong.
2. I also suggested to replace _URL_ with _Web site_, because URL is nerd acronym and the UX guideline is to remove nerd wordings. If nobody complained about URL, this does not mean much... What people don't understand, they don't use. I also seen wrong translations, because clearly translator did not understand what this URL really means.
3. I would also suggested until Title, Password and Confirm fields are filled in to grey out OK button, just like [step 30 in UX guidelines.](https://designlab.com/blog/form-ui-design-best-practices) No need user gets error when not filling the fields and pressing OK button.

I leave final words to @ronys, I just though I need to add a note, because those were my ideas and I think I need do add some explanation for the reasons behind the changes.

rafaelx:
Both variants, marking optional input fields as well as mandatory fields, correspond to good UI design. Basically, there is no right or wrong, you just have to determine what should be used or what seems more sensible. My favorite is marking with an asterisk. Why? On the Basic tab there are 8 potential input fields, of which only 3 are mandatory. In fact, only 2 are mandatory if the text is displayed in the password input field. Instead of marking the majority of input fields as optional, I find it less intrusive if only the few mandatory fields are marked with an asterisk. A short explanatory sentence about the meaning of the asterisk (which most users are sure to be familiar with) does not conflict with the goal of simplification for me. Especially when I think of the original version of the dialog, where there was significantly more text to read in the upper area. The goal of simplification has thus still been achieved. The user should be able to quickly grasp and understand the meaning of the individual visible input elements.

@ronys Maybe it would be good to "git squash" some of the recent commits, since the commit "Show asterisk in confirmation field only in Add mode" was simply wrong. I only noticed later that you didn't mean the Add mode, but that if the password is visible in plain text in the input field, the asterisk can be omitted. I somehow remembered it wrong when implementing it.

igor2x:
> On the Basic tab there are 8 potential input fields, of which only 3 are mandatory [...] I find it less intrusive if only the few mandatory fields are marked with an asterisk.

This now makes sense... But my original proposal in first post of [issue 1315](https://github.com/pwsafe/pwsafe/issues/1315) was to split current _Basic_ tab to two tabs. First tab with bare essential options and second tab with optional selections. In that case _Optional_ make more sense, because _Optional_ fields are in minority. This idea was to follow UX guidelines step 1: [Minimize the number of fields](https://designlab.com/blog/form-ui-design-best-practices). But now I agree since this idea (two tabs instead of one) was dismissed, then also _Optional_ idea gets dismissed. This is domino effect.

_Especially when I think of the original version of the dialog, where there was significantly more text to read in the upper area_

In my humble opinion, if we are required to explain what UI does, then we are not doing it the best way. Having long text especially proves this. Long text belong to Help not in UI. From the UI it should be immediately obvious what user should do.

ronys:
@rafaelx:
1. The asterisk next to "Confirm" works fine when the display icon is toggled, but is incorrectly shown if the display option "Show Password in Add & Edit" is set.
2. On my Debian 12 VM, commit [85884db](https://github.com/pwsafe/pwsafe/pull/1333/commits/85884db39451743a16bef1683e2441221fd356ca) has no visible effect. The Password and Confirm Edit controls are still too narrow:
image
(I've double-checked that the commit is applied in the code I've compiled)
Note that resizing the dialog affects the Password and Confirm fields. Perhaps this provides a clue as to the problem?
image

3. I'm fine with squashing commits in this branch. Just use '--force' when pushing the result.

@igor2x 
1. The asterisk convention is very common. So much so that I'm OK with removing the "All fields..." explanatory text from the top. Alternately, we can move it to the bottom, which is where the text associated with asterisks usually goes
2. As to URL/Web Address: How about "Web Address (URL)"? This would help us keep "URL" in the other places.

rafaelx:
@ronys I have tested it on different Linux distributions (VM) and do not observe the behavior you describe.

| Linux Distribution | Window Manager | wxWidgets Version | GTK Version | Wayland or X11 |
| ------------------- | -------------------- | --------------------- | ------------- | ----------------- |
| Debian 11            | Cinnamon              | wx 3.0.5                   | GTK 3.24.20 | X11                    |
| Debian 12            | Gnome                   | wx 3.2.2                   | GTK 3.24.36 | Wayland            |
| Debian 12            | KDE Plasma            | wx 3.2.2                   | GTK 3.24.36 | Wayland            |
| Fedora 40            | Gnome                    | wx 3.2.4                   | GTK 3.24.41 | Wayland            |
| Kubuntu 23.10     | KDE Plasma            | wx 3.2.2                   | GTK 3.24.38 | X11                    |
| Manjaro 23.0.0    | KDE Plasma             | wx 3.2.2                   | GTK 3.24.36 | X11                    |
| Ubuntu 23.10       | Gnome                   | wx 3.2.2                   | GTK 3.24.38 | Wayland            |

Do you perhaps have any further information that might be helpful?

ronys:
Commit https://github.com/pwsafe/pwsafe/commit/b64a63ad665ff54facec245778a010a15853579d (in my fork) fixes the problem on my Debian 12 VM.
Can you check it on your environments?

nobugshere:
@ronys 
The code originally at line 323 sets m_BasicPasswordBitmap, without that it's left as a nullptr.  That pointer is dereferenced as soon as you type something in the password field (in AddEditPropSheetDlg::UpdatePasswordConfirmationIcons).

ronys:
Thanks @nobugshere . Fixed in commit 23b47d2be
Now I get the following ugliness:
image
Putting the lower checkmark in the same column as the top one gives me an assert about overlapping controls. Any suggestions welcome.

rafaelx:
@ronys Thanks for the feedback. I don't know why I can't reproduce what you described, but inspired by your changes, here is another attempt to improve the layout behavior. Could you be so kind and test this again?

ronys:
The position of the controls is perfect - thanks!
There's still some inconsistency with the visibility of the checkmarks when the password is shown/hidden, but this PR has been open way too long already. I'll open another ticket describing the inconsistencies in detail.

PR #1335, opened 2024-07-09
File path display

### LINKED ISSUE #1331: If password is incorrectly typed in database file name path is hidden
Password Safe v1.19.2 on Ubuntu 24.04.

1. Launch a program and observe the Password Database field. At the end database name is displayed. Because path is too large to be displayed in current input box path is shortened with "..." dots. Works as expected.
image
2. Now intentionally type wrong master password and click on OK button when error is displayed. Problem appears. I expect the path is displayed in exactly the same way as in step 2, so I can see database file name, but it is not. Database file name is hidden and whole directory path is displayed instead. When wrong password is typed in it is even more important to see database file name, because when using multiple password databases, it is nice to get file name without resizing whole window.
image

### LINKED ISSUE #1332: At Unlock dialog database file name is hidden because of too short dialog
Password Safe v1.19.2 on Ubuntu 24.04.

1. Lock the database and then Unlock it. Small problem appears similar to [issue 1331](https://github.com/pwsafe/pwsafe/issues/1331). Database file name is not visible. I expect if database path is too large to fit the input box, then path should have "..." dots and database file name at the end displayed.
image

It is tinny inconvenience, but if having multiple databases, then I am now required to resize the dialog to see actual database file name.

rafaelx:
Resolves #1331.
Resolves #1332.

### DIFF
--- a/src/ui/wxWidgets/SafeCombinationEntryDlg.cpp
+++ b/src/ui/wxWidgets/SafeCombinationEntryDlg.cpp
@@ -630,10 +630,13 @@ void SafeCombinationEntryDlg::OnYubibtnClick(wxCommandEvent& WXUNUSED(event))
     bool oldYubiChallenge = ::wxGetKeyState(WXK_SHIFT); // for pre-0.94 databases
     if (PerformChallengeResponse(this, m_password, response, oldYubiChallenge)) {
       m_password = response;
-      ProcessPhrase();
-      UpdateStatus();
+      if (ProcessPhrase()) {
+        EndModal(wxID_OK);
+      }
     }
   }
+  EllipsizeFilePathname();
+  UpdateStatus();
 }
 
 void SafeCombinationEntryDlg::OnPollingTimer(wxTimerEvent &evt)
@@ -653,7 +656,7 @@ void SafeCombinationEntryDlg::OnDBSelectionChange(wxCommandEvent& WXUNUSED(event
 
 void SafeCombinationEntryDlg::UpdateReadOnlyCheckbox()
 {
-  wxFileName fn(m_filenameCB->GetValue());
+  wxFileName fn(m_filename);
 
   // Do nothing if the file doesn't exist
   if ( fn.FileExists() ) {
--- a/src/ui/wxWidgets/SafeCombinationEntryDlg.cpp
+++ b/src/ui/wxWidgets/SafeCombinationEntryDlg.cpp
@@ -289,8 +289,8 @@ void SafeCombinationEntryDlg::CreateControls()
   m_filenameCB->Append(recentFiles);
   // The underlying native combobox widget might not yet be ready
   //  to hand back the string we just added
-  wxCommandEvent cmdEvent(wxEVT_COMBOBOX, m_filenameCB->GetId());
-  GetEventHandler()->AddPendingEvent(cmdEvent);
+//  wxCommandEvent cmdEvent(wxEVT_COMBOBOX, m_filenameCB->GetId());
+//  GetEventHandler()->AddPendingEvent(cmdEvent);
   SetIcons(wxGetApp().GetAppIcons());
 }
 
--- a/src/ui/wxWidgets/SafeCombinationEntryDlg.cpp
+++ b/src/ui/wxWidgets/SafeCombinationEntryDlg.cpp
@@ -384,17 +384,14 @@ void SafeCombinationEntryDlg::OnOk( wxCommandEvent& )
       wxMessageDialog err(this, _("File or path not found."),
                           _("Error"), wxOK | wxICON_EXCLAMATION);
       err.ShowModal();
-      EllipsizeFilePathname();
       m_filenameCB->SetFocus();
       return;
     }
     if (ProcessPhrase()) {
       EndModal(wxID_OK);
     }
-    else {
-      EllipsizeFilePathname();
-    }
-  } // Validate && TransferDataFromWindow
+  } // TransferDataFromWindow
+  EllipsizeFilePathname();
 }
 
 bool SafeCombinationEntryDlg::ProcessPhrase()
--- a/src/ui/wxWidgets/SafeCombinationEntryDlg.cpp
+++ b/src/ui/wxWidgets/SafeCombinationEntryDlg.cpp
@@ -244,6 +244,7 @@ void SafeCombinationEntryDlg::CreateControls()
   m_filenameCB->Bind(wxEVT_KILL_FOCUS, [&](wxFocusEvent& WXUNUSED(event)) {
     m_filename = m_filenameCB->GetValue(); // The user may have changed the file name or path manually.
     EllipsizeFilePathname();
+    UpdateReadOnlyCheckbox();
   });
 
   // Event handler to update the file path name string if the size of the combobox changed.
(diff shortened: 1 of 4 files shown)


nobugshere (changes requested):
@rafaelx ,
Your last commit is causing a serious regression (at least on the Mac, I haven't tried Linux yet.)  Since you no longer save and restore m_filename, it's getting set to the "elipsised" version.  Among the problems this causes: I can't open a file unless I make the window wide enough for the name without ellipses.  Could this be because of the SetValidator @SafeCombinationEntryDlg:233 ?

Also, whatever you wind-up doing in OnOk() will most likely be needed in OnYubibtnClick() as well.

rafaelx:
For me it works without any problems under Kubuntu. When you move the mouse over the combobox, does the tooltip show the full path or the shortened path? (The tooltip is always updated based on m_filename.)

I have placed a trace function in each event handler and in the function SafeCombinationEntryDlg::EllipsizeFilePathname to be able to observe the member m_filename and the contents of the combobox. m_filename always points to the full path. Only the combobox shows the shortened path. In OnOk the combobox is set to the full path before further activities are executed.

nobugshere:
There's definitely some magic that happens in WX that I don't fully understand yet.  But, I what I think is happening is that the calls to ChangeValue() are generating wxEVT_TEXT events, even though the WX documentation says it doesn't!  This leads to the GetValue() call at line 272 (the wxEVT_TEXT event handler) after the shortened name has been set.

Focusing on SafeCombinationEntryDlg.cpp, here are some of the observations that lead me to that conclusion:
1. I can't reproduce the problem on a Fedora/Xfce VM, only the native Mac build.
2. On initial launch, if the file I want is selected by default, I can open it regardless of how it's displayed in the combobox.
3. On initial launch, if I simply click the text part of the filename box, then click back to the password box, I cannot open the file if the name was shortened.  (File not found.)
4. If I make the window wide enough that ellipses are not needed, I can always open the file.
5. I discovered that the combobox tooltip never did work on macOS.  Apparently, wxEVT_MOTION is not being delivered, so the tooltip is not being set at line ~240.
6. If I place the call to SetTooltip() at the end of EllipsizeFilePathname(), the tooltip works, **_but displays the ellipsized name_**.
7. Setting breakpoints, m_filename is correct on entry to EllipsizeFilePathname() and on call to Ellipsize().  
8. After return from ChangeValue(), before anything else, m_filename has been changed to the shortened version.  This happens even if I comment-out the validator at line 233.
9. I breakpointed the event handlers and clicked back and forth between boxes, I noticed the wxEVT_TEXT events each time ChangeValue() is called.
10. Finally, I can comment-out the wxEVT_TEXT event handler and things work better.

Do your traces show any evidence of wxEVT_TEXT following a call to ChangeValue()?  I assume not.

Conclusions:
1. Possible macOS specific bug in wxWidgets causing wxEVT_TEXT events after a call to ChangeValue(), contrary to the documentation.
2. Since wxEVT_TEXT doesn't work right and wxEVT_MOTION doesn't work at all, we can't rely on those events to manage the tooltip or m_filename.  
3. I think by setting the tooltip in EllipsizeFilePathname(), we can eliminate those two event handlers and rework the others so that GetValue() is never called after EllipsizeFilePathname().  However, I haven't actually thought through the various set_focus, kill_focus, combobox, scenarios yet.

rafaelx:
I don't see a wxEVT_TEXT event in my trace after ChangeValue was called. The event only occurs when the text in the combobox is edited. So I see a behavior that conforms to the documentation. It does indeed look like a bug in WX.

If it is helpful for further investigation, here is the trace function I used, which I placed in the header file.
```
void Trace(const wxString& name) {
  pws_os::Trace(L"[TRACE] Function: '%ls' / m_filename: '%ls' / Combobox: '%ls'", name.wc_str(), m_filename.wc_str(), (m_filenameCB->GetValue()).wc_str());
}
```

Similar to the previous version, the full path could be stored in a temporary variable to restore the path in m_filename after calling ChangeValue.
```
void SafeCombinationEntryDlg::EllipsizeFilePathname()
{
  if (m_filename.IsEmpty()) {
    return;
  }

  wxScreenDC dc;
#ifdef __WXOSX__
  wxString filename(m_filename);
#endif
  m_filenameCB->ChangeValue(
    wxControl::Ellipsize(
      m_filename, dc, wxEllipsizeMode::wxELLIPSIZE_MIDDLE,
      /* The limiting width for the text is the combobox width reduced by the drop-down button width and margins. */
      (m_filenameCB->GetSize()).GetWidth() - 50
    )
  );
#ifdef __WXOSX__
  m_filename = filename;
#endif
}
```
Alternatively, the event handler could be deactivated (via Unbind) before calling ChangeValue and then activated (via Bind) again afterwards.

nobugshere:
@rafaelx, Thanks for the trace and tips.  I don't want to leave you hanging, so here's an update.  I've been working on my own update to this branch which, IMHO, simplifies the code and overall flow by being less reliant on event processing.  Along the way, I found a couple more issues, such as EVT_COMBOBOX being defined twice, once in the event table and again via a call to bind.  (Only the latter was being executed.)  As a result, UpdateReadOnlyCheckbox() was never called. 

I got SafeCombinationEntryDlg working well on the Mac...but broke Linux.  Eventually, I realized the new Linux problem only happened on Fedora/Xfce, not on Fedora/KDE or Ubuntu (or Mac).  I now have it working on all the platforms I've tested.  I still need to work on SafeCombinationPromptDlg.  After that, and some more testing, I'll post it for your review.

rafaelx:
@nobugshere Sorry for the late response and thanks for the improvements.
@ronys This PR is ready for the final review and merge.

src/ui/wxWidgets/SafeCombinationEntryDlg.cpp
```diff
@@ -366,13 +346,19 @@ wxIcon SafeCombinationEntryDlg::GetIconResource( const wxString& WXUNUSED(name)
 void SafeCombinationEntryDlg::OnOk( wxCommandEvent& )
 {
   // For the validation process, put the full file path name back into the combo box.
+  // Calling 'EllipsizeFilePathname' will undo this.
   m_filenameCB->ChangeValue(m_filename);
 
-  if (Validate() && TransferDataFromWindow()) {
```
dskrvk, line 371: If we keep the logic this way, we can reduce the number of places EllipsizeFilePathname is called, i.e.
```
if (Validate() && TransferDataFromWindow()) {
...
}
EllipsizeFilePathname();
```

src/ui/wxWidgets/SafeCombinationPromptDlg.cpp
```diff
@@ -236,20 +273,35 @@ void SafeCombinationPromptDlg::ProcessPhrase()
 
 void SafeCombinationPromptDlg::OnOkClick(wxCommandEvent& WXUNUSED(evt))
 {
-  if (Validate() && TransferDataFromWindow()) {
```
dskrvk, line 239: Wanted to make the same comment as in `SafeCombinationEntry`, except that here there's no call to `EllipsizeFilePathname()` in the `!TransferDataFromWindow()` case - is that intentional?

dskrvk:
Seems like there are many places where we have to call Ellipsize, and it may be hard to avoid regressions as the code evolves (many mutually-interacting conditions to keep track of, no tests). Would it perhaps be easier to make "ellipsized" the default state and only expand the text in select circumstances where this is needed?

PR #1342, opened 2024-07-15
Fr1301: Support autotype of TOTP auth code

### LINKED ISSUE #1301: FR - perform autotype for auth code
Current (version 3.66.1) way to type the generated auth code is to copy it to clipboard and then paste it to right window. Or then just manually typing the number the window that needs the code.

I'd prefer to have separate autotype functionality for _auth code only_ from the context menu. This way there wouldn't be need to copy-paste the number.

The functionality could be implemented as more general one too; One could have multiple autotype sequences defined per password entry. In addition there should then be a separate code for autotype auth code ( `\a` ?). Once user has defined the different sequences they would be shown as entries in an autotype submenu of the context menu. This would make autotyping much more powerful and flexible to use.

ronys:
Resolves #1301

### DIFF
--- a/Makefile.windows
+++ b/Makefile.windows
@@ -57,8 +57,8 @@ I18N_DLL_DIR := ./out/build/windows-x86-release/I18N
 I18N_DLL_DIR_64 := ./out/build/windows-x64-release/I18N
 POT_FILE := $(I18N_DIR)/pwsafe.pot
 HELP_DIR := ./help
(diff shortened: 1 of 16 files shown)


dskrvk:
I would also suggest combining some of the commits - e.g. commit 2 won't compile without commit 4.

ronys:
@dskrvk - thanks for the review.
As to combining commits, I prefer avoiding this on stuff that's already been pushed.

Issue #1343, opened 2024-07-17
Check for the latest version is redudant in Help | About for flatpak


igor2x:
Password Safe v1.19.1 on Ubuntu 24.04.

1. Help | About.
2. Click on "Check for the latest version" and nothing happens.

![image](https://github.com/user-attachments/assets/24c9d4ad-5e9c-43ce-9472-a8c0df8e1454)

Password Safe flatpak by default does not have access to the network (setting in manifest file) and so this can only work if user manually enables access to network e.g. using flatpak command or some GUI like Flatseal.

But... I think that Check button in flatpak does not have any value because:
- we release Password Safe native (.deb and .rpm) few days/weeks before flatpak  - and this "Check for the latest version" is for native package
- flatpak is not updated manually by downloading some installation file (like .deb or .rpm), but is automatically updated from Flathub.org.

In my humble opinion, if Password Safe is installed as flatpak, then in Help | About link to check for new version is redundant and should be removed.

ronys:
Hmm.
A couple of points:
1. The check button works fine if '--share=network' is passed to the flatpak run command. This is also needed if the display is remote. Without this I get:
F: Remote X11 display detected.
F: X11 access will require --share=network permission.
2. The "Password Safe website" link is also broken, as well as the Help menu entry. Even with  '--share=network' set, I get 
(pwsafe:2): GLib-GIO-WARNING **: 08:14:59.041: Cannot create document portal proxy: Error calling StartServiceByName for org.freedesktop.portal.Desktop: Timeout was reached

Therefore I suggest the following:
1. Instead of removing the Check button, the error message should be changed to something like "--share=network permission required for this check".
2. If the visit website feature can't be fixed (I suspect we need to add a browser to the manifest, which seems like overkill to me), then we should replace the link with plain text in the Help->About menu, and remove the Help->Visit Password Safe website entry in the Help menu.

As for the redundancy of the check, I disagree. It's easier to keep it with the changes I suggested than have suspicious users asking why it was removed.

PR #1344, opened 2024-07-18
Add a FLATPAK define symbol

nobugshere:
Partial fix for #1343.  This will not take effect until -DFLATPAK is added to the Flatpak manifest file.
- Adds Cmake logic to append -DFLATPAK to the compiler flags
- Adds #ifndef to remove the version check link if FLATPAK is defined

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -51,6 +51,7 @@ include(create-harden-interface)
 
 
 # Configurable options:
+option (FLATPAK "Set ON to build for Flatpak" OFF)
 option (NO_YUBI "Set ON to disable YubiKey support" OFF)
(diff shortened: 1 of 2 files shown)


ronys:
In addition to my comment in the ticket (#1343), I'd much rather this be a runtime check for the FLATPAK_ID env variable than a compilation flag.

ronys (changes requested):
In addition to my comment in the ticket (https://github.com/pwsafe/pwsafe/issues/1343), I'd much rather this be a runtime check for the FLATPAK_ID env variable than a compilation flag.

nobugshere:
Withdrawn.

PR #1348, opened 2024-07-25
[wx] Rename menu entry from 'Recent Safes' to 'Recent Databases'

### LINKED ISSUE #1346: Inconsistent naming Database vs. Safes
Password Safe v1.19.1 on Ubuntu 24.04:
1. Manage | Options | System and there is option _Recent Databases on File Menu rather then as a sub-menu_
2. Close down Password Safe and reopen (Note: No info in program this is required).
3. Menu File and there is option _Recent Safes_.
4. Compare wording from step 1 (Database) with wording in step 3 (Safes). If I remember correctly we have already defined somewhere that "safes" is outdated and should be replaced with "database".

image

rafaelx:
Closes #1346


ronys:
Thanks!

Issue #1220, opened 2024-04-10
Recurring option is selectable when parent option is not selected


igor2x:
Password Safe v1.19-pre master branch on Ubuntu 23.10

1. Click on entry and right click menu Edit Entry.
2. Dates and Times tab.
3. In Password Expiry, click on On or Never selection.
4. Now in row which starts with "In" click on checkbox Recurring. If "On" or "Never" is selected then Recurring checkbox should be greyed out.

nobugshere:
It turns out the code is there to enable and disable the controls depending on the selected radio button.  But, it just doesn't work.  I suspect it's a problem in wxWidgets when the controls are contained within a panel.

nobugshere:
Hi @igor2x,
I think this has been resolved by PR #1353

igor2x:
I forgot to test this on v1.19.x, but I have now tested on v1.20 and this is fixed.
Thanks for fixing it.

PR #1353, opened 2024-08-04
[wx] Fix expiration controls

### LINKED ISSUE #935: Recurring checkbox default is updating existing entries
I'm using a local build pulled on 2/9/2023.  Now, whenever I view an existing entry and then click cancel, without making any changes, I get the pop-up that says something has changed.  I think it's related to this commit from 1/15:
[[FR890] "Recurring" checkbox is now set by default](https://github.com/pwsafe/pwsafe/commit/3e4c685bbc2f0bb0d19bb9a81df2588a8bca7484)

It's not only setting the default for a new entry, it's trying to modify the existing entries just because I look at them.  One way I confirmed this:
- Open an old existing entry that is set to Never expire
- Go to the date time tab and confirm the Recurring check box is now checked (it was not on by default when the entry was created)
- Click cancel, when I get the "are you sure" prompt click yes
- Open the same entry again and clear the Recurring check box
- click cancel - observe no confirmation pop-up

Environment:
Mac Studio M1 Max
MacOS 13.2 Ventura
Xcode 14.2
wxWidgets 3.2.2 (installed via homebrew)
PasswordSafe 1.16 pulled from git master on 2023-02-09

### LINKED ISSUE #1149: Password Expirity is not remembered

1. Right mouse and Add Entry.
2. Fill in Title, Password and Confirm.
3. Click on Dates and Time tab.
4. Click at "In" option and leave 90 days and click checkbox Recurring.
5. Click OK button to save the entry.
6. Right click on entry.
7. Click on Date and Time tab. Problem appears Password Expiry is set to Never. Because in step 4 I have set "In" option, I expect "In" to be set.

nobugshere:
Resolves #935
Resolves #1149
Reinstates FR890 (Recurring checkbox on by default when interval is selected)

This is a major re-work of the password expiration controls.  It took me a while to work out all the corner cases so, please, give it a good shake down.
- Fixed numerous bugs in the control management and cases of false or missed change detection.
- Fixed multiple issues when adding a new entry
- Use the default interval set in Options | Password History tab.
- Recurring checkbox is set by default when the interval radio button is selected
- Truncate Date/Time values to use only the date, this fixed a number of comparison problems
    (The Date picker control handles time differently on OSX vs. GTK)
- Added “(Expired)” or “(Expires in %d days)” to the “Original Value:” text
- Added a description of the records to the formatV3.txt document

Tested on:
macOS M1Max Sonoma 14.6,   Xcode 15.4, wx 3.2.4 and 3.2.5 (Development and most testing)
Ubuntu 24.04 ARM64 VMware, Gnome/Wayland, gcc 13.2.0, wx 3.2.4, GTK 3.24.41
Fedora 40 ARM64 VMware,    Xfce/X11,      gcc 14.1.1, wx 3.2.4, GTK 3.24.43
Fedora 40 ARM64 VMware,    KDE/Wayland,   gcc 14.1.1, wx 3.2.4, GTK 3.24.43
Fedora 38 ARM64 VMware,    Xfce/X11,      gcc 13.2.1, wx 3.0.5, GTK 3.24.41

The only layout change is to add “(Expired)” or “(Expires in %d days)” to the Original Value text:
Screenshot markup


src/ui/wxWidgets/AddEditPropSheetDlg.h
```diff
@@ -388,10 +392,12 @@ class AddEditPropSheetDlg : public wxPropertySheetDialog
...
   wxString m_AccessTime;
   wxString m_CreationTime;
-  wxString m_CurrentExpirationTime;
+  wxString m_OriginalExpirationDate;
   wxString m_ModificationTime;
+  wxRadioButton *m_OriginalButton;
   bool m_Recurring;
-  wxString m_ExpirationTime;
+  bool m_OriginalRecurring;
+  time_t m_OriginalDayttt;
```
dskrvk, line 400: Not sure what the third "t" stands for, but for consistency with the other time_t member:
```suggestion
  time_t m_tttOriginalDay;
```

src/ui/wxWidgets/AddEditPropSheetDlg.cpp
```diff
@@ -2656,16 +2668,40 @@ void AddEditPropSheetDlg::OnExpRadiobuttonSelected( wxCommandEvent& evt )
   bool Never = (evt.GetEventObject() == m_DatesTimesNeverExpireCtrl);
```
dskrvk, line 2668: Nit: redundant parentheses on these two lines

src/ui/wxWidgets/AddEditPropSheetDlg.cpp
```diff
@@ -2656,16 +2668,40 @@ void AddEditPropSheetDlg::OnExpRadiobuttonSelected( wxCommandEvent& evt )
...
-    TransferDataToWindow();
+
+  } else {  // Interval
+    wxDateTime xdt(wxDateTime::Today());  // Today returns time part 0
+    xdt += wxDateSpan(0, 0, 0, m_DatesTimesExpiryTimeCtrl->GetValue());
+    m_DatesTimesExpiryDateCtrl->SetValue(xdt);
+    m_tttExpirationTime = xdt.GetTicks();
+    m_Recurring = true;
   }
+  // In add mode, the item data is not prepared for this, fields will go blank.
```
dskrvk, line 2702: This seems like a tight coupling between this event handler and the control, in that we're making assumptions about another control's behavior. Instead, could we still send the data, but implement proper handling logic in the control? I don't know how we're populating the field there, so this may not make sense.

src/ui/wxWidgets/AddEditPropSheetDlg.cpp
```diff
@@ -2656,16 +2668,40 @@ void AddEditPropSheetDlg::OnExpRadiobuttonSelected( wxCommandEvent& evt )
   bool Never = (evt.GetEventObject() == m_DatesTimesNeverExpireCtrl);
 
   if (Never) {
-    m_ExpirationTime = _("Never");
-    m_CurrentExpirationTime.Clear();
-    m_tttExpirationTime = time_t(0);
-    m_ExpirationTimeInterval = 90;
-    wxDateTime xdt(wxDateTime::Now());
+    //m_tttExpirationTime = time_t(0);
```
dskrvk, line 2671: Remove this line?

src/ui/wxWidgets/AddEditPropSheetDlg.cpp
```diff
@@ -2656,16 +2668,40 @@ void AddEditPropSheetDlg::OnExpRadiobuttonSelected( wxCommandEvent& evt )
...
+    // If it has an interval, keep it.  Otherwise, use the user defined default.
+    // This applies to new entries and old entries with no interval setting.
+    //
+    // Some entries, created with recent versions of pwsafe, will have an interval but no date,
+    // which is interpreted as "Never".  But, we need to use the given interval to prevent
+    // a false change detection when the entry is opened and closed.  If an actual change is made,
+    // the entry will be re-written correctly.
+    if (!m_ExpirationTimeInterval) { // Compatibility
```
dskrvk, line 2679: To confirm, there's no "never" setting anymore - if the user hasn't specified any expiration settings, we'll always use the default interval - is that right?
E.g. if the user selected "never", why do we set the expiry date? Wouldn't that cause the entry to be marked expired on that date, contrary to what the user wanted?

And for additional context, is the problem basically that there was a bug which caused a nonsensical combination of settings to get stored (interval but no date), so we're doing a one-time migration to bring the stored settings to a valid state?

src/ui/wxWidgets/AddEditPropSheetDlg.cpp
```diff
@@ -2656,16 +2668,40 @@ void AddEditPropSheetDlg::OnExpRadiobuttonSelected( wxCommandEvent& evt )
...
+      m_ExpirationTimeInterval = PWSprefs::GetInstance()->GetPref(PWSprefs::DefaultExpiryDays);;
+    }
+    // Needed here for initialization and the above compatibility problem.
+    wxDateTime xdt(wxDateTime::Today());  // Today returns time part == 0
     xdt += wxDateSpan(0, 0, 0, m_ExpirationTimeInterval);
     m_DatesTimesExpiryDateCtrl->SetValue(xdt);
+    m_tttExpirationTime = xdt.GetTicks();
+    m_Recurring = (m_Type == SheetType::ADD) ? true : false;
```
dskrvk, line 2687:
```suggestion
    m_Recurring = m_Type == SheetType::ADD;
```

src/ui/wxWidgets/AddEditPropSheetDlg.h
nobugshere: Well...I'm going for consistency with the other m_Original's :)

src/ui/wxWidgets/AddEditPropSheetDlg.cpp
nobugshere: Removed.

src/ui/wxWidgets/AddEditPropSheetDlg.cpp
nobugshere: Removed.

src/ui/wxWidgets/AddEditPropSheetDlg.cpp
nobugshere: Uh...I must have been tired :)  Fixed.

src/ui/wxWidgets/AddEditPropSheetDlg.cpp
nobugshere: > To confirm, there's no "never" setting anymore - if the user hasn't specified any expiration settings, we'll always use the default interval - is that right?

There never was a "never" parameter written to the file, it's the absence of the date and interval fields.  Upon initialization of the dialog, my intention is to pre-populate the controls with the default interval and the corresponding date in case the user wants to use them, but the "Never" radio button will be selected.

> E.g. if the user selected "never", why do we set the expiry date? Wouldn't that cause the entry to be marked expired on that date, contrary to what the user wanted?

The control values won't matter if the "never" radio button is selected, neither field is written to the file.  Also, Never is still the default for a new entry.

> And for additional context, is the problem basically that there was a bug which caused a nonsensical combination of settings to get stored (interval but no date)

Exactly, the current version is doing that when Never is selected, so there could be lots of files with that condition.  

> so we're doing a one-time migration to bring the stored settings to a valid state?

No, I didn't want to make mysterious changes to the file.  I accept that condition as Never. (As does the current release.) But, if the user makes an expiry change, it will be written correctly.  One important goal to me is that the user can open the edit dialog and cancel without any changes and not have a false change detection pop-up or write changes to the file.
I did use that interval field to pre-populate the control.  I'm starting to re-consider that, though.  Maybe I should use the default value and completely ignore that field.

Any additional thoughts are welcome.

src/ui/wxWidgets/AddEditPropSheetDlg.cpp
nobugshere: I think the problem is that the function that initializes the dialog controls (it was called UpdateExpTimes, I changed it to InitializeExpTimes because it's only called once) finishes by calling OnExpRadiobuttonSelected() to complete the initialization.  It worked okay when editing an existing entry, but caused problems when filling-in a new entry.  I thought it was strange to call an event handler like that, but I didn't change it, I just tried to make it work.  

I'm going to re-think that decision and see if I can refactor that code for a cleaner separation.

nobugshere:
I removed the connection between the initialization routine and the radio button handler.  I also changed the special case to ignore the improperly written interval and use the user-set default instead.  (The entry won't be re-written unless the user makes a change.)   I think this version much cleaner, though there are still a lot of possible variations, especially changing from one mode to another.  I tried to handle everything I could think of :)  Please, have another look.

src/ui/wxWidgets/AddEditPropSheetDlg.cpp
```diff
@@ -1347,52 +1343,95 @@ static struct {short pv; wxString name;}
...
+  m_OriginalDayttt = 0;
+  m_OriginalExpirationDate = m_Item.GetXTimeL().c_str();
 
-  if (m_ExpirationTime.empty())
-    m_ExpirationTime = m_CurrentExpirationTime = _("Never");
+  // Special case: Some entries, created with recent versions of pwsafe, might have
+  // an interval but no date, which is interpreted as "Never".  We are going to ignore
+  // it and use the user-set default.  If an expiry change is made, the entry will
+  // be re-written correctly.
+  int defInterval = PWSprefs::GetInstance()->GetPref(PWSprefs::DefaultExpiryDays);
```
dskrvk, line 1359:
```suggestion
  int defaultInterval = PWSprefs::GetInstance()->GetPref(PWSprefs::DefaultExpiryDays);
```

src/ui/wxWidgets/AddEditPropSheetDlg.h
```diff
@@ -288,8 +288,14 @@ class AddEditPropSheetDlg : public wxPropertySheetDialog
 
   void ItemFieldsToPropSheet();
   void SetupDCAComboBoxes(wxComboBox *pcbox, short &iDCA, bool isShift);
-  void UpdateExpTimes();        // entry -> controls
+  void InitializeExpTimes();        // entry -> controls
   void SetXTime(wxObject *src); // sync controls + controls -> entry
+
+  // Today returns time part == 0
+  wxDateTime TodayPlusInterval() { return wxDateTime::Today().Add(wxDateSpan(0, 0, 0, m_ExpirationTimeInterval)); };
```
dskrvk, line 295: Stylistic preference, but for simple functions like this it's better IMO to not have side effects and instead take all inputs as parameters. It's just not obvious from the prototype what interval we're adding. Also, the function would be more reusable with a parameter for interval.

src/ui/wxWidgets/AddEditPropSheetDlg.cpp
```diff
@@ -1347,52 +1343,95 @@ static struct {short pv; wxString name;}
...
 
-  wxCommandEvent dummy;
   wxDateTime exp;
```
dskrvk, line 1361:
```suggestion
  wxDateTime expiryDate;
```
(I had to scroll back to this line a few times to remind myself what this variable contains)

src/ui/wxWidgets/AddEditPropSheetDlg.cpp
```diff
@@ -1347,52 +1343,95 @@ static struct {short pv; wxString name;}
...
+  // results in different values when converting to or from time_t and
+  // false or missed change detections.
+  // GTK seems to remove the time part, macOS preserves it.
+  // Since we only care about the date for expiration, let's just
+  // remove the time wherever we need the date.
+  // Note the wxWidgets documentation says Today() returns the
+  // time part set to 0, and Today() and Now() both use the local time zone.
+  m_DatesTimesExpiryDateCtrl->SetValue(exp);
+  m_tttExpirationTime = exp.GetTicks();
+  m_Recurring = true;
```
dskrvk, line 1409: Why is this set to true unconditionally?

src/ui/wxWidgets/AddEditPropSheetDlg.cpp
```diff
@@ -1347,52 +1343,95 @@ static struct {short pv; wxString name;}
...
+
+  if (m_OriginalDayttt) {
+    wxString rstr;
+    int interval = IntervalFromDate(wxDateTime(m_OriginalDayttt));
```
dskrvk, line 1413:
```suggestion
    int interval = IntervalFromDate(exp);
```
or is it crucial that we include the time part here?

nobugshere:
While working on the third round, I realized m_tttExpirationTime is always set to match the value of the m_DatesTimesExpiryDateCtrl date picker control, so it's not needed.

src/ui/wxWidgets/AddEditPropSheetDlg.cpp
nobugshere: Done

src/ui/wxWidgets/AddEditPropSheetDlg.h
nobugshere: Good point.  Changed.

src/ui/wxWidgets/AddEditPropSheetDlg.cpp
nobugshere: Done

src/ui/wxWidgets/AddEditPropSheetDlg.cpp
nobugshere: This is building a string to represent the number of days until expiration, so it needs the original date.  At this point, exp (expiryDate) may have changed.

src/ui/wxWidgets/AddEditPropSheetDlg.cpp
nobugshere: This is just setting the default state of the recurring checkbox when the dialog is initialized.  All three of the conditions in the preceding if/else blocks would set it to true, so I did it in one place to highlight it as a common default.   I've added a comment.  (Note: this is consistent with the Windows version.)

ronys:
I really appreciate the work you put in here, as the code is (very) tricky to get right. In fact, going back to the Windows version to compare behavior, I still see problems there...

A couple of issues that seem worth addressing:

1. Create a new entry, specify an expiration date. The "Never" button is still selected even though I've specified a date. So is the date in effect, or do I also have to explicitly click on the "On" radio button?
2. Likewise if I sent an interval.
3. Even more confusing: I set a date, then changed the interval, radio button left on "Never". Close and re-open entry, "On" is selected.

Also: setting a new entry by date, the interval track the date, which is nice, but closing and then opening the entry, the interval reverts to the default, which is confusing.

Finally, I think the text for original value when an interval's selected is awkward, e.g., "08/23/2024 (expires in 12 days) (every 12 days)". Perhaps the following's a bit clearer:
- When the date's specified: "08/23/2024"
- When an interval's specified: "12 day from last change (expiring on 08/23/2024)"
- When a recurring interval's specified: "Every 12 days from last change (next expiration on 08/23/2024)"

nobugshere:
Hi @ronys,
Thanks for the feedback.  I hope it's getting close :)

First of all, do you agree with the way I described the feature in the format docs?  Much of it was inferred.  I need to get that clear in my head first :)

- 1 and 2
My intention is for the radio buttons to determine which mode is being set.  Which, in turn, determines which other controls are meaningful.  The Windows version actually enables and disables the controls depending on which radio button is chosen. There is code in the WX version to do the same thing, but for some reason, it isn't working.  Issue #1220 is related.  I suspect it's a problem in wxWidgets, but so far, my attempt to make a simplified test case has failed to reproduce the problem.

- 3
I cannot reproduce this scenario on Mac, Ubuntu or Fedora.  Can you provide more info on the platform you're using?  And, was this a new entry or editing an existing entry?

> Also: setting a new entry by date, the interval track the date, which is nice, but closing and then opening the entry, the interval reverts to the default, which is confusing.

I went back and forth on this a few times; I don't remember why I left it this way.  I can change it to initialize the interval field based on the number of days until the specified date, if that makes more sense.  In that case, I would only use the default for a new entry or an existing entry that is currently set to Never.

> Finally, I think the text for original value when an interval's selected is awkward...

The static text "Original value" seems awkward too.  Would something like "Current Setting" be better?

This implementation doesn't record a case of an interval that is not recurring.  That's just recorded as a specific date.  So, how about something like this?  Perhaps, rendering expired messages in red?

- Specific date:
Expires in X days (\<date\>) -or- Expired on (\<date\>)
- Recurring interval:
Every 12 days from last change (Next expiration on \<date\>) -or- 
Every 12 days from last change (Expired on \<date\>)
- Never
Never Expires

ronys:
Hi @nobugshere,

Thanks for pointing to the format description - I was intent on checking the functionality, and missed that change. To the point: I think the only way case 3 (only interval field defined) occurred was if the user set an interval and the disabled it by selecting the "Never" radio button. This may have changed in later versions to just not recording it in this case. The semantics are indeed as you described: the value is ignored.

I'm testing on Debian 12 (bookworm), which uses wxWidgets 3.2.2. Scenario 3 happens both in existing and new entries.

Re dates and interval: The problem is the inconsistent behavior. I like the tracking, that is, setting the interval to the number of remaining days, but if it's easier to leave the default, then that's fine, as long as that's always what's displayed when the date's set.

Expiration texts: "Current Setting" is fine, as well as your proposed texts, except for the redundant parenthesis in "Expired on (<date>)" - should be just "Expired on <date>"

nobugshere:
> I'm testing on Debian 12 (bookworm), which uses wxWidgets 3.2.2. Scenario 3 happens both in existing and new entries.

I'll look into setting-up a Debian 12 ARM VM.  Which desktop environment do you use?

Just to be clear:
1. Open an entry or create a new one
2. Don't touch the radio buttons (Never is selected)
3. Set a date (via text entry or date chooser?)
4. Change the interval (via text entry or spin control?)
5. click OK
6. Re-open the entry and the "on" radio button is selected.

nobugshere:
@ronys 
Made changes as discussed above.  Please, see what you think.
1. When opening an entry with a specific date set, initialize the interval to "days until expiration" instead of the default value.
2. Removed some -W options that are not needed after @huven PR's.
3. I still cannot reproduce the error you reported, not even on Debian 12.6 ARM, wx3.2.2, and Xfce desktop.
4. Re-worked the expiration text.  The text for recurring intervals was too long so I put it on two lines.  I also turn the text red when the password is expired.

Screenshot 2024-08-15 at 11 30 18 AM

ronys:
The attached clip illustrates the problem I'm seeing.
[clip.zip](https://github.com/user-attachments/files/16643936/clip.zip)
It looks like changes to the expiry aren't detected/saved.
Specifically:
- the 'recurring' checkbox is always set
- changing from an absolute date to interval isn't persistent.

Also, I think the relevant radio button should be set when the control to its right gets focus.

Database is also attached, password "Sample123".
[safe.zip](https://github.com/user-attachments/files/16643967/safe.zip)

nobugshere:
@ronys,
I finally figured-out why the controls could not be disabled (grayed-out) corresponding to the selected radio button.  So, now I can offer you a choice:

1. Enable/disable the controls depending on the radio button selected (consistent with the Windows version) (I would still strive to keep the date and interval days in sync, just because.)
2. Continue down the current path where all controls are active and the RB follows focus.

IMHO, No.1 is probably simpler overall with fewer chances of corner cases and, maybe, less confusing.  But, I can go either way.

ronys:
Excellent!
I agree: Consistency with the Windows version + date-interval sync sounds clearest, even if it means an extra mouse click for the user.

nobugshere:
@ronys,
Once more into the breach....

Resolves #1220 

Now the controls are enabled/disabled depending on which radio button is selected.  Commit 5dc399f is the fix for that.

Some things you might notice:
1. When a specific date is set, the interval is initialized to the days until expiration except, if it's already expired, the interval is initialized to the default value.
2. The Recurring checkbox is checked the first time the "In" radio button is selected, whether by initialization or by the user.  After that, it maintains its state until the dialog is closed.

I think this covers the condition in your video by virtu of the fact the controls are grayed-out.

ronys:
Looks good to me.
@dskrvk - any comments?

Issue #1349, opened 2024-07-30
Question: Date time formats to use with CLI commands


cjlotz:
Hello

I'm trying to set the `Created Time` when adding an entry via the CLI command but I'm not able to figure out in what datetime format to provide the value.  I've tried File time UTC and ISO 8601 formats.  The CLI command returns exit code zero, but when I view the entry in the Safe, the Created On is not set

![image](https://github.com/user-attachments/assets/f9ae39e7-bee3-4dc3-86ab-e59eebe13e2d)

What format should I specify the value in?

rafaelx:
I had success with the following command line under Linux.

```
./pwsafe-cli pwsafe.psafe3 --add=Title=TestTitle,Username=TestUsername,Password=TestPassword,"Created Time"="2020/10/02 12:13:14"
```

The keyword "now" can also be used to refer to the current time.

```
./pwsafe-cli pwsafe.psafe3 --add=Title=TestTitle,Username=TestUsername,Password=TestPassword,"Created Time"=now
```

cjlotz:
Thank you so much.  I got it working using your examples.

PR #1356, opened 2024-08-09
[CLI] Add examples to usage output

rafaelx:
With reference to #1349 , notes on how to use some fields and notes on how to specify timestamps have been added to the usage output. In addition, some examples of how to use the fields have been added. I would have liked to have added an example of how to change existing fields of an entry, but unfortunately I was unable to do so.
If further questions are asked about how to use the command line version, the list of notes and examples can be expanded to make it easier for the user to use the command line version.

### DIFF
--- a/src/ui/cli/main.cpp
+++ b/src/ui/cli/main.cpp
@@ -97,7 +97,8 @@ const map<UserArgs::OpType, pws_op> pws_ops = {
 
 static void usage(char *pname)
 {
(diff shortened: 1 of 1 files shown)


ronys:
Nice.
If we add more text to the usage, it might be a good idea to break this out into a separate '--help' command. For example:
pwsafe-cli --help create for help on creating, including an example,
pwsafe-cli --help search for help in searching, including an example,
etc.

PR #1359, opened 2024-08-12
Do not use ptrace API on iOS

huven:
ptrace is not available on iOS. More info on that [here](https://forums.developer.apple.com/forums/thread/701621).

This commit changes nothing for macOS.

### DIFF
--- a/src/os/mac/debug.cpp
+++ b/src/os/mac/debug.cpp
@@ -16,6 +16,9 @@
 #include <syslog.h>
 #include <iostream>
 
+// Provides TARGET_OS_.. defines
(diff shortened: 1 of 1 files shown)


ronys:
What about iPads? Same build? That is, TARGET_OS_IPHONE covers iPad builds?
Excuse my ignorance, I've no idea if the same app/deliverable is compatible for both.

huven:
Please hold this PR, I see an update is needed...

huven:
Just confirmed that this works both in DEBUG and non-DEBUG modes on macOS and iOS. This can be merged.

PR #1360, opened 2024-08-13
Use CoreServices for determining mime type

huven:
This removes the dependency of mac/media.cpp on the wx framework. On iOS the wx framework is not available.

To determine a file's mimetype, the CoreServices framework is used, which is available on both macOS and iOS by default.
This uses the native mimetype database from macOS/iOS, as used by all programs and the OS itself.

### DIFF
--- a/src/os/mac/media.cpp
+++ b/src/os/mac/media.cpp
@@ -2,35 +2,46 @@
 * \file Mac-specific implementation of media.h
 */
 
-#include <wx/filename.h>
-#include <wx/mimetype.h>
+#include <CoreServices/CoreServices.h>
+#include <CoreFoundation/CoreFoundation.h>
 
 #include "../media.h"
-#include "../file.h"
 #include "../utf8conv.h"
-
-#include <cstdlib>
-#include <cstdio>
-
-#include "wxUtilities.h"
 
 using namespace std;
 
-static const char *unknown_type = "unknown";
+static const stringT unknown_type = L"unknown";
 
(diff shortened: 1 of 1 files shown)


ronys:
The unit tests include a minimal test for  pws_os::GetMediaType() that would be a nice sanity check for this PR.
Unfortunately, these are current not built or run in the mac-pwsafe workflow.
Could you add the test execution to the workflow, either as another commit on this PR or as a separate one?

huven:
Hi Rony, agree with you there. Unfortunately I don't have much experience with CMake and GitHub workflows, it looks like there is no testing at all enabled for macOS at the moment, so this needs some global work not related to this particular PR. 

I have tested the 3 tests in TEST(OSTest, testMedia) on my Mac for both iOS and macOS and they return 1, so it looks ok from that perspective. Is that sufficient?

  A77Log(@"%d", _T("unknown") == pws_os::GetMediaType(_T("nosuchfile")));
  A77Log(@"%d", _T("text/plain") == pws_os::GetMediaType(_T("data/text1.txt")));
  A77Log(@"%d", _T("image/jpeg") == pws_os::GetMediaType(_T("data/image1.jpg")));

output:

-[SafeFile initWithFilePath:password:] line 220 14683882 $ 1
-[SafeFile initWithFilePath:password:] line 221 14683882 $ 1
-[SafeFile initWithFilePath:password:] line 222 14683882 $ 1

ronys:
Yes, that's fine.
I'll look into adding the tests to the workflows separately. Thanks.

PR #1361, opened 2024-08-14
Add detection for ARM64 architecture (mobile devices, Apple Silicon, ..)

huven:


ronys:
Odd - Raspberry Pi builds are also for ARM 64 bit, and they work fine without this.

PR #1363, opened 2024-08-14
[CLI] Add example about Updating and Searching to usage output

rafaelx:


dskrvk (approved):
Nice!

ronys:
Indeed, very clear.
It might be worth breaking this into a separate command, e.g., '--help' and '--examples', where the last line of the usage (--help) output would be "Run %PROGNAME% --examples for usage examples".

PR #1366, opened 2024-08-17
[WX] Converting co-ordinates to width and height needs a +1

nobugshere:
Each time the app is started and exited, the window size, as saved in the config file, shrinks by one pixel.  When subtracting co-ordinates  (e.g. bottom-top) we need to add one to get the number of pixels, inclusive.


rafaelx:
This fixes the annoying problem of the main window shrinking by one pixel each time it is restarted. I noticed this especially when working on the dialogs, when I had to launch the application multiple times to check the layout adjustments. I ignored it since the focus was on the dialogs. Thanks for the fix.

ronys:
Thanks!
I wonder what's the root cause of this problem.

PR #1368, opened 2024-08-18
[wx] Update of Export dialog

### LINKED ISSUE #1352: [wx] Export dialogs need a UI update
File->Export To->XML Format...  is shown, but the comments also apply to the plain text dialog.

- Has both "Show Master Password" checkbox and the eyeball
- Yubikey button should be relocated to be consistent with other updated dialogs
- “Master Password” label should be relocated

Screenshot 2024-08-02 at 3 50 54 PM

rafaelx:
ExportDialogNew

ExportDialogNewYubi

See #1352 for current version of the export dialog(s).

Resolves #1352

### DIFF
--- a/src/ui/wxWidgets/ExportTextWarningDlg.cpp
+++ b/src/ui/wxWidgets/ExportTextWarningDlg.cpp
@@ -17,8 +17,9 @@
 #include "wx/wx.h"
 #endif
 
-#include "SafeCombinationCtrl.h"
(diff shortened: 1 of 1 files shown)


nobugshere:
Looks good on the Mac.  Thanks!

ronys:
Works fine with and without yubikey on Debian 12 - thanks!

Issue #1365, opened 2024-08-16
Core dump on Manjaro (v1.19.1)


fandango489:
When there is no existing user config file, running the application core dumps:- 

usr/include/c++/14.1.1/bits/stl_vector.h:1130: std::vector<_Tp, _Alloc>::reference std::vector<_Tp, _Alloc>::operator[](size_type) [with _Tp = std::__cxx11::basic_string<wchar_t>; _Alloc = std::allocator<std::__cxx11::basic_string<wchar_t> >; reference = std::__cxx11::basic_string<wchar_t>&; size_type = long unsigned int]: Assertion '__n < this->size()' failed.

App runs fine when config is present.

I looked into this and I think the problem is in the core/PWSPrefs.cpp module  (Line 265)
m_MRUitems.reserve(m_int_prefs[MaxMRUItems].maxVal);
Looks like this is expected to allocate the vector to an initial size (20), but reserve() doesn't actually allocate any items (size() reports 0 after this call).

The core dump occurs when the vector is passed out to the RecentDBList::Load function which iterates over non-existent items.
I changed the line on a local copy to :-
m_MRUitems.reserve(m_int_prefs[MaxMRUItems].maxVal); and that fixed it.

fandango489:
typo in last line - should be
m_MRUitems.resize(m_int_prefs[MaxMRUItems].maxVal); and that fixed it.

ronys:
Odd that the bug was only triggered on Manjaro, but that's C++ for you...

Thanks!

PR #1369, opened 2024-08-19
[BR1361] Fix core dump on Manjaro

ronys:
Thanks to fandango489


ronys:
Fixes #1365 (core dump on Manjaro)

PR #1370, opened 2024-08-23
Added SetTotpStartTime(time_t t) similar to other time_t fields

huven:
This avoids an unnecessary time_t -> StringX -> time_t conversion roundtrip in the iOS app, I think useful to other platforms too

### DIFF
--- a/src/core/ItemData.h
+++ b/src/core/ItemData.h
@@ -210,6 +210,7 @@ public:
   void UpdatePassword(const StringX &password); // use when password changed!
   void SetTwoFactorKey(const StringX& value) { CItem::SetField(TWOFACTORKEY, value); }
   bool SetTotpConfig(const StringX& value) { return SetFieldAsByte(TOTPCONFIG, value.c_str()); }
+  void SetTotpStartTime(time_t t) { CItem::SetTime(TOTPSTARTTIME, t); }
   bool SetTotpStartTime(const StringX& value) { return SetTime(TOTPSTARTTIME, value.c_str(), true); }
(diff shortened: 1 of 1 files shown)


src/core/ItemData.h
```diff
@@ -210,6 +210,7 @@ class CItemData : public CItem
   void UpdatePassword(const StringX &password); // use when password changed!
   void SetTwoFactorKey(const StringX& value) { CItem::SetField(TWOFACTORKEY, value); }
   bool SetTotpConfig(const StringX& value) { return SetFieldAsByte(TOTPCONFIG, value.c_str()); }
+  void SetTotpStartTime(time_t t) { CItem::SetTime(TOTPSTARTTIME, t); }
```
dskrvk, line 213: Is this meant to overload the existing SetTotpStartTime? If so, shouldn’t it also return a bool? Also, can we omit the CItem prefix for consistency with the other method?

huven: Yes, this overloads SetTotpStartTime. I copied the interface used by the other fields defined in the same file:

```
  void SetATime(time_t t) {CItem::SetTime(ATIME, t);}  // V30
  void SetCTime(time_t t) {CItem::SetTime(CTIME, t);}  // V30
  void SetXTime(time_t t) {CItem::SetTime(XTIME, t);}  // V30
  void SetPMTime(time_t t) {CItem::SetTime(PMTIME, t);}  // V30
  void SetRMTime(time_t t) {CItem::SetTime(RMTIME, t);}  // V30
```

ronys:
No sense in having this method return a bool, as all time_t values are valid, unlike the method that accepts a string.

PR #1375, opened 2024-08-28
[macOS] Use a more specific locale for english

nobugshere:
Mac only follow-on to PR #1353.
There are multiple locale variations for english.  (i.e. en_US, en_GB, etc.), but pwsafe only saves the two letter language code in the config file.  (i.e. "en")  Just using en.UTF-8 causes some inconsistent results on macOS. Specifically, with en_US set as the system locale, the date format seems to default to en_GB (D/M/Y) in WX but not in native macOS controls, such as the date picker, where en_US uses M/D/Y.  See the screenshot, below.

With this patch, if we are trying to activate generic English (en), check the system UI locale and, if it is more specific, use that instead.  This has been tested on:
macOS M1Max Sonoma 14.6.1, Xcode 15.4, wx 3.2.4 and 3.2.5

This is ifdef'ed for Mac only because the behavior is not consistent across Linux versions.  Using en_US on all systems, Ubuntu and Debian display M/D/Y in both places, so they don't need this change.  Whereas, Fedora displays D/M/Y, wrong but consistent.  With this patch enabled, it displays M/D/Y for the formatted text and D/M/Y in the date picker.  

Screenshot 2024-08-28 at 3 28 14 PM


ronys:
Since English isn't the only language that has local variations (Brazilian Portuguese comes to mind), it would be nice if this were generalized. Can it?

nobugshere:
Sure.  Sometimes I'm a bit reluctant to change too much, especially if I'm not confident I can test it well enough.  Since it's only being used to set LC_TIME and LC_CTYPE, maybe it makes sense to ignore the pwsafe language selection and just use the system default locale for those parameters.  This new commit changes it to do that.  It fixes my problem with en_US and is still Mac only.

ronys:
I'm not sure that's a good approach:
1. Consider a native French speaker running PasswordSafe on an American friend's machine. She would expect that switching to French would also expect the dates to be in European, not American, representation
2. I think that the Locale also determines the texts displayed for the standard dialog buttons, e.g., "OK", "Cancel", "Help", etc. Again, if the user specified a language, that should be used for these as well.

nobugshere:
I see your point, but I don't think it's that simple.  wxWidgets uses the native platform controls, in this case the macOS date picker control, which seems to always use the macOS settings.  I tried setting pwsafe to French (with my system set to en_US) and I got the same result as the screenshot above.  Looks like, to get consistent results, you need to change the system language, too.  After some more digging, here are some observations:

1. macOS doesn't accept the "no region" form in setlocale() (e.g. "en").  It wants a region in the form xx_YY.  (with or without the .UTF-8 suffix)  I'm not sure about Linux.
2. When pwsafe sets-up "en" (or wxLANGUAGE_ENGLISH) wxWidgets sets LC_ALL to en_GB.  I imagine that's what most of the world would want.
3. I've realized that, in the original code, the two calls to setlocale() where I made the mod are not effective because they are being passed "xx.UTF-8", which is rejected.  en_GB is set in a call to wxLocale::Init() at line 431, before we get to my mod, so I was able to override it with en_US (or whatever region I get from the system.)
4. If the pwsafe language is "fr", the call to wxLocale::Init() sets fr_FR.  So, in your scenario, my second mod would override that for LC_TIME.  My original, English only, version would not.

At this point, I'm not quite sure what the fully correct process should be.  WX wants to use the Kings English when not told otherwise.  Maybe English really is something of a special case.

ronys:
I think wx's preference of British English is more due to the project's origins see https://wxwidgets.org/about/),

In any case, I think the app should behave as follows:
1. By default, if the system language is supported by pwsafe (that is, English or a .mo file in the relevant /usr/share/locale directory), then that's what should appear when the user starts the program.
2. By default, if the system language isn't supported by pwsafe (no .no file), then English should be used.
3. If the user sets a language via Manage->Select Language, then that locale setting should be used by pwsafe, to the extent supported by wxWidgets and the operating system.

In other words, language selection in the app should be done on a "best effort" basis. I don't think it's worth the effort to force wxWidgets to behave otherwise.

nobugshere:
@ronys,
I agree with all of your points.  Just to reiterate, the place where this PR is focused happens after the language is selected and set; it's just trying to add ".UTF-8" to LC_TIME and LC_CTYPE.  Anyway, I played around with it and came to the conclusion that, even when I try to make the C strftime locale match the system, it can still result in a different format than the date picker control.  So, I gave up and took the cowards way out: I changed the "Original Value" string to always use the ISO format (YYYY-MM-DD). Hopefully, that will be wrong in a less confusing way.  This part applies to Linux as well as macOS.

As for calling setlocale(), I still wanted to have that code set correct values for the Mac, so this latest attempt does the following (with .UTF-8 appended):
- If the selected language (two letter code) matches the system language, and the system has a region (which I think it always will on macOS) use the system full name (e.g. en_US)
- Else If wxWidgets has defined a default region to go with the language, use that (en_GB for English.  I've also seen fr_FR and de_DE.  I haven't checked all the languages pwsafe uses.)
- Else, do nothing.
- No change to Linux behavior.

Issue #1051, opened 2023-11-02
Question about the staus of coretest on Linux


nobugshere:
After successfully building pwsafe from , I tried running "make test" from my build directory.  Coretest reported 8 failures (see attached log.)  
[LastTest.log](https://github.com/pwsafe/pwsafe/files/13234353/LastTest.log)

These failures occurred when pulled from git tag 1.18.0, but it also happens on 1.17.0.  This was on Fedora 38 Arm and wxWidgets 3.2.2.1.

Before I dig into them, I just wanted to know if this is a known issue, I did something wrong, or if this test just doesn't work in this environment.

Thanks, 
Richard

ronys:
Looks like all the tests that read/write files are failing.
The tests assume that they're run in the src/test directory, with a 'data' subdirectory.
Did you use cmake to build & run tests? Cmake copies these under the build directory and runs them correctly.

nobugshere:
I create and cd to the build directory then run "cmake ..;make".  When the build completes, coretest is under the src/test sub-directory with the data directory, as you described.  But, I'm still not clear on how to properly invoke the test.  Running "make test" invokes ctest and produces the output above, whether I run it from the build dir or src/test.  Cd to src/test and run "./coretest" also fails.  What's the command and what directory should I run it from?

ronys:
Running 'make test' from the build directory should work - that's what I do.
Ultimately it runs coretest from build/src/test.
Perhaps there's a file permission problem? Did you checkout and build the repo under the same user you're running the tests under?

Other possibility is that's it's somehow due to the distro + ARM architecture. I've successfully built the project on a Raspberry Pi under Debian, though. Odd.

nobugshere:
I do all the work under the same user; the permissions look good.  So far, I've tracked the initial error to the check starting at CoreImpExp.cpp:1083
```
    // convert linebuf from UTF-8 to StringX
    StringX slinebuf;
    if (!conv.FromUTF8(reinterpret_cast<const unsigned char*>(linebuf.c_str()),
      linebuf.length(), slinebuf)) {
      // XXX add an appropriate error message
      numSkipped++;
      continue;
    }
```

nobugshere:
I figured it out: it was my locale setting. I had LC_ALL=POSIX set in my environment. If I unset LC_ALL, and let it default to en_US.UTF-8, "make test" works from either the build directory or build_dir/src/test.  Submitted PR #1052 to update the build instructions.

PR #1377, opened 2024-08-31
If caller requests UTF8 encoding then use UTF8 encoding

huven:
On macOS, pws_os::wcstombs ignored the isUTF8 argument. By calling std::wcstombs the user-preferred encoding is used (as instructed by Startup). This causes e.g. incorrect calculations of stretched keys if a non-UTF8 locale is used. This PR fixes that.

Below is a simple code snippet that demonstrates the issue on macOS (unit-tests do not run on macOS, so that is why I include it here for reference). After applying this PR the test works ok.

- PWScore::CheckPasskey will try both true and false for isUTF8 (using PWS_PK_CP_ACP). This allows unlocking safes that were saved with incorrect stretched keys as pws_os::wcstombs will fallback to the old (erroneous) behavior.
- pws_os::mbstowcs needs similar fix, seems to have less impact, will add that to this PR.
- there might be compatibility issues reading other parts of the safe (not yet encountered though), is a mechanism needed similar to PWS_PK_CP_ACP to revert to the old behavior during reading a safe?

Thoughts?

```
    chdir(NSTemporaryDirectory().fileSystemRepresentation);
    StringX passphrase(_T("é"));
    StringX fname(_T("test.psafe"));
    
    const char* loc = std::setlocale(LC_ALL, "en_US.UTF-8");
    NSAssert(loc != NULL, @"Failed to set locale to en_US.UTF-8");

    // Create empty safe with non-ascii passkey
    PWSfileV4 fw(fname, PWSfile::Write, PWSfile::V40);
    NSAssert(fw.Open(passphrase) == PWSfile::SUCCESS, @"write open error");
    NSAssert(fw.Close() == PWSfile::SUCCESS, @"write close error");
    
    // Reread with en_US.UTF-8 locale
    PWSfileV4 fr(fname, PWSfile::Read, PWSfile::V40);
    NSAssert(fr.Open(passphrase) == PWSfile::SUCCESS, @"read 1 open error");
    NSAssert(fr.Close() == PWSfile::SUCCESS, @"read 1 close error");

    // Now change the locale, should have no impact in theory
    loc = std::setlocale(LC_ALL, "de_DE.ISO8859-15");
    NSAssert(loc != NULL, @"Failed to set locale to de_DE.ISO8859-15. To list the available locales: locale -a");

    // Reread with de_DE.ISO8859-15 locale
    PWSfileV4 fr2(fname, PWSfile::Read, PWSfile::V40);
    NSAssert(fr2.Open(passphrase) == PWSfile::SUCCESS, @"read 2 open error");  //  <-- crashes here
    NSAssert(fr2.Close() == PWSfile::SUCCESS, @"read 2 close error");
```

### DIFF
--- a/src/os/mac/utf8conv.cpp
+++ b/src/os/mac/utf8conv.cpp
@@ -45,6 +45,9 @@ size_t pws_os::wcstombs(char *dst, size_t maxdstlen,
   if (!isUTF8)
     return ::wcstombs(dst, src, maxdstlen) + 1;
 
+  if (srclen == size_t(-1))
+    srclen = wcslen(src);
+
   // Convert to UTF-16
   CFStringRef str = CFStringCreateWithBytes(kCFAllocatorDefault, reinterpret_cast<const unsigned char *>(src), srclen*sizeof(wchar_t), wcharEncoding, false);
   if (str == NULL)
@@ -71,6 +74,9 @@ size_t pws_os::mbstowcs(wchar_t *dst, size_t maxdstlen,
 {
   if (!isUTF8)
     return ::mbstowcs(dst, src, maxdstlen) + 1;
+
+  if (srclen == size_t(-1))
+    srclen = strlen(src);
 
   // Convert to UTF-16
   CFStringRef str = CFStringCreateWithBytes(kCFAllocatorDefault, reinterpret_cast<const unsigned char *>(src), srclen, kCFStringEncodingUTF8, false);
--- a/src/os/mac/utf8conv.cpp
+++ b/src/os/mac/utf8conv.cpp
@@ -21,9 +21,9 @@
 #include <CoreFoundation/CoreFoundation.h>
 
 #if defined(PWS_LITTLE_ENDIAN)
-static const CFStringEncoding wcharEncoding = kCFStringEncodingUTF32LE;
+#define wcharEncoding kCFStringEncodingUTF32LE
 #else
-static const CFStringEncoding wcharEncoding = kCFStringEncodingUTF32BE;
+#define wcharEncoding kCFStringEncodingUTF32BE
 #endif
 
 using namespace std;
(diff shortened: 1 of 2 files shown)


nobugshere:
Is this possibly related to #1051?  There are several tests that fail on Linux under a non-UTF-8 locale (e.g. "C"). It might be worth noting that the Startup static constructor trick will just default to "C" in a GUI app because macOS doesn't pass the LANG/LC* variables in the environment.  It should work as expected in a CLI app.  I'm not sure about the various Linux environments.

huven:
Yup, looking at the code in os/unix the same issue seems to be present there. When encoding from/to UTF8, the locale set by the user should really be ignored.

src/os/mac/utf8conv.cpp
```diff
@@ -32,15 +40,68 @@ class Startup {
...
+  if (!isUTF8)
+    return ::wcstombs(dst, src, maxdstlen) + 1;
+
+  if (srclen == size_t(-1))
+    srclen = wcslen(src);
+
+  // Convert to UTF-16
+  CFStringRef str = CFStringCreateWithBytes(kCFAllocatorDefault, reinterpret_cast<const unsigned char *>(src), srclen*sizeof(wchar_t), wcharEncoding, false);
+  if (str == NULL)
+    return 0;  // return wcstombs + 1, so 0 to signal error
```
dskrvk, line 54: What if the input string was of length 0? We could return -1 to distinguish the error case from the empty string case.

huven: The logic has not changed here with the previous implementation. If the input string has length 0 then this function will return 1 (see the last statement of the method).

Previously the implementation was simply 'return ::wcstombs(dst, src, maxdstlen) + 1;' so it would also return 1 in that case.

PR #1378, opened 2024-09-01
Use normalized UTF8 for filesystem operations

huven:
This PR fixes two issues on macOS

- on macOS, filesystem operations must use UTF8 encoding instead of user-preferred locale (similar issue as #1377)
- on older filesystems, UTF8 normalization is important for uniqueness

This PR introduces a new function createFileSystemRepresentation that uses the correct canonical encoding, prototype chosen to stay as close to the original implementation as possible. This allows replacing multiple lines of code with just one function call in multiple places.

First commit only replaces code in FOpen to demonstrate the change (and because it allows me to close several user reported issues...). Similar changes need to be made in the other filesystem operations (and possibly in other files like dir.cpp).

Thoughts?

### DIFF
--- a/src/os/mac/file.cpp
+++ b/src/os/mac/file.cpp
@@ -22,9 +22,7 @@
 #include <dirent.h>
 #include <fnmatch.h>
 
-#ifdef UNICODE
 #include <CoreFoundation/CoreFoundation.h>
-#endif
 
 #include "../file.h"
 #include "../env.h"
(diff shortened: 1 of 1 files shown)


huven:
Some background technical details, not a very recent post, but still applicable: https://lists.apple.com/archives/darwin-dev/2005/Nov/msg00107.html

ronys:
I think it would be cleaner to have a dummy implementation of createFileSystemRepresentation() if UNICODE isn't defined, rather than sprinkling #ifdefs throughout the file.

huven:
@ronys Have looked at this. It looks like UNICODE is not optional at all, if you currently #undef UNICODE at the top of the file you get lots of errors (since all parameters are stringT, the non-unicode path takes c_str() of that (which is still wchar *) and feeds that to the filesystem layer (which expects char *). 

Just an example:

Screenshot 2024-09-06 at 20 24 24

I guess everybody is using unicode mode on Mac, and I really cannot think of any reason not to.

So... instead of fixing all this, wouldn't it be better to have just one #ifndef UNICODE check at the top that #errors out and remove all the other ones?

huven:
Committed the changes as described above. Makes the file much easier to read.

The UNICODE define is now explicitly required at the top, which it always (implicitly) was as the deleted code did not compile.

Added a comment how non-Unicode support can be re-added easily, although I don't recommend doing that as apparently nobody needs/uses that on Mac.

ronys:
Actually, the UNICODE define is a relic from the days when PasswordSafe was ASCII only, and the transition to Unicode was done incrementally, over time.

PR #1380, opened 2024-09-07
Fix some different type warnings

nobugshere:
Related to #1377 

Clang is giving me two of these warnings:
Comparison of integers of different signs: 'CFIndex' (aka 'long') and 'size_t' (aka 'unsigned long')

I added static_cast<size_t> statements.


nobugshere:
I have no idea what happened here.  Closing and starting over.

Issue #1371, opened 2024-08-24
Only one name is saved to MRU list - Regression from #1369


nobugshere:
Related to issue #1365 and PR #1369

With PR #1369 applied, it appears that only the current filename is being saved to the MRU list, any others are discarded.  I suggest reverting it until it can be investigated more throughly.

PR #1382, opened 2024-09-07
[BR1371] Fixed regression, recently used files are now remembered co…

ronys:
…rrectly.

Only tested on Windows, but code is common to all platforms. Would appreciate tests on other platforms.

Fix for #1371

### DIFF
--- a/docs/ReleaseNotes.md
+++ b/docs/ReleaseNotes.md
@@ -10,6 +10,7 @@ PasswordSafe 3.67.0 Release ??? ?? 2024
 New features in 3.67.0
 ----------------------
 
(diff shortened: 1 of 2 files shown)


nobugshere:
Seems to work on the Mac, except the order of the drop-down list is different, the latest item is on the bottom, the others seem jumbled, not exactly reverse order.  I think the current release keeps the list in MRU order, latest at the top, oldest at the bottom.

ronys:
Yes, the order is indeed getting mixed up, looks like a wx-specific issue in src/ui/wxWidgets/RecentDbList.h

Merging this now, will look into the order issue later.

PR #1384, opened 2024-09-10
Make writing safe optional on changing passkey

huven:
This allows for custom save strategies like backup file etc.
Also, on Mac filesystem operations must be wrapped inside a file coordinator.

### DIFF
--- a/src/core/PWScore.cpp
+++ b/src/core/PWScore.cpp
@@ -1593,11 +1593,12 @@ bool PWScore::BackupCurFile(unsigned int maxNumIncBackups, int backupSuffix,
   return pws_os::RenameFile(m_currfile.c_str(), bu_fname);
 }
 
-void PWScore::ChangePasskey(const StringX &newPasskey)
+void PWScore::ChangePasskey(const StringX &newPasskey, bool bWriteFile)
 {
   SetPassKey(newPasskey);
   time(&m_hdr.m_whenpwdlastchanged); // update master password changed timestamp
-  WriteCurFile(); // Save immediately!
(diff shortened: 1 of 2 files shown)


ronys:
The rationale behind the current implementation was to avoid the scenario where a user isn't sure if the new master password has "taken effect" or not, say in the case of the database locking, or a crash, or shutdown.... Clicking on OK meant the change took effect.

I'm unfamiliar with the Mac "file coordinator", but if it requires that writing the new file be delayed, I'd make sure that the 'modified' flag is set when the master password's changed until it's saved.

huven:
Agree, the best user experience is having the change saved to disk asap. As indicated, that can involve other logic, like making a backup first, writing to a temporary file, moving it to the correct spot.

The iOS/Mac pwSafe does that, and also wraps the file coordinator around it.

I can wrap all that around this call (which would make this PR unnecessary), but it seemed more clean to make the write optional here since the method name suggests it changes the passkey only. Adding the optional parameter makes the prototype much more descriptive of what the method is actually doing, don't you agree?

ronys:
I understand what you're saying, but still have reservations about adding the Boolean flag, as this is an invitation for a maintainer down the line to misuse this, breaking the "save immediately" experience that I've described above.
Perhaps renaming the method from ChangePasskey() to ChangePasskeyAndSave() would further clarify this?

huven:
Hi Rony, I'll delete this PR as it clearly does not align, and that is ok, really.

I will use SetPassKey instead since it is publicly exposed, so no problem there.

Issue #1323, opened 2024-06-30
Program is not locked when OS locks


igor2x:
Password Safe v1.18.2 flatpak on Ubuntu 24.04.

1. Check Manage | Options | Security | Lock password database on workstation lock is checked.
2. Press Super+L to lock Ubuntu.
3. Login back to Ubuntu. Now problem appears. Password Safe is still unlocked, but according to step 1, it should get locked on workstation lock event.

nobugshere:
I see the same issue on Mac and X11 variants.  Currently, the lock-on-idle-timeout feature is the closest alternative.  I haven't been able to find any way in wxWidgets to get a screen-lock notification.  Given all the different window managers, I would hate to have to support a low-level solution for every platform.  I'm sure I'm missing something obvious, but unless someone knows how to make this work, I suggest we remove the option to avoid confusion.

Issue #1223, opened 2024-04-15
System Tray enabled and when not working unable to accesss Password Safe again


igor2x:
Password Safe v1.19-pre on master branch on Ubuntu 23.10.

If I enable Manage | Options | System | Put icon in System Tray then when clicking on Lock button from toolbar then I am unable to access Password Safe anymore. a) Alt+Tab does not show. b) Application disappears from Ubunut's dash. c) Password Safe is not displayed in system tray. It looks like I can not access it anymore, but I see it from terminal ps aux | grep pwsafe` application is still alive. Opening another new pwsafe instance and I get "Lock is done by yourself?" question, indicating application didn't release lock on database yet, probably still alive.

In Manage | Options | System | System Tray: There is warning displayed in red color: "There appears to be no system tray support in your current environment. Any related functionality may not work as expected."

My suggestion is, if application recognizes there is no system tray support, then both options from System Tray: "Put icon in System Tray" and "Start Password Safe at Login" should get automatically unchecked and greyed out. I assume non-technical user may get into problem to never get Password Safe program again or when launching every time getting "Lock is done by yourself".

Idea of this issue: Non working options should be disabled and not allowed user to get into nasty situations.

Issue #823, opened 2021-12-15
Ubuntu 20.04 - options create key shortcuts displays empty 2col table - how does it work?


steffen-wirth:
Hi!

Can someone explain how the key shortcuts on the linux version works?

rafaelx:
It's just not fully implemented yet.

PR #1389, opened 2024-09-16
[wx] Disable unimplemented settings

nobugshere:
Partially addresses: #823, #1223, #1323

This simply disables (gray-out) or hides some Option settings that are not implemented or are known not to be supported, in an attempt to not mislead the users.  No code has been deleted, they can simply be re-enabled if and when they are implemented.

1. Disable "Lock DB on workstation lock".  I don't see an obvious way to do this in wxWidgets.
2. Disable the "Start Password Safe at login" control.
3. Disable the system tray setting when the warning that it may not work is displayed.  Mainly happens on Wayland based desktops.
4. Hide the entire shortcuts tab in the Options dialog.

Tested on:
macOS M1Max Sonoma 14.6.1, Xcode 15.4, wx 3.2.4 and 3.2.5
Fedora 40 ARM64 VMware,    KDE/Wayland,   gcc 14.1.1, wx 3.2.4, GTK 3.24.43
Ubuntu 24.04 ARM64 VMware, Gnome/Wayland, gcc 13.2.0, wx 3.2.4, GTK 3.24.41
Debian 12.7 ARM64 VMware,  Xfce/X11,      gcc 12.2.0, wx 3.2.2, GTK 3.24.38

---
Edit: I went back and forth over whether it's better to keep 1 and 2 visible but gray, or to hide them completely.  Any thoughts?


ronys:
I think that keeping them visible but disabled (gray) increases the chances of the features being eventually implemented.

nobugshere:
Yeah, I had a similar thought.  I made some refinements:
- Enable Shortcuts tab but gray-out all controls and add "Not Implemented" to the title.
- Decouple the "Lock on workstation lock" control from the actual preference setting, so it will display false without changing the config file (in case it's shared with Windows)
- Add TODO markers to comments

Karneades:
I think that the project should track "to be implemented" features outside of the tool's GUI to track dev work in GitHub/or SF as those are the places where devs are around. I asked myself the same when started with pwsafe, why there is an empty table for shortcuts as well.

PR #1391, opened 2024-09-19
Fixed regression, an entry's password history list responds to clicks again

### LINKED ISSUE #1390: Password History list broken
(reported by Bob L.)

1. Clicking on a row in the Password History list of an entry no longer copies the password in that row to the clipboard
2. Clicking on a header in the same no longer sorts by that header.

ronys:
Fixes #1390

### DIFF
--- a/src/ui/Windows/AddEdit_PropertyPage.cpp
+++ b/src/ui/Windows/AddEdit_PropertyPage.cpp
@@ -58,9 +58,6 @@ BOOL CAddEdit_PropertyPage::OnNotify(WPARAM wParam, LPARAM lParam, LRESULT* pRes
   ASSERT(pResult != NULL);
   NMHDR* pNMHDR = (NMHDR*)lParam;
 
-  if (pNMHDR->hwndFrom != m_hWnd && pNMHDR->hwndFrom != ::GetParent(m_hWnd))
-    return FALSE;
-
   if (pNMHDR->code == PSN_QUERYINITIALFOCUS) {
     CWnd* pCtl = GetDlgItem(IDC_TITLE);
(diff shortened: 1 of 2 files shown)


src/ui/Windows/AddEdit_PropertyPage.cpp
```diff
@@ -58,9 +58,6 @@ BOOL CAddEdit_PropertyPage::OnNotify(WPARAM wParam, LPARAM lParam, LRESULT* pRes
   ASSERT(pResult != NULL);
   NMHDR* pNMHDR = (NMHDR*)lParam;
 
-  if (pNMHDR->hwndFrom != m_hWnd && pNMHDR->hwndFrom != ::GetParent(m_hWnd))
```
dskrvk, line 61: Trying to understand what's happening here and what the issue was. So we're only looking for events coming from either the current window (the edit dialog I assume?) or its parent (the main pwsafe window?), but in this case the event source is something else - what is it? And I wonder why this check was put in place - what events were we trying to filter out and whether they still happen.

ronys: This method was added in order to set the focus on the Title field in the Basic tab of the Add/Edit entry dialog.

The check I removed prevented Notify events from being propagated from other controls, which is what caused the regression: OnNotify events for the password history table were ignored due to this predicate.

I'm not sure why the predicate was put there in the first place, I suspect in an attempt to avoid false positives for other controls with the same ID as the title field.

In any case, removing this check fixed the problem, and, as far as I could tell by testing, had no other side effects.

dskrvk (approved):

PR #1399, opened 2024-10-12
Add 'Not Implemented' text

### LINKED ISSUE #1394: Simplify text in Sytem Tray if there is no support for it
I see [PR 1389](https://github.com/pwsafe/pwsafe/pull/1389) was merged. This PR greys out functionalities that do not work. This is great.

I have downloaded today's latest version for Ubuntu from Github Actions and I see the following in Manage | Options | System:
image

In my humble opinion red warning: "Any related functionality may not work as expected." is now redundant. This was warning before above PR was accepted, if user enabled setting something "may not work as expected." But now when option is grayed out, we can remove this sentence (blue rectangled on picture above). Now we removed option (greyed out) and there is nothing that can work as unexpected.

This is small simplification. One by one small improvement to get a better product.

### LINKED ISSUE #1395: Add "not implemented" at workstation lock option
I see [PR 1389](https://github.com/pwsafe/pwsafe/pull/1389) was merged. This PR greys out functionalities that do not work. This is great.

I have downloaded today's latest version for Ubuntu from Github Actions and I see the following in Manage | Options | Security:
image

"Lock password database on workstation lock" option is greyed out. This is excellent, if not working then there is not good to have option to enable something.

If we look at the "Shortcuts" the title has changed to "Shortcuts (Not Implemented)", this is great to tell user that this functionality currently is not working (but it probably will in future when option is implemented).

Now back to the above picture. From current design it is not clear what should user do to enable "workstation lock" option. May I suggest to add additional text "(Not Implemented) and string name becomes: "Lock password database on workstation lock (Not Implemented)"

Two benefits:
- user is informed that option is not implemented yet
- some new developer may see this info and implement solution

nobugshere:
Resolves #1394 
Resolves #1395

### DIFF
--- a/src/ui/wxWidgets/OptionsPropertySheetDlg.cpp
+++ b/src/ui/wxWidgets/OptionsPropertySheetDlg.cpp
@@ -744,7 +744,7 @@ wxPanel* OptionsPropertySheetDlg::CreateSecurityPanel(const wxString& title)
   security_LockOnMinimizeCB->SetValue(false);
   itemBoxSizer87->Add(security_LockOnMinimizeCB, 0, wxALIGN_LEFT|wxALL, 5);
 
-  wxCheckBox* security_LockOnWindowLockCB = new wxCheckBox( itemPanel86, ID_CHECKBOX28, _("Lock password database on workstation lock"), wxDefaultPosition, wxDefaultSize, 0 );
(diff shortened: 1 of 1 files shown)


ronys:
The Ubuntu-latest build fail appears to be a problem with the build environment, and not the code.

Issue #1408, opened 2024-11-08
Entry time fields not updated in sync


ronys:
(From Volker via Email)

last week I had to merge two PWsafe files using "compare", checking manually the different entries and passwords.
So I noticed that the password change time (data record change time) was not included in the transfer.
I think this is probably a mistake, because the old date of the password change remains on the target page.

To reproduce:

03.11.2024:
Creating a new psafe3.File (2024-11-03 Test-DB#1.psafe3)
PW "1234"
2 entries
04.11.2024:

Creating a new psafe3.File (2024-11-04 Test-DB#2.psafe3)
PW "1234"
2 entries (same name but different PWs)

Now I try to compare the two files.

First, I open the older version (2024-11-03 Test-DB#1.psafe3)
After that I open the newer version (2024-11-04 Test-DB#2.psafe3)
PWsafe shows, that there is different content in Password and Password-History.

I would like to transfer newer content to the older content.
(Just for explanation why I used this way - in my real problem, the older version contained more data and only a few entries are in the newer version, so I thought it makes no sense moving everything to the newer version.)

So I go to the second column and select "synchronize" - all fields are selected to be transferred in the next window.

If you do now a comparison (doesn't matter from column 1 or column 2) you will see, that the timestamp  (Password modifacation time) is not transfered from 2024-11-04 Test-DB#2.psafe3.
Also Group:Title:Username is shown as different, but it isn't.
You can also see this in the "statistic" where to old information is still available - I renamed the file to 2024-11-03 Test-DB#1 (Result after Sync).psafe3.
[repwsafepossibleerrorindbcomparisonandtransfer.zip](https://github.com/user-attachments/files/17677345/repwsafepossibleerrorindbcomparisonandtransfer.zip)

ronys:
Needs to be fixed both in Windows and wx.
This will be in 2 separate PRs.

ronys:
fixed.

PR #1409, opened 2024-11-08
[GH1408] Update time fields correctly in sync operations

ronys:
Fix for #1408 is currently Windows-only. Needs to be applied to wx as well.

### DIFF
--- a/src/core/Item.h
+++ b/src/core/Item.h
@@ -124,6 +124,13 @@ public:
     LAST_FIELD
   };
 
+  static bool IsTimeField(int ft)
+  {
+    return ft == CTIME || ft == PMTIME || ft == ATIME || ft == XTIME || ft == RMTIME ||
+      ft == TOTPSTARTTIME ||
+      ft == ATTCTIME || ft == FILECTIME || ft == FILEMTIME || ft == FILEATIME;
+  }
+
   // Status returns from "ProcessInputRecordField"
   enum {SUCCESS = 0, FAILURE, END_OF_FILE = 8};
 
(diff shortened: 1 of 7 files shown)


src/core/Item.h
```diff
@@ -124,6 +124,13 @@ class CItem
     LAST_FIELD
   };
 
+  static bool IsTimeField(int ft)
```
rafaelx, line 127: Although the _FieldType_ data type exists as an enumeration, the API uses an integer type instead. Is this for historical reasons, programming reasons, or is a C API intended to be provided to make it easier for other projects to create bindings?

ronys: Historical reasons. Today I'd keep this as a reminder that the values passed may be outside those defined in the enum, as in forward compatibility. That is, reading a database written on a newer version that has fields (enums) that are unrecognized by the reading version. PasswordSafe will read and write these fields, but won't know how to display them.

src/ui/Windows/MainFile.cpp
```diff
@@ -3778,10 +3778,16 @@ LRESULT DboxMain::SynchCompareResult(PWScore *pfromcore, PWScore *ptocore,
              sxSync_DateTime, IDSC_SYNCPOLICY);
         if (pPolicyCmd != NULL)
           pmulticmds->Add(pPolicyCmd);
+      } else if (i == 8) { // XXX hack test theory
```
dskrvk, line 3781: Did you mean to keep this XXX comment? If so, it should have more context to help future maintainers.

Also, is 8 one of the values of `FieldType` - could we use that instead?

ronys: Oops - wrong commit pushed...
More work due in any case - thanks!

dskrvk (approved):

PR #1412, opened 2024-11-16
[i18n] Updates in Russian Help/.po

pm-kan:
Sorry for the delay. Was too busy when new version was released(


ronys:
Thanks! Better late than never...

PR #1414, opened 2024-11-29
[wx] BR1317 - alias syntax detected in entry's password

ronys:
Still needs error checking - refactor code from Windows version.


src/ui/wxWidgets/AddEditPropSheetDlg.cpp
```diff
@@ -2446,8 +2446,15 @@ Command* AddEditPropSheetDlg::NewEditEntryCommand()
...
-    m_Item.UpdatePassword(tostringx(m_BasicPasswordTextCtrl->GetValue()));
-    m_Item.SetPMTime(t);
+    BaseEntryParms pl;
+    auto password = tostringx(m_BasicPasswordTextCtrl->GetValue());
+    if (m_Core.ParseBaseEntryPWD(password, pl) && (pl.ibasedata > 0)) {
+      m_Item.SetAlias();
+      m_Item.SetBaseUUID(pl.base_uuid);
+    } else {
+      m_Item.UpdatePassword(password);
+      m_Item.SetPMTime(t);
```
dskrvk, line 2456: Should we `SetPMTime` in both cases?

rafaelx:
Doesn't this change only affect editing an existing entry so that an alias can be created from it?
The dialog for creating a new entry doesn't seem to support this yet, so 1317 would not be solved if this was the intention.
This PR is still a draft, so there may be more to come or I may have missed a step in testing.

src/ui/wxWidgets/AddEditPropSheetDlg.cpp
```diff
@@ -2446,8 +2475,39 @@ Command* AddEditPropSheetDlg::NewEditEntryCommand()
   time_t t;
   time(&t);
   if (changes & Changes::Password) {
-    m_Item.UpdatePassword(tostringx(m_BasicPasswordTextCtrl->GetValue()));
-    m_Item.SetPMTime(t);
+    BaseEntryParms pl;
+    pl.InputType = m_Item.GetEntryType();
+    auto password = tostringx(m_BasicPasswordTextCtrl->GetValue());
+    if (m_Core.ParseAliasPassword(password, pl)) {
+      // Core validations:
```
dskrvk, line 2482: The validation code below is the same as in `NewAddEntryCommand` (and in `CAddEdit_Basic::OnApply`) - could we extract a common method (or at the very least have a method to construct the GTU string so it's consistent)?

PR #1417, opened 2024-12-13
Attachment support in V3

huven:
Here is a draft PR to accommodate attachments in V3.

- Unit tests copied from V4 to V3 (with minimal changes applied)

- I've moved some of the V4 method prototypes (related to attachments) up from PWSfileV4.h to PWSfile.h to make them available to V3 (and future versions). I kept the implementation where it was (forking V4 to V3) to allow version specific changes in the future. This also allowed cleaning up some of the V4 specific hacks that were present in CItemData. In V1V2 I've added minimal code to always fail as this does not apply there. An alternative would be to introduce another PWSfileWithAttachment layer and move V3 and V4 there, but that introduces more complexity than it's worth.

- There is some SaveState() / RestoreState() wizard stuff in V4 ReadRecord. I've copied that to V3, though it might be better to move that to PWScore. That is, leaving the save/restore code in the file code, but moving the logic when to call it to PWScore::ReadFile. What do you think?

- Documentation updated (mostly copied from V4).

    For the AttRef field there are two representations, the on-disk on (which can differ per database version) and the in-memory one (stored in CItem::m_fields, which must be database version independent, and needs special attention when storing multi-valued fields).

    So we can just mimic the V4 behavior and store multiple AttRef fields per entry. The core lib just needs to be extended to store/retrieve these correctly in one m_fields element to support more than 1 attachment per entry. Which is best done in a separate PR as that will also improve V4.

### DIFF
--- a/docs/formatV3.txt
+++ b/docs/formatV3.txt
@@ -204,6 +204,7 @@ PasswordSafe V3.29 introduced Format 0x030B
 PasswordSafe V3.29Y introduced Format 0x030C
 PasswordSafe V3.30 introduced Format 0x030D
 PasswordSafe V3.47 introduced Format 0x030E
+PasswordSafe V3.68 introduced Format 0x030F
 
 [2] A universally unique identifier is needed in order to synchronize
 databases across different machine. Representation is described in
@@ -543,6 +544,7 @@ Note: A 'base' entry can not have both aliases and shortcuts.
 
 [22] The UUID of an attachment record associated with an entry.
 Different entries can refer to the same attachment uuid.
+This was introduced in version 0x030F (PasswordSafe V3.68).
 
 [23] This is the shared secret for sites using Time-Based One-Time
 Password Algorithm (per RFC6238) such as Google Authenticator. At
--- a/docs/formatV3.txt
+++ b/docs/formatV3.txt
@@ -362,7 +362,7 @@ Own symbols for password    0x16        Text          Y              [18]
 Shift Double-Click Action   0x17        2 bytes       Y              [15]
 Password Policy Name        0x18        Text          Y              [19]
 Entry keyboard shortcut     0x19        4 bytes       Y              [20]
-*RESERVED*                  0x1a        UUID          -              [22]
+AttRef                      0x1a        UUID          Y              [22]
 Two-Factor Key              0x1b        Binary        N              [23]
 Credit Card Number          0x1c        Text          N              [24]
 Credit Card Expiration      0x1d        Text          N              [24]
@@ -541,8 +541,8 @@ This was introduced in version 0x030D (PasswordSafe V3.30).
   e. Shortcut entry (as described in note [4])
 Note: A 'base' entry can not have both aliases and shortcuts.
 
-[22] This value is used in V4 of the PasswordSafe format. Reserving it here to
-maintain consistency with values that follow.
+[22] The UUID of an attachment record associated with an entry.
+Different entries can refer to the same attachment uuid.
 
 [23] This is the shared secret for sites using Time-Based One-Time
 Password Algorithm (per RFC6238) such as Google Authenticator. At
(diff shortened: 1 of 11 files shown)


jefftharris:
Will v3 files with this feature retain compatibility with previous pwsafe versions as well as android and iOS ports?  The hmac calculation changes seem like breaking changes for previous versions.  The older versions are no longer able to preserve fields they do not support, for example.  

Additionally, the new record type without the standard id 0 uuid field could pose problems reading the file and indexing records.

If users need attachments, shouldn't they use a v4 file?

huven:
Hi Jeff, there is definitely a compatibility issue. This draft PR originated from an email conversation with Rony. There is a desire to allow users to use attachments without forcing them to V4, so I decided to try it this way. But as you already indicated, there is a compatibility issue here. I was planning to close this PR, was waiting for Rony's judgement on it.

I did provide another suggestion: Maybe it is feasible to introduce attachments for V3 in a less intrusive way by using a new field type (0x25?) with some sort of binary encoding of the attachment and its metadata. Just as an example, it could be a normal attachment record (with the same fields as V4 just without the crypto fields) but then stored as blob inside a regular item's field.

What do you think of that path?

ronys:
Right. I forgot about the HMAC change (it's been a while since I looked at V4...).

Of course, breaking backwards compatibility is not an option, as we don't want to force users to upgrade pwsafe on all their platforms at once, even if we release simultaneously.

Looking at V4 from today's perspective, I see that I may have tried too hard to optimize for performance, (a) by changing the hmac logic, (b) by planning for reading/writing attachments separately, that is, not loading the entire DB into memory, as we do today.

Perhaps a "gentler" approach would be along what @huven suggested, a new field or fields for the metadata and data, but using the same encryption/hmac logic. We can address performance by limiting each attachment to, say, 16MB? My current master DB is ~150KB with a few hundred entries, but I think we can handle a couple of hundred MB with reasonable performance.

(New fields are marked as "unknown" by versions that don't recognize them, and are read/written without change, so this would address the compatibility issue.)

What say?

huven:
Using multiple fields instead of encapsulating this into one field fits better in the current implementation, I like that. 

It will probably limit this to one attachment per entry but as V4 has that (currently) too, I don't think that is an issue. Would be weird to have multiple attachments support in V3, that would make it impossible to upgrade such a database to V4.

On a related note: wouldn't it be better to drop the possibility of having multiple attachments from the V4 spec? Using the same compatibility argument, it will never be possible to actually introduce this in V4 as current clients will only store the last AttRef field on reading a V4 database, after which the database validation happily deletes any non-referenced attachments.

jefftharris:
Simply adding new fields to the existing record format should work, provided the HMAC covers the full field contents like others.  Records could then have at most one, non-shared attachment.  A reasonable size limit would be nice.  The existing logic to preserve unknown fields as binary blobs should work fine as it has in the past.

Speaking for the Android port, the code has to read the entire file as a stream whenever it is opened.  The Android APIs do not allow random-access to the file contents to optimize not reading all of the attachment data.  In planning for V4, the app would maybe have to save the attachments as temp files or simply load into memory.  Likewise, the whole file has to be streamed when written.  So, the HMAC optimizations really won't do much.

Regarding the number of attachments, doesn't the v4 format allow for multiples by allowing for multiple AttRef fields per record? Comment 21 seems to imply that behavior.

A couple hundred MB file will be fun to handle on Android...  May have to cache those attachments out to disk while the file is open anyways.

docs/formatV3.txt
```diff
@@ -591,6 +595,88 @@ These TOTP parameters fields have the following format:
...
+--------------------------------------------------------------------------
+AttUUID                     0x60        UUID          Y              [1]
+Title                       0x61        Text          N              [2]
+Creation Time               0x62        time_t        Y              [3]
+MediaType                   0x63        Text          Y              [4]
+FileName                    0x64        Text          N              [5]
+FilePath                    0x65        Text          N              [6]
+FileCreationTime            0x66        time_t        N              [7]
+FileModificationTime        0x67        time_t        N              [7]
+FileAccessTime              0x68        time_t        N              [7]
```
jefftharris, line 614: For this feature and v4, what are the rules for updating the access time?  Having to resave the entire file each time a user, say, views an attachment seems like a lot (thinking of Android and syncing files).  

The file creation, mod, and access times may be difficult to know on Android other than 'now'.  The attachments won't be real files but URLs and data streams provided by the system.

huven:
Yup, the v4 format allows the multiple AttRef fields. The point I tried to make is that the current implementation in the core lib only keeps the last one it reads (for a specific entry). So if an entry has multiple attachments, the first ones will have no references and therefor be deleted from the database by PWScore::Validate.

As that logic is currently included in the clients support V4, I don't think there is a clear path to allowing multiple attachments per entry.

docs/formatV3.txt
huven: The fields are marked non mandatory, so maybe just ignoring those?

Doesn't Android provide the modification time? It's been a while since I touched some Android code.. From user-perspective, I think that's the most interesting one to have.

jefftharris: I'll either ignore or set to 'now' whenever the attachment is added or modified.  Maybe I could keep an access time internally and write only if the file is written otherwise.

Android recently has disabled direct OS file access except for apps with special all files permissions.  The standard C/Java File APIs no longer work, and you have to use their custom storage framework.  Maybe there is some time info in it, I'd have to check.  I do know that there is nothing like a file path available only a file name and probably a MIME type.

jefftharris:
I haven't delved into V4 that much, unfortunately.  I've only looked a little at the format documentation and none of the C++ code itself.  It's been a long-standing TODO for Android.

docs/formatV3.txt
jefftharris: Or do like Linux does with the relatime filesystem option to update the access time only if the file has been modified since the previous access.

huven: Personally I never found atime very useful, most of the time it was causing performance issues on high I/O servers so I just disabled it.

Looking at the current code, it looks like not being updated by the existing clients either

 % grep -irl SetFileATime .
./src/core/ItemAtt.h

huven:
Will close this PR now (as it clearly is a wrong approach) and start a new one (probably tomorrow), starting with the documentation so everybody can comment on that in an early stage.

huven:
@jefftharris just adding a link here to the new PR in case you want to tune in:
https://github.com/pwsafe/pwsafe/pull/1424

PR #1420, opened 2024-12-18
V4 rounding issue for files with size multiple of 16 bytes

huven:
test

To reproduce, simply add the above file as attachment to a V4 database, close the database and notice that you cannot open it anymore.

This PR fixes the issue by:

- adding another unit test to check this specific case
- adding new inline roundUp inline function to Util.h. Nothing spectacular, but apparently complicated enough to mixup.
    Could be deployed in a few other places too (PWScore.cpp and Util.cpp), not done in this PR to keep focus.
- using this function to fix the calculations in src/core/ItemAtt.cpp and src/core/PWSfileV4.cpp

Without this PR, there is a 1:16 chance of not being able to reopen a database after adding an attachment. The database is written correctly, so no risk of a corrupted database, but still not openable without this PR applied.

### DIFF
--- a/src/core/ItemAtt.cpp
+++ b/src/core/ItemAtt.cpp
@@ -402,8 +402,8 @@ int CItemAtt::Read(PWSfile *in)
         ASSERT(in4 != nullptr);
         size_t nread = in4->ReadContent(&fish, IV, content, content_len);
         // nread should be content_len rounded up to nearest BS:
-        ASSERT(nread == (content_len/BS + 1)*BS);
-        if (nread != (content_len/BS + 1)*BS) {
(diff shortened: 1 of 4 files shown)


ronys:
Thanks for catching this and implementing the fix.

I think it would be nice to add test cases for more "special" attachment sizes, say, 0, 1, blocksize +/- 1

What say?

huven:
Absolutely. I suggest doing that in a separate PR.

I expect the 0-bytes to fail, can remember reading some part of the code and noticing that, just can't remember now where it was. Oh well, the unit-test will probably show us :-)

huven:
Noticed that the commit comment was not very descriptive, changed that to allow easy merge. No code change.

PR #1421, opened 2024-12-20
Added macos-cmake-latest.yml workflow

ronys:
Build on mac using cmake.
- As an additional check for the Cmake files
- As a step in removing the xcode project from the repo?

### DIFF
--- a/.github/workflows/macos-cmake-latest.yml
+++ b/.github/workflows/macos-cmake-latest.yml
@@ -1,0 +1,90 @@
+name: Build pwsafe with CMake on macOS
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
+
+jobs:
+  build:
+    runs-on: macos-latest
+
+    env:
+      MACOS_VER_MIN: 10.15
+      WX_VERS: 3.1.5
+      WX_SRC_DIR: ${{ github.workspace }}/wxWidgets
+      WX_BUILD_DIR: ${{ github.workspace }}/wxWidgets/build
+      WX_ARCH: "arm64,x86_64"
+      HOST_NCPU: 4
+      XCODE_CONFIG: Release
+      FORCE_REBUILD_WX: 0
+
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v4
+
+    - name: Build ykpers from upstream
+      run: |
+        git clone --no-checkout https://github.com/Yubico/yubikey-personalization.git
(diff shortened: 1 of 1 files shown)


.github/workflows/macos-cmake-latest.yml
```diff
@@ -0,0 +1,133 @@
...
+      uses: actions/checkout@v4
+
+    - name: Set build parameters
+      # If you need to rebuild wxWidgets for some reason, say you changed the build params, just increment FORCE_REBUILD_WX
+      # Some usefull debug parameters:
+      # "WX_DEBUG_FLAG=--enable-debug --enable-debug_info"
+      # "XCODE_CONFIG=Debug"
+      # "XCODE_CONFIG=Debug-no-yubi"
+      run: |
+        echo "WX_VERS=3.2.4" >> $GITHUB_ENV
```
dskrvk, line 36: Why does this differ from the version on line 17?

.github/workflows/macos-cmake-latest.yml
```diff
@@ -0,0 +1,133 @@
...
+        echo "WX_SRC_DIR=$HOME/wxWidgets" >> $GITHUB_ENV
+        echo "WX_BUILD_DIR=$HOME/wxbuild" >> $GITHUB_ENV
```
dskrvk, line 38: Different from line 19

.github/workflows/macos-cmake-latest.yml
```diff
@@ -0,0 +1,133 @@
...
+        echo "WX_ARCH=arm64,x86_64" >> $GITHUB_ENV
+        echo "WX_DEBUG_FLAG=--disable-debug" >> $GITHUB_ENV
+        echo "MACOS_VER_MIN=10.14" >> $GITHUB_ENV
```
dskrvk, line 41: Different from 16

.github/workflows/macos-cmake-latest.yml
```diff
@@ -0,0 +1,133 @@
...
+      - master
+  pull_request:
+    branches:
+      - master
+
+jobs:
+  build:
+    runs-on: macos-latest
+
+    env:
```
dskrvk, line 15: Why need this section if we're setting most of the same parameters again on lines 36-45?

ronys:
Works now, except the dmg isn't signed and therefore, I think, can't be deployed as part of the script.
Also, actions done in install/macosx should be folded into the cmake framework.
Merging to master, to be continued...

PR #1423, opened 2024-12-27
Improve naming of IsDbOpen

huven:
The m_bIsOpen ivar is updated internally but currently never used.

From a naming perspective this PR is almost obviously correct, though some code may depend on the old behavior.

Current invocations can be replaced with !GetCurFile().empty() which will make sure everything stays unchanged, but most likely the invokers will want the correct logic.

### DIFF
--- a/src/core/PWScore.cpp
+++ b/src/core/PWScore.cpp
@@ -174,7 +174,6 @@ PWScore::PWScore() :
                      m_ReadFileVersion(PWSfile::UNKNOWN_VERSION),
                      m_bIsReadOnly(false),
                      m_bNotifyDB(false),
-                     m_bIsOpen(false),
                      m_nRecordsWithUnknownFields(0),
                      m_DBCurrentState(CLEAN),
                      m_pFileSig(nullptr),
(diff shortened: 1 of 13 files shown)


ronys:
Can you add the following check:

bool IsDbOpen() const { ASSERT(m_bIsOpen == !m_currfile.empty()); return m_bIsOpen;}

With this I'll try various scenarios under Windows to get more confident in this change.

huven:
done. I think it's best to remove the assert before merging to allow these two to diverge.

ronys:
Just got around to trying this on Windows, and the assertion fails in DboxMain::OnInitDialog():

  auto initialIcon = m_core.IsDbOpen() ? m_LockedIcon : m_ClosedIcon;


Changing IsDbOpen() to reflect the flag indeed changes the initial icon in the case where there's a most recent file (the most common case).

So I think the safer course of action is to remove the unused m_bIsOpen member.

huven:
That is definitely safer. Wouldn't it also be better then to rename the method to IsCurFileSet() or something alike? That would remove the confusion (at least for me 😄 ).

Let me know and I'll update this PR.

ronys:
Good point - the name is indeed confusing.

I think IsDbFileSet() is better than IsCurFileSet(), as it's clear what file we're talking about.

huven:
That is a better name indeed, ok, I'll update the PR later today.

PR #1424, opened 2024-12-27
Draft for adding attachments to V3

huven:
This is a draft PR to discuss how to add attachment support to the V3 format.

In the V3 doc I have copied most of the fields from the V4 attachment record.

I dropped the uuid, creation time and crypto fields as the attachment is not a separate record in this setup.

Added a compression algo field that might help saving space for PDF, text and the like. The deflate algorithm seems supported on all OS's, will need an addition to the os/ tree to use the os-native libs, or adding an implementation that builds on all platforms. Do we need an additional field for storing the uncompressed size?

I'm not really sure about the added value of the FilePath field? @ronys what was the intended use of that field?


ronys:
Good start.

- Re FilePath - the idea was to help the user return the attachment to its original location. I think we can do without it.
- Re Compression - I'd start without it, under the YAGNI principle. If/when we decide to add it, we can always define the default as "no compression" for backwards compatibility.
- Re Zip: Again, for simplicity I'd start without it.

huven:
I like the concept of keeping it simple. Updated the doc as suggested.

Need to think how this can be fit elegantly into the current ItemAtt setup.

When reading an entry, a (virtual, random uuid) record can be added to PWScore and linked in 0x1a field, and the other way around on writing. Which automatically cleans up unreferenced att records :-)

docs/formatV3.txt
```diff
@@ -591,6 +598,26 @@ These TOTP parameters fields have the following format:
...
+
+    Att FileName = Original file name - kept mainly for the extension, to support
+      applications that require this for proper viewing.
+
+    Att FileCreationTime, Att FileModificationTime, Att FileAccessTime =
+      Times associated with the attachment file at the time of attachment.
+      Application may try to restore these when exporting the attachment.
+      Representation is as described in Section 3.1.3.
+
+    Att Content = The binary representation of the attachment.
```
ronys, line 618: I would add something like "interpretation of this field's value is per the media type specified in Att MediaType (0x26)".

huven: Added. I might have some time tomorrow to experiment a bit how to fit this into the existing ItemAtt api.
If that works, then that will require minimal to no changes to apps currently supporting attachments.

huven:
Do we really need the creation and access times of the attachment?

I mean, Password Safe is not a backup tool, so in case I ever export an attachment, I would actually expect these two times to equal the time of export, as it is not a restore, just a copy.

I think users are mostly concerned with retaining the last modification date as they actually use that one for sorting etc.
Do you remember why these were added to the V4 spec, for the sake of completeness, or is there an actual use-case for restoring these fields on export?

huven:
I have committed a patch how reading could look like.

Noteworthy: I've set the field numbers into the 0x40 series to keep it outside of "user" data, since these fields are used just temporarily during reading and writing. The AttRef is used for comparing.

The general idea is: after reading an entry from the file, check if it has the new fields. If yes, create a new ItemAtt (with TransferAtt()) and store it as usual. Store AttRef in ItemData and clear the "temporary" fields.

huven:
As to the reading part, the code should be pretty complete. Haven't tested anything yet.

@ronys What do you think of the concept of transferring the field-data on reading the entry to an ItemAtt?

I am not sure if the content blob is copied or decrypted multiple times implicitly by this setup, I don't know the internals of Item well enough for that. Perhaps most easy to just log those events when the writing part is available and try.

huven:
With these commits applied (and also #1426 ) the code actually starts working 😄 

Here's a v3 database with 1 entry with an image attached using this code. Password to the database is: 1234

[att.zip](https://github.com/user-attachments/files/18278011/att.zip)

I have successfully opened this database in an existing client (which does not recognize the attachment), changed a field of the entry, saved it again. And then re-opened in the patched version, image was still there, with the changed field done by the other client.

Let me know what you think!

docs/formatV3.txt
```diff
@@ -591,6 +598,28 @@ These TOTP parameters fields have the following format:
...
+      Note that this is limited to 4GB (2^32 bytes), which should suffice.
```
jefftharris, line 619: So, no smaller practical limit on the file size?

jefftharris (commented):
The changes to the file format look good for compatibility.  The MIME type will be helpful on Android to trigger the appropriate app to view the attachment (it doesn't use file extensions).

Syncing a file with large attachments will be a challenge, but I suppose it would be difficult without them if there were sufficient numbers of records.

docs/formatV3.txt
```diff
@@ -373,6 +373,13 @@ TOTP Config                 0x21        1 byte        N              [29]
 TOTP Length                 0x22        1 byte        N              [29]
 TOTP Time Step              0x23        1 byte        N              [29]
 TOTP Start Time             0x24        time_t        N              [29]
+Att Title                   0x44        Text          N              [30]
```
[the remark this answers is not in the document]
huven: That is due to the way it is integrated into the current implementation, fields above 0x40 are treated differently internally. See remark here https://github.com/pwsafe/pwsafe/pull/1424#issuecomment-2564701088

That could change btw, I will post a summary later today which will also reference this.

docs/formatV3.txt
huven: The 4GB limit mentioned is just the technical limit of the field, we could impose any arbitrary limit here, or maybe just an advice?

Are there any limits imposed by Android? On iOS there is a 120 MB limit for the password integration (used by other apps to query the psafe database), although it is pretty easy to add an option to just ignore (after decryption) the fields inside that integration since it doesn't need the attachments. 

I tried with a few large images on iOS and that seems to work ok, but still not preferable of course. Just to see if there are any performance issues, which seems not.

I am currently working on integrating attachment support into the iOS version (for V4 format, to be extended to V3 once the spec settles). This will allow a user to select an image from their device's photo library to import. The current code will offer the user to "optimize" the image as described here (https://support.huvisoft.dev/hc/en-us/articles/33413903586196-Optimizing-images-when-importing-in-your-safe).
Note that this is not released yet, so subject to change.

But that's the way I am thinking to keep the database small / manageable. Importing photos taken with the camera as-is will quickly get you a monster database.

huven:
A summary of the current state of affairs, now that the current code is able to read/write a database.

Things to decide upon:

- Do we really need the creation and access times of the attachment? (https://github.com/pwsafe/pwsafe/pull/1424#issuecomment-2564556017)

- Is the integration into the core library (transfer in/out ItemAtt on read/write) done the correct way? Another way would be to keep the fields inside the ItemData object which would avoid the copying around, and add the necessary getters/setters to ItemData to manage those fields. This would move the fields to the 0x25... range, and require a bit more code at the client side (if V4 then use existing logic, else use new getters/setters). The current copying around gives just one api for the clients at the expense of wasting CPU cycles (as there is decryption/encryption on copy).

huven:
Happy New Year!

I dropped the ctime and atime, these can be re-added easily if needed.

With that, things to decide upon:

- Is the integration into the core library (transfer in/out ItemAtt on read/write) done the correct way? Another way would be to keep the fields inside the ItemData object which would avoid the copying around, and add the necessary getters/setters to ItemData to manage those fields. This would move the fields to the 0x25... range, and require a bit more code at the client side (if V4 then use existing logic, else use new getters/setters). The current copying around gives just one api for the clients at the expense of wasting CPU cycles (as there is decryption/encryption on copy).

- @jefftharris I've added a comment on the attachment size, is it strong enough in your opinion? If you like the comment what would you prefer to put on the ...?

huven:
New test database since the content field got renumbered.. password is 1234

[att2.zip](https://github.com/user-attachments/files/18286319/att2.zip)

docs/formatV3.txt
jefftharris: I don't think there are any 32-bit Androids phones these days, and I don't think there are any hard limits on memory use.  Android will kill background/paused apps more readily if they use larger amounts of memory.  Opening a large file will be slower (obviously) and more-so on devices vs a PC.  

I have native code for the password stretching, and I wonder if I'll need native for the core decryption and HMAC computations.

I'll have to experiment a bit to see if caching to temp files is better than storing in memory even to preserve the fields until the app supports the feature.

docs/formatV3.txt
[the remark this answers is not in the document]
jefftharris: Does the internal difference warrant a note in the format doc for future additions?  Is there an end range for the 0x40 field handling?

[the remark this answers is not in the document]
huven: I don't think that is necessary, that is just an implementation detail for how it is implemented for Windows and friends (see comments in Item.h). The doc is leading.

Ends at 0x5f, as the V4 attachment starts after that.

docs/formatV3.txt
huven: The 32-bit limit is imposed by the structure of the fields themselves

> The first block contains the field length in the first 4 bytes (little-endian), followed by...

huven:
@ronys By adding the tests to V3 I found the reason again why zero size attachments are not working, see line 170 in Item.cpp. That holds for V4 too I presume.

Since this PR is for V3, I will list the requirement that the content is at least 1 byte. Which I think is consistent with how other fields behave (empty = default).

huven:
I've set the recommended limit to 100MB as there was no further input, and this works for iOS. Can be changed of course, and it is just a recommendation so anybody can use larger sizes if the environment of the user supports it.

Which then leaves just one item

- Is the integration into the core library (transfer in/out ItemAtt on read/write) done the correct way? Another way would be to keep the fields inside the ItemData object which would avoid the copying around, and add the necessary getters/setters to ItemData to manage those fields. This would move the fields to the 0x25... range, and require a bit more code at the client side (if V4 then use existing logic, else use new getters/setters). The current copying around gives just one api for the clients at the expense of wasting CPU cycles (as there is decryption/encryption on copy).

@ronys what's your opinion on this one?

The current code has chosen the easy route by using the existing ItemAtt object and just shifting the data around on read/write of the save. That was also the least amount of work.

Personally I think it would be better to leave the data inside the ItemData, the only drawback is that some of the machinery in ItemAtt will be duplicated to ItemData. I think (not sure) that can be done in a sub-class like ItemDataAtt to keep it separate.
That would imply other changes, like renumbering the fields to 0x25.., not abusing the AttRef anymore, checking the fields on item compare, ...

Looking forward to your ideas. And if you like the PR as it is, then that is ok too of course, then we are almost done 😄

huven:
To keep the momentum I will choose 😎 and refactor this to keep the fields inside ItemData. That will renumber the fields to 0x25.. and require the UI to use another set of calls to add/remove an attachment.

Will also make the 'Att Content' field optional, with absence of the field equivalent to a zero-size attachment.

huven:
Ok, done with the refactoring.

What do you think? I am pretty content with the changes.

The changes to src/ui/Windows/ShowCompareDlg.cpp are without any testing, had to do something to make the assert in that file happy..

huven:
Here's a new test database which holds an attachment with the refactored implementation.

Password: 1234

[Test V3 with Attachment.psafe3.zip](https://github.com/user-attachments/files/18475843/Test.V3.with.Attachment.psafe3.zip)

PR #1426, opened 2024-12-30
Getfield fix

huven:
This PR adds just one fix (commit 382eeb112875b9f0facf73686ad620031abc8c43) to #1420 

As that PR has not been committed yet, it is included here as it references the new roundUp function. I can rebase this PR after #1420 is applied, leaving just the one commit.

An issue in CItem::GetField(const CItemField &field, std::vector<unsigned char> &v) is fixed:

v is correctly resized before decryption, this was only done for small (< BS) lengths, but not for bigger ones.


huven:
Rebased this one to master as #1420 was applied, the change is now isolated in this PR.

PR #1428, opened 2025-01-10
Added sha256 mac using commoncrypto (improved performance)

huven:
This adds a separate implementation for Mac, almost trivial wrapper around CommonCrypto.

Measurements indicate that this provides 8 to 10 times better performance, allowing the user to select higher Unlock Difficulty, as the unlock speed is almost completely determined by sha256 performance.

### DIFF
--- a/src/core/crypto/sha256.cpp
+++ b/src/core/crypto/sha256.cpp
@@ -10,6 +10,47 @@
 // Tom St Denis, tomstdenis@iahu.ca, http://libtomcrypt.org
 //-----------------------------------------------------------------------------
 #include "sha256.h"
+
+#ifdef __APPLE__
+/*
+  Initialize the hash state
+*/
+SHA256::SHA256()
+{
+  CC_SHA256_Init(&ctx);
+}
+
+SHA256::~SHA256()
+{
+  // CC_SHA256_Final has erased ctx
+}
+
+/*
+  Process a block of memory though the hash
+  @param in     The data to hash
+  @param inlen  The length of the data (octets)
+*/
+void SHA256::Update(const unsigned char *in, size_t inlen)
+{
+  while (inlen) {
+    unsigned int len = inlen > UINT32_MAX ? UINT32_MAX : (unsigned int) inlen;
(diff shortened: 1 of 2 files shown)


src/core/crypto/sha256_mac.cpp
```diff
@@ -0,0 +1,45 @@
...
+
+/*
+  Process a block of memory though the hash
+  @param in     The data to hash
+  @param inlen  The length of the data (octets)
+*/
+void SHA256::Update(const unsigned char *in, size_t inlen)
+{
+  while (inlen) {
+    unsigned int len = inlen > UINT32_MAX ? UINT32_MAX : (unsigned int) inlen;
```
ronys, line 31: I think that this will silently break compatibility if inlen > UINT32_MAX.
Better to fail hard, I think.

src/core/crypto/sha256_mac.cpp
ronys, line 1: My personal preference is for this to be #ifdef'ed into the existing sha256.cpp file.
However, if you prefer to keep it separate, a comment explaining that this is the Apple-specific implementation (wrapper) would be useful.
It should also be conditionally added to the src/core/CMakeLists.txt file.

ronys (changes requested):

src/core/crypto/sha256_mac.cpp
huven: I don't understand. I added this little piece of code specifically for the case that inlen > UINT32_MAX, as the CC_SHA256_Update accepts 32-bit argument only. So this loop will feed blobs of size UINT32_MAX if inlen > UINT32_MAX.

In what sense could this break?

src/core/crypto/sha256_mac.cpp
huven: Let's follow your preference, I'll update the PR.

src/core/crypto/sha256_mac.cpp
ronys: My mistake. The code is fine. Apologies.

PR #1429, opened 2025-01-10
Update MIN_HASH_ITERATIONS to 16,384

ronys:
Also, allow files to be silently upgraded to this new value upon save.

This is a fully backwards comaptible change, although users may notice the slower open and save times.

### DIFF
--- a/docs/formatV3.txt
+++ b/docs/formatV3.txt
@@ -51,7 +51,7 @@ version 3 file. This tag has no cryptographic value.
 2.3 P' is the "stretched key" generated from the user's passphrase and
 the SALT, as defined by the hash-function-based key stretching
 algorithm in [KEYSTRETCH] (Section 4.1), with SHA-256 [SHA256] as the
-hash function, and ITER iterations (at least 2048, i.e., t = 11).
+hash function, and ITER iterations (at least 16,348).
 
 2.4 ITER is the number of iterations on the hash function to calculate P',
 stored as a 32 bit little-endian value. This value is stored here in order
--- a/src/core/PWSfile.h
+++ b/src/core/PWSfile.h
@@ -29,7 +29,7 @@
 // HASH_ITERATIONS is used by the key stretching algorithm.
 // MIN_HASH_ITERATIONS is a lower limit - anything lower than this
 // is considered inherently insecure.
-#define MIN_HASH_ITERATIONS 2048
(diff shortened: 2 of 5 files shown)


src/core/PWSfile.h
```diff
@@ -29,7 +29,7 @@
 // HASH_ITERATIONS is used by the key stretching algorithm.
 // MIN_HASH_ITERATIONS is a lower limit - anything lower than this
 // is considered inherently insecure.
-#define MIN_HASH_ITERATIONS 2048
+#define MIN_HASH_ITERATIONS 16384
 // MAX_USABLE_HASH_ITERS is a guesstimate on what's acceptable to a user
 // with a reasonably powerful CPU. Real limit's 2^32-1.
 #define MAX_USABLE_HASH_ITERS (1 << 22)
```
huven, line 35: This one could go up too, 22 is easy-peasy for my iPhone.  I actually also tried 2^32-1 on the phone, which is not really acceptable (takes between 30-60 seconds), but still it gives an indication that hardware is coming close to the maximum limit.

docs/formatV3.txt
```diff
@@ -51,7 +51,7 @@ version 3 file. This tag has no cryptographic value.
 2.3 P' is the "stretched key" generated from the user's passphrase and
 the SALT, as defined by the hash-function-based key stretching
 algorithm in [KEYSTRETCH] (Section 4.1), with SHA-256 [SHA256] as the
-hash function, and ITER iterations (at least 2048, i.e., t = 11).
+hash function, and ITER iterations (at least 16,348).
```
huven, line 54: I think 16,348 is much too low. NIST is recommending at least 600,000. I have tested this on several devices (including an older Android device), even with 1,000,000 I don't notice any delay on unlocking the safe.

Can we increase the bare minimum to 2^19 = 524,288 ?

As this change the V3 doc, it might be a good idea to increase the minor version and leave a historic note about the 2,048 minimum. That way, future implementors know what to expect from existing safes.

huven: To illustrate this, below is a dialog that will become available on iOS in the near future for advanced users, the screenshot is taken on my iPhone Pro 15. After accepting the recommended value the safe still opens with unnoticeable delay.
Screenshot 2025-01-22 at 16 04 51

huven:
262,144 sounds like an excellent start to me!

PR #1435, opened 2025-01-22
Use filename extension to detect V4 database

huven:
This prevents executing the key stretching twice (one on version detection, one to read the file). It also allows returning wrong_password instead of failure for incorrect passwords.

### DIFF
--- a/src/core/PWSfileV4.cpp
+++ b/src/core/PWSfileV4.cpp
@@ -1281,7 +1281,8 @@ int PWSfileV4::ReadHeader()
 
 bool PWSfileV4::IsV4x(const StringX &filename, const StringX &passkey, VERSION &v)
 {
-  if (CheckPasskey(filename, passkey) == SUCCESS) {
+  if (filename.rfind(_T(".") V4_SUFFIX) == filename.size() - wcslen(V4_SUFFIX) - 1
+      || CheckPasskey(filename, passkey) == SUCCESS) {
     v = V40;
     return true;
   } else
--- a/src/test/FileV3Test.cpp
+++ b/src/test/FileV3Test.cpp
@@ -13,6 +13,8 @@
 
 #include "core/PWSfileV3.h"
 #include "os/file.h"
+#include "core/PWScore.h"
+#include <cstdio>
 
 #include "gtest/gtest.h"
 
@@ -176,4 +178,23 @@ TEST_F(FileV3Test, UnknownPersistencyTest)
   EXPECT_EQ(d1, item);
   EXPECT_EQ(PWSfile::END_OF_FILE, fr.ReadRecord(item));
   EXPECT_EQ(PWSfile::SUCCESS, fr.Close());
+}
+
+TEST_F(FileV3Test, V4Extension)
+{
+  PWScore core;
+  stringT fname4(_T("V3test.psafe4"));
+
+  PWSfileV3 fw(fname.c_str(), PWSfile::Write, PWSfile::V30);
+  ASSERT_EQ(PWSfile::SUCCESS, fw.Open(passphrase));
+  ASSERT_EQ(PWSfile::SUCCESS, fw.Close());
+  ASSERT_TRUE(pws_os::FileExists(fname));
+
+  // Change extension to psafe4
+  ASSERT_EQ(0, std::rename("V3test.psafe3", "V3test.psafe4"));
+
+  EXPECT_EQ(PWSfile::WRONG_PASSWORD, core.ReadFile(fname4.c_str(), L"WrongPassword", true));
+  EXPECT_EQ(PWSfile::SUCCESS, core.ReadFile(fname4.c_str(), passphrase, true));
+
+  ASSERT_EQ(0, std::rename("V3test.psafe4", "V3test.psafe3"));
 }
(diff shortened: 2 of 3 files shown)


src/core/PWSfile.cpp
```diff
@@ -86,6 +86,8 @@ PWSfile::VERSION PWSfile::ReadVersion(const StringX &filename, const StringX &pa
     VERSION v;
     if (PWSfileV3::IsV3x(filename, v))
       return v;
+    else if (filename.rfind(_T(".") V4_SUFFIX) == filename.size() - wcslen(V4_SUFFIX) - 1)
+      return V40;
     else if (PWSfileV4::IsV4x(filename, passkey, v))
```
dskrvk, line 91: Why can't we extend this check so there continues to be 1 if branch for each version?

dskrvk (commented):
It's usually considered insecure to assume file format based solely on name or extension. If someone renames a non-v4 file and causes this check to pass, could this result in a crash later on when the program tries to parse the file? I'd expect there to be a few preconditions that may no longer be true because of the less rigorous check.
What cost would we be paying if we leave things as they are?

huven:
> It's usually considered insecure to assume file format based solely on name or extension. If someone renames a non-v4 file and causes this check to pass, could this result in a crash later on when the program tries to parse the file? I'd expect there to be a few preconditions that may no longer be true because of the less rigorous check. What cost would we be paying if we leave things as they are?

If you rename a non-v4 file to ,psafe4, then the password will just always be wrong, as this will be followed by the password check.

The cost that is currently being paid is 2x the password check, even if the correct password is supplied. With high number of iterations that can be noticeable.
And incorrect status is returned in case of wrong password now.

src/core/PWSfile.cpp
huven: You could do the same for V3, but it is less important there since the V3 check can be done very quickly (as the header of the file reveals it). For V4 the current check is expensive, which is why I suggest the change in this PR.

dskrvk: I'm only talking about v4. There are now 2 branches - one checks for extension, and the other checks the contents. Can we combine them?

dskrvk:
> If you rename a non-v4 file to ,psafe4, then the password will just always be wrong, as this will be followed by the password check.

Have you tested this though? I'm worried about unintentional weakening of security checks or introducing security-related bugs.

src/core/PWSfile.cpp
huven: Ah, now I understand your question. To accommodate that, I think it is best to move the new "if" to the PWSfileV4::IsV4x method, would you prefer that over the current implementation? Then I'll update the PR.

huven:
Yes, that is tested. Although the theory is really much more powerful here:

- PWSfile::ReadVersion is called from one location: PWSfile::MakePWSfile (only if version == UNKNOWN_VERSION)
- The only call to this method for reading an existing database is in PWScore::ReadFile line 1269, which is immediately followed by the password check

So supplying a corrupt file (with psafe4 extension) will feed it to the password check (in PWScore::ReadFile). In the current logic the file is also immediately fed immediately to the password check (in PWSfileV4::IsV4x), so if there is a security-related bug it is already there.

Also note the change in src/test/FileV4Test.cpp in this PR which IMHO makes the API cleaner and less error-prone. It also shows that supplying an incorrect password now returns the expected result (as in V3).

src/core/PWSfile.cpp
dskrvk: Yes, that would better, but let's close on the other thread first.

dskrvk:
@ronys do you think change is safe?

Personally, I'd prefer to see some tests covering the scenario discussed (a non-v4 file with a v4 extension being opened causes the program to return an error message). Perhaps it would be fairly straightforward to adapt/copy one of the existing tests to: 1. create an entry, 2. write a new pwsafe file in v3 format, 3. change the file extension, 4. open it and check the result.
I don't think it's sufficient to just manually check the current behavior as it may change in the future, and that's where automated tests would help.

huven:
> 1. create an entry, 2. write a new pwsafe file in v3 format, 3. change the file extension, 4. open it and check the result.

That won't work (to test for an error) as the PWSfileV3::IsV3x test is executed first and will catch the v3 file regardless of the extension. That could be added as a positive test though, I will include that in this PR.

I understand your general concern to be careful in parsing files to protect the user from potential bugs in the parser, but I don't think that should be part of this PR as the existing logic already happily feeds the file to the V4 password checking logic (and continue if it passes), that is not changed by this PR, so it does not introduce a new attack vector.

The same applies to V3 btw,, currently just the PWS3 magic check there to trigger the parser. Easy to check for the PWS3-EOFPWS3-EOF too.

dskrvk (approved):

ronys:
Two comments:
1. What's the threat scenario? Only one that I an think of is that the attacker creates a file that corrupts the running process resulting in unwanted behavior, e.g., saving the all databases with a fixed (known) key. This requires that the attacker convince the target to open the poison file, and then say "oops - I meant another file" when this fails to open (but corrupts the process without crashing it). I think that such an attack can be mounted with a valid .psafe4 (or .psafe3) file as well, so the suffix change doesn't really buy us anything here.
2. My thinking in not checking the suffix was from another direction: The idea was to allow the user to rename the file to something that would camouflage it from an attacker. Not "security by obscurity" but an additional protection, perhaps.

In light of the above, do we still want to consider this PR?

huven:
I think we do. Your two comments confirm that this PR does not introduce a new attack vector, and that a user should be able to use a different extension (which is still supported with this PR).

The PR short-circuits the case when the extension is psafe4, which saves a (possibly expensive) key-stretching round for the 99% of normal use cases.

PR #1442, opened 2025-02-09
Properly handle nullptr in release code

huven:
There are many reasons the FOpen can fail in production scenarios (files being moved, permissions, etc.).

This check makes sure that a proper response is returned instead of continuing (resulting is segmentation fault).

### DIFF
--- a/src/core/PWSfileV3.cpp
+++ b/src/core/PWSfileV3.cpp
@@ -1001,6 +1001,8 @@ int PWSfileV3::ReadHeader()
     FILE *fd = pws_os::FOpen(filename.c_str(), _T("rb"));
 
     ASSERT(fd != nullptr);
+    if (fd == nullptr)
+      return false;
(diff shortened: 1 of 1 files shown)


src/core/PWSfileV3.cpp
```diff
@@ -1001,6 +1001,8 @@ int PWSfileV3::ReadHeader()
     FILE *fd = pws_os::FOpen(filename.c_str(), _T("rb"));
 
     ASSERT(fd != nullptr);
+    if (fd == nullptr)
+      return false;
```
dskrvk, line 1005: Is a similar change also useful for v4 and v2?

huven: They both call CheckPasskey which already contains this check (for both versions).

ronys:
Nice catch - thanks!

PR #1415, opened 2024-12-05
[wx] Disable Autotype controls on Wayland

rafaelx:
Resolves #1396

PR #1444, opened 2025-02-12
The absence of X11 != Wayland (i.e. macOS)

nobugshere:
This is a regression from 1.20
PR [#1415](https://github.com/pwsafe/pwsafe/pull/1415) tries to disable all autotype controls when running on a system using Wayland. However, the logic only tests for the absence of X11, so autotype is also disabled on macOS. ~~This PR adds a wxUtilities::IsDisplayManagerWayland() function which is: OS == Linux and not X11.~~. This PR adds a WhatWindowSystem() function to return the type of environment.  The recognized types are: X11, Wayland, macOS, Windows, and Unknown.  The design easily allows other types to be added if and when needed.  Currently, the only use case is to determine when running on a Wayland based system.

### DIFF
--- a/src/ui/wxWidgets/wxUtilities.cpp
+++ b/src/ui/wxWidgets/wxUtilities.cpp
@@ -311,34 +311,25 @@ enum wxUtilities::WindowSystem wxUtilities::WhatWindowSystem()
 
   // Get the env. variable and OS version only once
   if (wsType == Undefined) {
+    wsType = Unknown;
     osid = wxGetOsVersion();
-    switch (osid) {
-    case wxOS_MAC:
+    
+    if (osid & wxOS_MAC) {
       wsType = macOS;
-      break;
-    case wxOS_WINDOWS:
+    } else if (osid & wxOS_WINDOWS) {
       wsType = Windows;
-      break;
-    case wxOS_UNIX:
-      {
-        wsType = Unknown;
-        wxString XDG_SESSION_TYPE = wxEmptyString;
+    } else if (osid & wxOS_UNIX) {
+      wxString XDG_SESSION_TYPE = wxEmptyString;
 
-        if (wxGetEnv(wxT("XDG_SESSION_TYPE"), &XDG_SESSION_TYPE)) { // provides 'x11' or 'wayland'
-          if (!XDG_SESSION_TYPE.IsEmpty()) {
-            if (XDG_SESSION_TYPE == wxT("x11")) {
-              wsType = X11;
-            } else if (XDG_SESSION_TYPE == wxT("Wayland")) {
-              wsType = Wayland;
-            }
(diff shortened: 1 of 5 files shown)


src/ui/wxWidgets/wxUtilities.cpp
```diff
@@ -340,6 +340,11 @@ bool wxUtilities::IsDisplayManagerX11()
   return (isDisplayManagerX11 == 1);
 }
 
+bool wxUtilities::IsDisplayManagerWayland()
+{
+  return (wxGetOsVersion() == wxOS_UNIX_LINUX) && !wxUtilities::IsDisplayManagerX11();
```
dskrvk, line 345: Maybe we should change `IsDisplayManagerX11` to return an enum corresponding to the actual name of the display manager (if known) and rename the method accordingly. Linux users may be running other DM's besides x11 and wayland, so `IsDisplayManagerWayland` will provide an inaccurate result for them.

nobugshere:
Re-worked to generalize the window system detection.  Tested on the following:
```
macOS M1Max Sequoia 15.3, Xcode 16.2, wx 3.2.4
Fedora 41 ARM64 VMware,    KDE/Wayland,   gcc 14.2.1, wx 3.2.6, GTK 3.24.43
Fedora 40 ARM64 VMware,    Xfce/X11,      gcc 14.1.1, wx 3.2.4, GTK 3.24.43
Ubuntu 24.04 ARM64 VMware, Gnome/Wayland, gcc 13.3.0, wx 3.2.4, GTK 3.24.41

```
Other Unix/X11 systems should continue to enable/disable autotype as before.  Generally, if they define
XDG_SESSION_TYPE=x11 in the environment, it should be enabled.

src/ui/wxWidgets/wxUtilities.cpp
nobugshere: Reworked to be more general.

seamusdemora:
> This is a regression from 1.20 PR [#1415](https://github.com/pwsafe/pwsafe/pull/1415) tries to disable all autotype controls when running on a system using Wayland. However, the logic only tests for the absence of X11, so autotype is also disabled on macOS. ~This PR adds a wxUtilities::IsDisplayManagerWayland() function which is: OS == Linux and not X11.~. This PR adds a WhatWindowSystem() function to return the type of environment. The recognized types are: X11, Wayland, macOS, Windows, and Unknown. The design easily allows other types to be added if and when needed. Currently, the only use case is to determine when running on a Wayland based system.

Excellent work! I hope this isn't interpreted as an *over-bearing* post, but I feel that your addition to the PasswordSafe team is the best thing that's happened for macOS users (and likely others as well!) ever. :)

nobugshere:
@seamusdemora ,
Thank you for the kind words!

PR #1449, opened 2025-02-20
[wx] TOTP - Proposal to fix error message when copying auth code from an unsaved entry

rafaelx:
Proposed change to db85aa9, because 'm_ItemTotp' already contains the two-factor key the user entered at the Additional tab.


rafaelx:
I must confess that I forgot to test copying the auth code after working on making an added two-factor key effective in the Basic tab. I was more focused on the last added functionality.  :|

ronys:
Thanks.

How easy/hard would it be to apply the timer mechanism to this copy as well? That is, keep the clipboard updated with the current auth code as you've implemented in the main window?

rafaelx:
Oh, I didn't think of that with the Add-Edit dialog. I thought that this functionality was only relevant in the main view. - I'll have to look into that.

rafaelx:
In the Add-Edit dialog, the contents of the clipboard can be changed by copying the password. I would also like to add this as a stop condition for the periodic copying of the authentication code.

rafaelx:
I was wrong in my assumption. I just see that there is no option in the Add-Edit dialog to have the application copy the password to the clipboard. This means that the PR is ready to be merged.

ronys:
Note that the clipboard isn't updated when auth code changes under Wayland. Works fine under X11, though.

PR #1458, opened 2025-03-06
Split min hash iterations

ronys:
1. Define the  min (default) number of hash iterations for V4 files as a fixed fraction of the V3 value, to keep the times comparable.
2. Speedup the FileV3Test.AttachmentTest by using a single file, rather than 8.

### DIFF
--- a/src/core/PWSfileV4.h
+++ b/src/core/PWSfileV4.h
@@ -52,7 +52,7 @@ public:
   size_t ReadContent(Fish *fish, unsigned char *cbcbuffer,
                      unsigned char *&content, size_t clen);
 
-  uint32 GetNHashIters() const {return m_nHashIters;}
+  uint32 GetNHashIters() const {return m_nHashIters * HASH_FACTOR;} // we're fine with rounding errors
   void SetNHashIters(uint32 N) {m_nHashIters = N / HASH_FACTOR;}
   
   // Following for low-level details that changed between format versions
--- a/src/core/PWSfileV4.h
+++ b/src/core/PWSfileV4.h
@@ -53,7 +53,7 @@ public:
                      unsigned char *&content, size_t clen);
 
   uint32 GetNHashIters() const {return m_nHashIters;}
-  void SetNHashIters(uint32 N) {m_nHashIters = N;}
+  void SetNHashIters(uint32 N) {m_nHashIters = N / HASH_FACTOR;}
   
   // Following for low-level details that changed between format versions
   virtual size_t timeFieldLen() const {return 5;} // Experimental
@@ -69,7 +69,7 @@ public:
     ~CKeyBlocks();
     CKeyBlocks & operator=(const CKeyBlocks &that);
     bool AddKeyBlock(const StringX &current_passkey, const StringX &new_passkey,
-                     uint nHashIters = MIN_HASH_ITERATIONS);
+                     uint nHashIters = MIN_V4_HASH_ITERATIONS);
     bool RemoveKeyBlock(const StringX &passkey); // fails if m_keyblocks.size() <= 1...
     // ... or if passkey doesn't match.
   private:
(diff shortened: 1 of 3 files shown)


src/core/PWSfileV4.h
```diff
@@ -53,7 +53,7 @@ class PWSfileV4 : public PWSfile
                      unsigned char *&content, size_t clen);
 
   uint32 GetNHashIters() const {return m_nHashIters;}
-  void SetNHashIters(uint32 N) {m_nHashIters = N;}
+  void SetNHashIters(uint32 N) {m_nHashIters = N / HASH_FACTOR;}
```
huven, line 56: Is it intended that Get after Set does not return the original N?

src/core/PWSfileV4.h
```diff
@@ -105,6 +105,8 @@ class PWSfileV4 : public PWSfile
 
  private:
   enum  {NONCELEN = 32};
+  static constexpr uint32 HASH_FACTOR = 7; // Since number of hash iterations is defined for V3, this represents how much V4 hash is slower than V3. Otherwise, V4 is too slow for comfort.
+  static constexpr uint32 MIN_V4_HASH_ITERATIONS = MIN_HASH_ITERATIONS / HASH_FACTOR;
```
nobugshere, line 109: Is it assumed this single factor is "good enough" for all platforms?  Could a given platform have HW acceleration for one but not the other?

ronys: Good question, but since both V3 and V4 use the same hash function for key stretching (SHA256), h/w acceleration should affect both equally, so the ratio between them (HASH_FACTOR) remains the same.

It would be nice to see the unit tests run on CPUs with h/w support for hash, though.

src/core/PWSfileV4.h
ronys: Fixed - thanks!

src/core/PWSfileV4.h
huven: Just for the record, I still think the optimized implementation for Mac does not use h/w acceleration, but just the system lib which is release compiled. Adding -O2 for just sha256.cpp would bring the same speed to the unit tests.

ronys: Since I don't have access to a Mac, I can't check this.

In general, I'm worried about user complaints once the new MIN_HASH_ITERS value is released.

nobugshere: > Since I don't have access to a Mac, I can't check this.
Gee, I wonder if we know anyone who does?

I haven't been able to find any good information of what HW the M series chips may have for crypto acceleration, or how it's used. (Other than the in-line encryption for the flash storage.)
However, just an observation, using the current master branch (prior to this PR):
On an M1Max a Release build, using the Xcode project file in git, compiles with -Os and coretest runs in 4.8s wall clock.  In a Linux VM, running on the same Mac, it is compiled with -O2 and coretest runs in 26.6s.

I modified the test in SHA256Test.cpp to run 1,000,000 iterations.  On the native Mac, it took 208ms, on the Linux VM it took 1374ms. (As reported by the gtest framework.)

huven: I understand. For Mac I checked that personally some years ago, and determined that a C++ hash implementation taken from a main stream crypto library was only marginally slower than the system-provided library, that is, when -O2 optimization was applied to the code. With -O0, there was a huge difference.

If it helps: pwSafe 11 has been released with the new minimum hash iters applied (256k both for v3 and v4). No support requests received about it (yet?). Users using multiple platforms might be the first to notice.

nobugshere: @huven, I use your IOS app with a file that was originally created on Windows and since maintained with the Mac/WxWidgets version.  I haven't seen any issues, but my number of iterations was already well above the new minimum.

huven: Thanks for confirming. In the iOS version, you can enable 'Advanced - Show iterations' option. After doing that, you will see a new 'Iterations' option in the settings of a safe (after unlocking it) allowing you to measure the hash speed of your device. It is basic, but effective.

huven: Wow, it's a small world 😄

PR #1459, opened 2025-03-10
Fixed IsTextField treatment of DATA_ATT_... fields

huven:
Looks like I missed the ! at line 407 of ItemData.h for the new attachment fields.. 🙃

Also added the two non-text fields to CItemData::GetFieldValue for completeness.


ronys:
Thanks.

Ideally, unit tests should catch this sort of thing. Will look into how they can be improved.

Issue #1470, opened 2025-03-25
[wx] Edit Entry and Cancel button and Password Safe freezes


igor2x:
Password Safe v1.20 flatpak and also tested with the latest  PR1468 (that fixes flatpak creation):
1. Edit Entry.
2. Click on Cancel button and dialog closes successfully.
3. Edit Entry
4. Type in something in Title field and click on Cancel and Password Safe freezes.

rafaelx:
In my case (Ubuntu 24.10 with X / Wayland in a VM), I'm asked if I want to discard the changes. After clicking Discard, I can then continue interacting with the application normally.

nobugshere:
I can't reproduce the problem either.  I built the Flatpak from the same PR1468 branch on each of the following:
Fedora 41 ARM64 VMware,      Xfce/X11,      gcc 14.2.1, wx 3.2.4, GTK 3.24.43
Ubuntu 24.04.2 ARM64 VMware, Gnome/Wayland, gcc 13.3.0, wx 3.2.4, GTK 3.24.41

abhishekd2016:
I am having a similar issue on Bazzite and flatpak. Password safe freezes on start itself with no way to type anything.

igor2x:
On Ubuntu 24.04 LTS using Wayland:
- I have now created flatpak from latest commit from master repository. The same issue, typing "t" letter in Title and click on "X" on top right of dialog or Cancel button and Password Safe freezes. I have opened terminal and typed in `top` command I see 0% cpu usage. When I move mouse to Password Safe dialog to some field I see mouse pointer changes to typing mouse pointer (vertical line) and if I click on Password Safe I can't type anything but I see 1% spike of pwsafe in top command.
- The same issue in currently officially Flathub's Password Safe  v1.20 release (following four steps from my first post).
- I have also built native package from Password Safe latest commit from master branch. The same freeze issue in Edit dialog when "X" from top-right dialog is clicked or Cancel button is clicked.

@abhishekd2016, I am not really sure this is the same issue. I only get freeze in Edit dialog, everything else (what I have tested) seems to work fine. Maybe you should report a open new issue for your problem, to not get mixed (and be forgotten) with "my" issue.

igor2x:
Now I have installed Password Safe v1.21 from official Flathub repository on two separate machine:
- Ubuntu 24.04
- Fedora 41
and repeated steps from my first post and in both cases Password Safe freezes after step 4.

[...]

PR #1473, opened 2025-03-29
[wx] Make message dialog child of Add-Edit dialog

rafaelx:
I created this as a draft because it still needs to be tested in combination with the auto-lock functionality. It seems to work well with these changes in my environment, but a different environment might exhibit different behavior.

Related to #1470

### DIFF
--- a/src/ui/wxWidgets/AddEditPropSheetDlg.cpp
+++ b/src/ui/wxWidgets/AddEditPropSheetDlg.cpp
@@ -3762,7 +3762,7 @@ bool AddEditPropSheetDlg::SyncAndQueryCancel(bool showDialog) {
     if (showDialog) {
       wxGenericMessageDialog dialog(
         this,
-        _("One or more values have been changed.\nDo you want to discard the changes?"), wxEmptyString,
+        _("One or more values have been changed.\nDo you want to discard the changes?"), _("Warning"),
         wxOK | wxCANCEL | wxCANCEL_DEFAULT | wxICON_EXCLAMATION
       );
       dialog.SetOKLabel(_("Discard"));
--- a/src/ui/wxWidgets/AddEditPropSheetDlg.cpp
+++ b/src/ui/wxWidgets/AddEditPropSheetDlg.cpp
@@ -3760,7 +3760,7 @@ bool AddEditPropSheetDlg::SyncAndQueryCancel(bool showDialog) {
   }
   else if (!(Validate() && TransferDataFromWindow()) || GetChanges() != Changes::None) {
     if (showDialog) {
-      wxMessageDialog dialog(
+      wxGenericMessageDialog dialog(
         this,
         _("One or more values have been changed.\nDo you want to discard the changes?"), wxEmptyString,
         wxOK | wxCANCEL | wxCANCEL_DEFAULT | wxICON_EXCLAMATION
@@ -3779,7 +3779,7 @@ bool AddEditPropSheetDlg::SyncAndQueryCancel(bool showDialog) {
 
 void AddEditPropSheetDlg::OnCancel(wxCommandEvent& WXUNUSED(evt))
 {
-  if (SyncAndQueryCancel(true) && IsShown()) {
+  if (SyncAndQueryCancel(true)) {
     EndModal(wxID_CANCEL);
   }
 }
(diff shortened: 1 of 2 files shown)


igor2x:
I have tested this PR and I can confirm that if Gnome settings on 
`gsettings set org.gnome.desktop.wm.preferences focus-new-windows 'strict'`
or default:
`gsettings set org.gnome.desktop.wm.preferences focus-new-windows 'smart`
and Discard/Cancel dialog displays with this fix just fine.

rafaelx:
@ronys Unfortunately, this isn't the final solution.
It solves the problem with the "strict" setting under Gnome, but it causes the application to block after an automatic lock because the wxMessageDialog isn't properly terminated by an EndModal(). So, there was a good reason for the null parent.

rafaelx:
I would like to summarize some observations I made so far.

#### Windows Version
If the user opens the Add/Edit dialog and the database is locked due to user inactivity after the time specified in the options, all open windows are hidden.
If the user chooses to restore the application, all previously visible dialogs will be shown again. This includes message dialogs about unsaved changes.
The user is thus not confronted with any surprises and finds the application in the same situation before the lock was activated.

#### Wx Version
The wx version behaves differently. It closes all sub-dialogs and hides the main window before locking the database.
When the user triggers the restore/unlock, only the main window is shown, and previously visible sub-dialogs are not restored. For example, a previously visible Add/Edit dialog will no longer be shown, and any changes previously made to an element will be lost.

In the current implementation, if the database is automatically locked due to inactivity and the Add/Edit dialog is open, along with the message dialog informing the user of unsaved changes, only the Add/Edit dialog is closed and the main window is hidden. However, the message dialog remains open. This occurs due to the lack of a parent-child relationship between the two dialogs. If the application is now restored and the main window is visible, interaction with the application is not possible because the message dialog was not properly closed. If the message dialog is subsequently closed, an assert is triggered in the debug build (see below).

The first attempt to fix #1470 was to simply add the missing parent-child relationship between the Add/Edit dialog (parent) and the message dialog (child of type wxMessageDlg). However, this confirms a problem already indicated by comments in the code.

In the case of wxMessageDlg, wxWidgets uses the platform-specific native message dialog, which lacks some features that wxDialogs typically have. This makes it difficult to handle this dialog correctly. An alternative to wxMessageDlg is wxGenericMessageDlg, which differs slightly visually from its native variant, but is more compatible with wxWidgets and, in particular, harmonizes better with the existing pwsafe code.

- wxMessageDlg is not closed if the Close(true) method is called, whereas wxGenericMessageDlg does.
- wxMessageDlg triggers an assert if EndModal method is called. Maybe because it is not registered or recognized as a modal dialog even it is started via ShowModal.

> ./src/gtk/dialog.cpp(196): assert ""Assert failure"" failed in EndModal(): either wxDialog:EndModal called twice or ShowModal wasn't called

The second attempt to fix #1470 is to replace wxMessageDlg with wxGenericMessageDlg and is based on changes in PasswordSafeFrame::IconizeOrHideAndLock.
In the current version (master) of IconizeOrHideAndLock all sub-dialogs are closed/hidden, the main dialog is hidden, the application is minimized/iconized and the database is locked.
Closing the dialogs with Close triggers a Close event, which seems to take some additional time to process. Executing the next actions in IconizeOrHideAndLock during this time seems problematic.
The changes in IconizeOrHideAndLock therefore use CloseAllWindows to delay closing the individual dialogs and wait for them to close. Once the process of closing the sub-dialogs is complete, the rest of the code is executed.
The changes currently only cover the case from #1470. However, wxMessageDlg is used more frequently in the code. All occurrences of wxMessageDlg could be replaced with wxGenericMessageDlg so that the closing process for all subdialogs also runs correctly in other cases. To have more control over the visual appearance of the message dialog, a derived implementation of wxGenericMessageDlg could be created. The focus of this PR, however, is the functional handling of the subdialogs.

ronys:
The wxGenericMessageDlg looks better to me On Debian/Wayland:
Here's the wxMessageDialog:
image
And here's the wxGenericMessageDlg 
image
It would be nice to add a text to the title bar, e.g., "Warning".

nobugshere:
Sadly, on the Mac, the native control looks better or, at least, more Mac like :-)  (I can't speak to its functional behavior.) I wonder how different it is on other Linux desktops (Gnome, KDE, Xfce, etc.) Or, are they the same because they all use GTK?

MacOS Sequoia (looks the same with WX 3.2.4 and 3.2.8):
wxMessageDialog:
Screenshot 2025-05-05 at 4 32 43 PM
wxGenericMessageDlg:
Screenshot 2025-05-05 at 4 32 07 PM

PR #1477, opened 2025-04-01
Removed 'const' attribute from IsYubiInserted

huven:
The IsYubiInserted de-initializes the object if no key is inserted, which is IMHO not to be expected from a const method.

With this PR applied, it is (more) clear that the object changes significantly.

The 'mutable' on the ivar allowed this behavior, removed that keyword too.

### DIFF
--- a/src/os/unix/PWYubi.cpp
+++ b/src/os/unix/PWYubi.cpp
@@ -43,7 +43,7 @@ PWYubi::~PWYubi()
   pthread_mutex_unlock(&s_mutex);
 }
 
-bool PWYubi::IsYubiInserted() const
+bool PWYubi::IsYubiInserted()
 {
   bool retval = false;
   pthread_mutex_lock(&s_mutex);
--- a/src/os/unix/PWYubi.h
+++ b/src/os/unix/PWYubi.h
@@ -16,7 +16,7 @@ class PWYubi {
 public:
   PWYubi();
   ~PWYubi();
-  bool IsYubiInserted() const;
+  bool IsYubiInserted();
   bool GetSerial(unsigned int &serial) const;
   bool WriteSK(const unsigned char *sk, size_t sklen);
 
@@ -38,7 +38,7 @@ public:
   const std::wstring &GetErrStr() const {return m_ykerrstr;}
 private:
   void report_error() const;
-  mutable bool m_isInited;
+  bool m_isInited;
   static bool s_yubiDetected;
   static pthread_mutex_t s_mutex;
   mutable std::wstring m_ykerrstr;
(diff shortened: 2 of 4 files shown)


huven:
Just as a side note, here is some 'interesting' behavior from the Swift compiler (https://www.swift.org/documentation/cxx-interop/):

```
The Swift compiler assumes that constant member functions do not mutate the instance that this points to. A violation of this assumption by a C++ member function could lead to Swift code not observing the mutation of the instance pointed to by this and using the original value of such instance for the rest of the Swift code execution.

C++ permits the mutation of mutable fields in constant member functions. Constant member functions in structures or classes with such fields still become nonmutating methods in Swift. Swift doesn’t know which constant functions mutate the object, and which don’t, so for the sake of better API usability Swift still assumes that such functions do not mutate the object. You should avoid calling constant member functions that mutate mutable fields from Swift, unless they’re explicitly annotated with a SWIFT_MUTATING macro.

The SWIFT_MUTATING macro allows you to explicitly annotate constant member functions that do mutate the object. Such functions then become mutating methods in Swift.
```

Just to avoid any misunderstanding: I am not saying that Password Safe should follow the guideline above, I just reference it here as it is good to know how at least one compiler is moving away from mutating const functions (which does make sense once they are called mutating const 😄 )

In the case of IsYubiInserted, the side-effect of de-initializing the object is obviously non-const, I think.

dskrvk:
Not a C++ expert, but based on [this answer](https://stackoverflow.com/a/2384201) it seems the previous behavior was deliberate and in line with intended usage of the `const`/`mutable` combo. As far as I can see, the only member that's modified is `m_isInited` which is not observable publicly through accessors (so we can consider the "logical constness" to be preserved).

huven:
There really is no "logical constness" here. Just as an example (there are more): if there is no key inserted the IsYubiInserted will return false (as it should), but it will also return false on the next attempt, even if there is a key inserted by then. So a call to the method impacts next calls.

Moreover, a const method is also expected to be thread safe, and the current implementation is far from being thread safe.

I do get why someone would like to annotate IsYubiInserted as const from a logical standpoint, and I'm sure the implementation can be fixed, but the current code cannot be marked as such.

ronys:
Hm,

- The current implementation should be thread safe - that's what the mutex is there for.
- The Windows version also has IsYubiInserted() marked as const with mutables. I'd rather keep them as similar as possible.
- Logically, this is a predicate that shouldn't change externally visible state, so I think const with mutable is correct here.
- If IsYubiInserted() returns true when the device has been removed, that's obviously a bug, and I don't think that this PR addresses it. Am I wrong?

huven:
Closing this PR. Fixing the IsYubiInserted logic is best addressed in a new one.

PR #1478, opened 2025-04-04
Replaced internal Windows random function with offically sanctioned one

ronys:
The function was added in Windows Vista, so we should be good for all supported versions of Windows.


src/os/windows/rand.cpp
```diff
@@ -9,40 +9,39 @@
...
-    }
+  // Check if BCryptGenRandom is available
+  HMODULE hBcrypt = (HMODULE)pws_os::LoadLibrary(TEXT("bcrypt.dll"), loadLibraryTypes::SYS);
+  if (hBcrypt == nullptr) {
+    return false; // BCryptGenRandom not available
   }
-  return (hLib != NULL && pfnGetRandomDataT != NULL);
+  // Check if BCryptGenRandom function is available
+  FARPROC pFunc = GetProcAddress(hBcrypt, "BCryptGenRandom");
+  if (pFunc == nullptr) {
```
dskrvk, line 32: Suggestion: remove this if block and replace line 34 with
```
return pFunc != nullptr;
```
so the `FreeLibrary` invocation is in a single place.

ronys: Thanks!

PR #1483, opened 2025-04-07
[wx] Remove an unnecessary  parameter that causes a debug message

nobugshere:
Using a debug builds of wxWidgets 3.2.4 and 3.2.7 on macOS:
Running pwsafe from the command line with the -d or -e options caused a debug message to pop-up because of an unnecessary, and inconsistent, parameter. (See screenshot.)  Click continue and the program continues normally.  As the message says, the fix is to remove it.

Screenshot 2025-04-06 at 6 31 44 PM


ronys:
Thanks!

PR #1484, opened 2025-04-08
[wx] Fix CryptKeyEntryDlg layout issue on Wayland

### LINKED ISSUE #1476: Using pwsafe -d filename.PSF results in obscured text input area
After upgrading to Ubuntu 24.04.2 from 22.04.5 the pwsafe -d filename.PSF command no longer provides a clear text input area in which to enter the decryption key. This was after upgrading pwsafe to 1.21.0 for ubuntu 24.04. Interestingly, the old version for Ubuntu 22.04 (1.20.0) still works correctly on this Ubuntu 24.04 machine.

Here is a screenshot of the old working version 1.20.0 ....

Image

And here is the version 1.21.0 upgraded version screenshot ....

Image

Thank you for any assistance with fixing the geometry of this dialog.

BTW, I am using Wayland/Gnome on Ubuntu 24.04.2

EDIT: the user can still enter the decryption text in the obscured area and it works correctly; however, the text input feedback is missing due to the obscurity of the text input field.

rafaelx:
Fixes #1476


ronys:
Thanks @rafaelx !

PR #1486, opened 2025-04-11
[macOS] Support launching the app by clicking a file

nobugshere:
On macOS, when an app is launched by double-clicking a file, the file name is not passed on the command line, as it is in Unix/Linux.  Instead, the file name(s) are passed in event messages which wxWidgets passes to the app by calling these Mac-only functions:
MacNewFile() - When the app icon itself is clicked to launch the program (No file name is passed.)
MacOpenFiles() - When one or more files are double-clicked (we only use the first one in the list)
This is also called if a file is double-clicked while the program is running, in which case it's handled the same as using the File->Open menu item.

However, the framework does not call those functions until OnInit() returns, so MacOpenFiles() does not have a chance to override the default file name displayed in the password entry dialog.

This PR splits OnInit() into two parts. When the first part (OnInit()) returns "true", the event is processed and one of the above functions will call the second part, FinishInit().  On non-macOS builds, OnInit() calls FinishInit() directly, so there should be no change in functionality.
 
Although there is no intended change in function for non-macOS builds, some code changes are in effect.  Some minimal testing has been done on Linux to check for regressions, including command line option processing.  

Testing was done on:
macOS M1Max Sequoia 15.4, Xcode 16.3, wx 3.2.4 and 3.2.7 (local builds)
Fedora 41 ARM64 VMware,      Xfce/X11,      gcc 14.2.1, wx 3.2.6, GTK 3.24.43
Fedora 41 ARM64 VMware,      KDE/Wayland,   gcc 14.2.1, wx 3.2.6, GTK 3.24.43
Ubuntu 24.04.2 ARM64 VMware, Gnome/Wayland, gcc 13.3.0, wx 3.2.4, GTK 3.24.41

EDIT: InitPart2() changed to FinishInit() per review comment.

### DIFF
--- a/src/ui/wxWidgets/PWSafeApp.cpp
+++ b/src/ui/wxWidgets/PWSafeApp.cpp
@@ -742,7 +742,7 @@ bool PWSafeApp::ActivateLanguage(wxLanguage language, bool tryOnly)
 // Not to be confused with the system tray (menu bar) icon.
 void PWSafeApp::MacReopenApp()
 {
-//  (void)wxMessageBox("We're here!", "MacReopenApp", wxICON_INFORMATION | wxOK );
+//  (void)wxMessageBox("We're here!", "MacReopenApp", wxOK );
   if (m_initComplete)
     GetPasswordSafeFrame()->UnlockSafe(true, false);
 }
@@ -763,22 +763,32 @@ void PWSafeApp::MacOpenFiles(const wxArrayString& fileNames) {
 
   // Really, we can only open one file at a time, so just use the first one.
   filename = fileNames[0];
-  if (m_initComplete) {
-    int rc = m_frame->Open(filename.c_str());
-    if (rc == PWScore::SUCCESS) {
-      m_frame->FinishGoodOpen();
+
+  // Only allow one pass through InitPart2. While SafeCombinationEntryDlg is displayed,
+  // clicking another file causes another pass through here.  In this situation, the mutex
+  // seems to only be advisory.
+  wxMutexLocker lock(m_MacFileEventMutex);
+  if (lock.IsOk()) {
+    if (m_initComplete) {
+      int rc = m_frame->Open(filename.c_str());
+      if (rc == PWScore::SUCCESS) {
+        m_frame->FinishGoodOpen();
+      }
+    } else {
+      m_core.SetCurFile(filename);
+      InitPart2();
     }
-  } else {
-    m_core.SetCurFile(filename);
-    InitPart2();
   }
 }
 
 // This is called when the app is started by itself (not by clicking a data file)
 // Just finish the init.
 void PWSafeApp::MacNewFile() {
-//  (void)wxMessageBox("We're here!", "MacNewFile", wxICON_INFORMATION | wxOK );
-  InitPart2();
+//  (void)wxMessageBox("We're here!", "MacNewFile", wxOK );
+  wxMutexLocker lock(m_MacFileEventMutex);
+  if (lock.IsOk()) {
+    InitPart2();
+  }
 }
 #endif // __WXMAC__
 
--- a/src/ui/wxWidgets/PWSafeApp.cpp
+++ b/src/ui/wxWidgets/PWSafeApp.cpp
@@ -141,9 +141,8 @@ static const wxCmdLineEntryDesc cmdLineDesc[] = {
    wxCMD_LINE_VAL_NUMBER, wxCMD_LINE_PARAM_OPTIONAL},
 #endif
   {wxCMD_LINE_PARAM, nullptr, nullptr, STR("database"),
-   wxCMD_LINE_VAL_STRING,
-   (wxCMD_LINE_PARAM_OPTIONAL | wxCMD_LINE_PARAM_MULTIPLE)},
-  {wxCMD_LINE_NONE, nullptr, nullptr, nullptr, wxCMD_LINE_VAL_NONE, 0}
+   wxCMD_LINE_VAL_STRING, wxCMD_LINE_PARAM_OPTIONAL},   // We only really use one param
+  wxCMD_LINE_DESC_END
 };
 
 #undef STR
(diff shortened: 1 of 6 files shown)


src/ui/wxWidgets/PWSafeApp.cpp
```diff
@@ -718,11 +736,62 @@ bool PWSafeApp::ActivateLanguage(wxLanguage language, bool tryOnly)
...
 #ifdef __WXMAC__
-// On macOS, this enables file unlock and UI restore upon left-click of the dock icon.
+// macOS specific functions for UI interactions...
+
+// This enables file unlock and UI restore upon left-click of the dock icon.
 // Not to be confused with the system tray (menu bar) icon.
 void PWSafeApp::MacReopenApp()
 {
-  GetPasswordSafeFrame()->UnlockSafe(true, false);
+//  (void)wxMessageBox("We're here!", "MacReopenApp", wxOK );
```
dskrvk, line 745: Do you still need this debug line?

src/ui/wxWidgets/PWSafeApp.cpp
```diff
@@ -718,11 +736,62 @@ bool PWSafeApp::ActivateLanguage(wxLanguage language, bool tryOnly)
...
+// InitPart2() does not create the PasswordSafeFrame, (i.e. the user clicked cancel)
+// the the framework will call OnExit() for us.
+void PWSafeApp::MacOpenFiles(const wxArrayString& fileNames) {
+  StringX filename;
+  if (fileNames.IsEmpty())
+    return;
+
+//  wxString msg = wxString::Format("MacOpenFiles entered with %zd names, the first one is: %s",
+//                                  fileNames.GetCount(), fileNames[0].c_str());
+//  (void)wxMessageBox("We're here!", msg, wxICON_INFORMATION | wxOK );
```
dskrvk, line 762: These too

src/ui/wxWidgets/PWSafeApp.cpp
```diff
@@ -718,11 +736,62 @@ bool PWSafeApp::ActivateLanguage(wxLanguage language, bool tryOnly)
...
+      m_core.SetCurFile(filename);
+      InitPart2();
+    }
+  }
+}
+
+// This is called when the app is started by itself (not by clicking a data file)
+// Just finish the init.
+void PWSafeApp::MacNewFile() {
+//  (void)wxMessageBox("We're here!", "MacNewFile", wxOK );
```
dskrvk, line 790: Another one

src/ui/wxWidgets/PWSafeApp.cpp
nobugshere: I guess not, but I couldn't have figured it out without them :-)  I'll remove them.

dskrvk (approved):

src/ui/wxWidgets/PWSafeApp.cpp
```diff
@@ -141,9 +141,8 @@ static const wxCmdLineEntryDesc cmdLineDesc[] = {
    wxCMD_LINE_VAL_NUMBER, wxCMD_LINE_PARAM_OPTIONAL},
 #endif
   {wxCMD_LINE_PARAM, nullptr, nullptr, STR("database"),
-   wxCMD_LINE_VAL_STRING,
-   (wxCMD_LINE_PARAM_OPTIONAL | wxCMD_LINE_PARAM_MULTIPLE)},
-  {wxCMD_LINE_NONE, nullptr, nullptr, nullptr, wxCMD_LINE_VAL_NONE, 0}
+   wxCMD_LINE_VAL_STRING, wxCMD_LINE_PARAM_OPTIONAL},   // We only really use one param
```
ronys, line 144: wxCMD_LINE_PARAM_MULTIPLE was a mistake, the comment is unnecessary.

src/ui/wxWidgets/PWSafeApp.cpp
```diff
@@ -496,12 +496,29 @@ bool PWSafeApp::OnInit()
...
+ * On Linux/Unix, just call the second part direct, so this is still like one big happy function.
+ */
+#ifdef __WXMAC__
+  return true;
+#else
+  return PWSafeApp::InitPart2();
+#endif //__WXMAC__
+}
+
+bool PWSafeApp::InitPart2() {
```
ronys, line 520: "InitPart2" is a confusing name, I think, esp. since there's no InitPart1, and there's no indication if there should be InitPart3 or not.
"FinishInit" would be better, I think.
Also, since this always returns true, it could be void, and the calling logic could be changed accordingly.

seamusdemora:
Please consider this question as coming from the ["peanut gallery"](https://idioms.thefreedictionary.com/peanut+gallery), but as I read it, I couldn't help but wonder ... : 

I _**assume**_ the purpose of this change is is to allow someone (on a Mac) to open a password database file (i.e. a `*.psafe3` file) with the mouse??  If so, that sounds fine, as long as the option to specify the database file in the "opening screen" is not going to be removed.  Please tell me that the existing interface is not being "thrown out"  ... 

Assuming I guessed the implications of this change correctly above, I cannot help but further wonder where it leads us. It ***seems to me*** (from the peanut gallery where I sit) that it adds a bit of complexity to the software due to `wxWidgets` not being "native" to the Mac. I've long harbored the idea that complexity should be avoided except when "necessary". I won't comment further as I have no useful insights into the "_necessar-iness_" of this change. You need not reply to this second question - but I would appreciate a response to the first :) 

Best Rgds,
~S

nobugshere:
@seamusdemora ,
Have no fear, the opening screen hasn't changed at all.  Before this change, if you were to select a file and use "open with...", for example, the app would launch but the name in the opening dialog would be that of the last one opened, not the one you selected.  With this change, the name of the selected file is pre-set in the box, but you can still use the drop-down list or type a different name, just as always.  

I suppose it does add a little complexity, but wxWidgets has the features to support it, I just made use of them.  It works this way on Windows and Linux, it just bugged me that I couldn't do it on the Mac!

PR #1489, opened 2025-04-16
Move control IDs into a single enum to avoid conflicts

### LINKED ISSUE #1480: [wx] Duplicate menu ID
While working on something else, I noticed this duplicate ID at PasswordSafeFrame.h:69. Their use doesn't seem to conflict, but it just doesn't look right.  Comments?
```
#define ID_EXPORTMENU 10013
#define ID_EXPORT2OLD1XFORMAT 10013

```

Would it make sense to just move all the #define IDs into the enum at line 156?

nobugshere:
Resolves #1480


nobugshere:
Is there a way to tell GitGardian it's wrong, or shall I revert the last commit?

ronys:
I've marked this as a false positive in GitGuardian's dashboard.

PR #1493, opened 2025-04-21
[SF1607] Fix some font scaling issues

ronys:


src/ui/Windows/Fonts.cpp
```diff
@@ -191,10 +191,25 @@ void Fonts::SetTreeListFont(LOGFONT *pLF, const int iPtSz)
   }
 
   if (iPtSz == 0) {
```
dskrvk, line 193: Maybe extract this if-else into a helper method that returns a scaled `LOGFONT`, so `SetTreeListFont` could remain focused on one thing - setting `m_pTreeListFont`?

src/ui/Windows/Fonts.cpp
```diff
@@ -221,10 +236,25 @@ void Fonts::SetAddEditFont(LOGFONT *pLF, const int iPtSz)
...
```
dskrvk, line 238: We could utilize the same proposed helper method here.

dskrvk (commented):
I don't know how feasible this is, but could we centralize all logic that sets font sizes and related sizes (like the toolbar height)? This is so that 1. we're sure that all existing font-related operations apply scaling, 2. any font manipulations added in the future would use the same logic.

ronys:
I've cherry-picked the first commit that deals with scaling fonts, as this seems both necessary and correct. Abandoning the other commits as they had no effect on the toolbar display issue.

pm-kan:
Looks like that merged changes are incompatible with `tweakFontSizes` in some cases. Build from commit 22612894 at Win10 22H2 on monitor with resolution 2560x1440 and 125% scaling uses too large font size for password field, tree (and maybe some other places) even after removing all font settings from pwsafe.cfg. Commenting `tweakFontSizes` fixes font size for me (haven't looked ad font scaling logic yet, so not sure, that this is correct "fix").

test

PR #1494, opened 2025-04-23
[wx] Rename Cancel button to Close in Add-Edit dialog when Db is read-only

### LINKED ISSUE #1469: Differences in Linux vs. Windows version in Edit/View entry
Password Safe on Windows:
1. Open database in read and write mode.
2. Right click on entry and Edit entry. At the bottom there are OK and Cancel button.
3. Open database in read only mode.
4. Right click on entry and View entry.
What happens? At the bottom there is Close button.

Password Safe on Linux.
1-4 The same steps as above.
What happens? At the bottom there is only Cancel button alone. Instead of Cancel there should be Close button.

Cancel = When data changes are possible in dialog and we want to cancel the changes.
Close = When data changes are not possible and closing down dialog is by using Close button.

My suggestion: On Linux in read-only database in View entry dialog rename Cancel button with Close button.

rafaelx:
Resolves #1469 for non-Mac builds.

### DIFF
--- a/src/ui/wxWidgets/AddEditPropSheetDlg.cpp
+++ b/src/ui/wxWidgets/AddEditPropSheetDlg.cpp
@@ -168,7 +168,15 @@ AddEditPropSheetDlg::AddEditPropSheetDlg(wxWindow *parent, PWScore &core,
   SetExtraStyle(wxWS_EX_VALIDATE_RECURSIVELY|wxWS_EX_BLOCK_EVENTS);
   wxPropertySheetDialog::Create( parent, id, caption, pos, size, style );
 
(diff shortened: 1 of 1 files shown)


dskrvk (approved):

nobugshere (approved):

ronys:
Thanks!

nobugshere:
Just for reference: This was apparently a solved problem for wxCANCEL, but they didn't fix it for wxCLOSE.  I'll try submitting a PR to wxWidgets.  If it is accepted then, someday maybe, this #ifdef can be removed.

See: https://github.com/wxWidgets/wxWidgets/issues/15678

PR #1495, opened 2025-04-27
Database format changes to support Passkeys

huven:
In this PR the list of new fields to support passkeys on multiple platforms.

Suggestions are more than welcome!

Using the fields in this list, I was able to register a passkey and use it to login on passkeys.io.

## Use cases

- App and Website Sign-ins
- Workplace Authentication
- Payment Verification

## Benefits

- Phishing Resistance. Unlike passwords, passkeys are never manually entered. This means attackers can't trick you into typing them on fake websites.
- Stronger Cryptography. Passkeys use public-private key pairs, meaning even if someone compromises a service, they only get access to the public key (which is useless without the private key stored on your device).
- Automatic Storage & Management. No need to remember them or deal with weak passwords. They’re stored securely in pwSafe.
- No Reuse Issues. People often reuse passwords across multiple sites, making them vulnerable to breaches. Passkeys are unique per account, eliminating that risk.


docs/formatV3.txt
```diff
@@ -628,6 +635,35 @@ files will impact performance and synchronization times, and could make the data
...
+
+    Passkey User Handle
+    https://www.w3.org/TR/webauthn-2/#user-handle
+
+    Passkey Algorithm ID = IANA COSE Algorithm Identifier
+    https://www.w3.org/TR/webauthn-2/#sctn-alg-identifier
+    Encode as signed (!) 32 bit little-endian value, example for ES256:
+    0xf9 0xff 0xff 0xff
+
+    Passkey Private Key
```
ronys, line 657: Since we will not export the private key, I see no advantage in keeping it in PEM format. DER is more compact and just as universal.

huven: Thanks, good suggestion, I've updated the PR.

ronys:
Can you elaborate on the PasswordSafe specific use cases:

1.How does a user create an entry with passkey credentials?
2. How does a user use PasswordSafe to authenticate with passkey credentials?

huven:
> Can you elaborate on the PasswordSafe specific use cases:
> 
> 1.How does a user create an entry with passkey credentials? 2. How does a user use PasswordSafe to authenticate with passkey credentials?

Good question, I don't think it has a straightforward answer since you will need a layer interfacing between the database and the Relying Party.

That extra layer most likely will be the OS on which Password Safe is running, but it could just as well be an individual application using a .psafe3 file for credential storage on its own or offering an API to integrate.

Let's take website login as example, as it is most widely used. The user uses a browser. It will depend on how passkey support is implemented in the browser. Does it integrate with the OS or does it manage passkeys on its own? If integrated in the OS, then the OS runs the show and PS needs to integrate there. The OS might present some or all of the UI, or delegate it to the integration. Similarly in case the browser runs the show, then PS needs to integrate there (assuming the browser offers an integration at all).

This PR focuses on the necessary extension of the database format. After that, core needs to be extended (I will send a separate PR for that). After that, the integration into the platform will be specific to each platform.

ronys:
The database format and core extensions are the simple (or at least, the straightforward) part of adding support for passkeys.

I think the main advantage of adding passkey support to the project is for giving cross-platform support for passkeys. That is, if a user shares a PasswordSafe database with her PC, tablet and phone, the same passkey can be used to sign in from all devices independently. From what I've read, this looks like the main pain point for using passkeys today.

My suggestion for moving forward: Merge this PR to main, along with core support for the new fields (and unit tests).
This can even be released, since there's no overhead in existing databases for unused fields. Perhaps Platform-specific support for creating and using passkeys should be in separate branches in the main repo (github.com/pwsafe/pwsafe), for better collaboration?

huven:
Absolutely, platform compatibility is the main reason for registering these new fields, otherwise I could just have used the 0xe0-... fields.

I expect the core changes to be ready in 1-2 days, will merge those changes into this PR so you can merge them both at once.

Thanks!

huven:
Added the changes to core for passkey support.

Of interest is the new VectorX type reusing the secure allocator from StringX to provide the same protection for binary data. It is a template allowing other base types, which is how I prefer it, but maybe you prefer to fix the base type to unsigned char. That is fine too of course, I can easily change that.

Note: this type could be used by the attachment code too for better protection, but let's keep that outside this PR.

ronys:
Spec and core code look fine to me.

Indeed, VectorX<> can and should be used in the attachment code, in a future PR.

Now that we have a foundation for passkeys, I'd like to consider this from the UX perspective.
How do you propose a passkey entry be created, represented and used?

Note that the existing Add/Edit tabs are mostly irrelevant: No password (!), TOTP, expiration, history, or password policy.

huven:
The way I see it, is that the passkey is extra, so the user probably already has an existing entry, with password/totp, and the passkey is just added to that entry. The password must remain operational, for instance when a platform/application is used that does not support passkeys.

In the iOS version, when an entry has a passkey I will show minimal information to the user, as pwSafe cannot initiate anything with this. But I'll provide some UI, e.g. to remove the passkey.

Would it help if I post a small video showing the current state of affairs in the iOS version? This is very much work in progress, so far from complete, but you can see some of the UI stuff done by iOS.

huven:
I have added a few screenshots to a [support article](https://support.huvisoft.dev/hc/en-us/articles/33494755659668) for the iOS version showing some of the UI involved.

Note: this article is work in progress, far from complete, I just link it here to give some idea of how the UX for creating, using, etc. could look like, although this is very platform dependent.

ronys:
Nice.

I've created #1499 to track related functionality that should be added to complete support for this, mostly platform-independent.

PR #1496, opened 2025-04-29
Improve CItemField::Get

huven:
This removes the temporary buffer improving

- performance, saves an unnecessary copy
- security, the temporary memory was not cleaned, potentially leaking sensitive data


ronys:
Nice - thanks!

PR #1498, opened 2025-05-01
[wx] Add password strength meter to Add/Edit dialog

### LINKED ISSUE #1237: Add password strenght meter
I am looking for example at the following password strength web site: https://passwordmeter.com/
I have typed in password "test" and show "very week" with 4%.
image

If I generated 100 character password in pwsafe and I get:
image

I have also seen some of password strength web pages to have estimated time password could be cracked. For example displaying "1 second", "1 day", "1 week", "1 month", "1 year" and "centuries".

May I suggest to add password strength meter into pwsafe? Something like:
image

And meter could be one of:
image

The idea is to tell user how strong is actually password they are selecting as password and make sure password they choose is really strong.

rafaelx:
The current algorithm for calculating password strength is quite simple.

Note: Files src/ui/wxWidgets/StrengthMeter.h and src/ui/wxWidgets/StrengthMeter.cpp have been added.

Resolves #1237

### DIFF
--- a/src/core/PWCharPool.cpp
+++ b/src/core/PWCharPool.cpp
@@ -20,7 +20,6 @@
 
 #include <cmath>
 #include <string>
-#include <unordered_set>
 #include <vector>
 
 #include <algorithm>
@@ -547,19 +546,20 @@ bool CPasswordCharPool::CheckMasterPassword(const StringX &pwd, StringX &error)
 
 double CPasswordCharPool::CalculatePasswordStrength(const StringX &password)
 {
-  if (password.empty()) {
+  auto length = password.length();
+
+  if (length == 0) {
     return 0.0; // Empty password has no strength
   }
 
   // Determine the character pool size
-  std::unordered_set<char> charSet(password.begin(), password.end());
   int poolSize = 0;
-
   bool hasLower = false, hasUpper = false, hasDigits = false, hasSpecial = false;
-  for (char c : charSet) {
-    if (std::islower(c)) hasLower = true;
-    else if (std::isupper(c)) hasUpper = true;
-    else if (std::isdigit(c)) hasDigits = true;
+  for (size_t i = 0; i < length; i++) {
+    charT c = password[i];
+    if (_istlower(c)) hasLower = true;
+    else if (_istupper(c)) hasUpper = true;
+    else if (_istdigit(c)) hasDigits = true;
     else hasSpecial = true;
   }
 
@@ -569,15 +569,11 @@ double CPasswordCharPool::CalculatePasswordStrength(const StringX &password)
   if (hasSpecial) poolSize += 32; // Special characters (approximation)
 
   // Calculate entropy
-  double entropy = password.size() * std::log2(poolSize);
-
-  // Normalize entropy to a range of 0 to 100
-  double maxEntropy = 100.0; // Threshold for max entropy
-  double normalizedStrength = (entropy / maxEntropy) * 100.0;
+  double entropy = length * std::log2(poolSize);
 
   // Cap the value between 0 and 100
-  if (normalizedStrength > 100.0) {
-    normalizedStrength = 100.0;
+  if (entropy > 100.0) {
+    entropy = 100.0;
   }
-  return normalizedStrength;
+  return entropy;
 }
--- a/src/core/PWCharPool.cpp
+++ b/src/core/PWCharPool.cpp
@@ -18,7 +18,9 @@
 #include "os/typedefs.h"
 #include "os/pws_tchar.h"
 
+#include <cmath>
 #include <string>
+#include <unordered_set>
 #include <vector>
 
 #include <algorithm>
@@ -541,4 +543,41 @@ bool CPasswordCharPool::CheckMasterPassword(const StringX &pwd, StringX &error)
     LoadAString(error, IDSC_PASSWORDPOOR);
     return false;
   }
+}
+
+double CPasswordCharPool::CalculatePasswordStrength(const StringX &password)
+{
+  if (password.empty()) {
+    return 0.0; // Empty password has no strength
+  }
+
+  // Determine the character pool size
+  std::unordered_set<char> charSet(password.begin(), password.end());
+  int poolSize = 0;
+
+  bool hasLower = false, hasUpper = false, hasDigits = false, hasSpecial = false;
+  for (char c : charSet) {
+    if (std::islower(c)) hasLower = true;
+    else if (std::isupper(c)) hasUpper = true;
+    else if (std::isdigit(c)) hasDigits = true;
+    else hasSpecial = true;
+  }
+
+  if (hasLower) poolSize += 26; // a-z
+  if (hasUpper) poolSize += 26; // A-Z
+  if (hasDigits) poolSize += 10; // 0-9
+  if (hasSpecial) poolSize += 32; // Special characters (approximation)
+
+  // Calculate entropy
+  double entropy = password.size() * std::log2(poolSize);
+
+  // Normalize entropy to a range of 0 to 100
+  double maxEntropy = 100.0; // Threshold for max entropy
+  double normalizedStrength = (entropy / maxEntropy) * 100.0;
+
+  // Cap the value between 0 and 100
+  if (normalizedStrength > 100.0) {
+    normalizedStrength = 100.0;
+  }
+  return normalizedStrength;
 }
--- a/src/ui/wxWidgets/StrengthMeter.cpp
+++ b/src/ui/wxWidgets/StrengthMeter.cpp
@@ -16,10 +16,10 @@ wxBEGIN_EVENT_TABLE(StrengthMeter, wxControl)
   EVT_SIZE(StrengthMeter::OnSize)
 wxEND_EVENT_TABLE()
 
-const int StrengthMeter::STRENGTH_MIN = 0;
-const int StrengthMeter::STRENGTH_WEEK = 33;
-const int StrengthMeter::STRENGTH_MEDIUM = 66;
-const int StrengthMeter::STRENGTH_MAX = 100;
+// const int StrengthMeter::STRENGTH_MIN = 0;
+// const int StrengthMeter::STRENGTH_WEAK = 33;
+// const int StrengthMeter::STRENGTH_MEDIUM = 66;
+// const int StrengthMeter::STRENGTH_MAX = 100;
 
 StrengthMeter::StrengthMeter(wxWindow* parent, wxWindowID id,
                              const wxPoint& pos, const wxSize& size)
--- a/src/ui/wxWidgets/StrengthMeter.cpp
+++ b/src/ui/wxWidgets/StrengthMeter.cpp
@@ -1,0 +1,134 @@
+/*
+ * Initial version created as 'StrengthMeter.h'
+ * by rafaelx on 2025-05-01.
+ *
+ * Copyright (c) 2019-2025 Rony Shapiro <ronys@pwsafe.org>.
+ * All rights reserved. Use of the code is allowed under the
+ * Artistic License 2.0 terms, as specified in the LICENSE file
+ * distributed with this code, or available from
+ * http://www.opensource.org/licenses/artistic-license-2.0.php
+ */
+
+#include "StrengthMeter.h"
+
+wxBEGIN_EVENT_TABLE(StrengthMeter, wxControl)
+  EVT_PAINT(StrengthMeter::OnPaint)
+  EVT_SIZE(StrengthMeter::OnSize)
+wxEND_EVENT_TABLE()
+
+const int StrengthMeter::STRENGTH_MIN = 0;
+const int StrengthMeter::STRENGTH_WEEK = 33;
+const int StrengthMeter::STRENGTH_MEDIUM = 66;
+const int StrengthMeter::STRENGTH_MAX = 100;
+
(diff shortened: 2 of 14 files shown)


rafaelx:
**No Password**
pwd_strength_indicator_nopwd

**Week Password**
pwd_strength_indicator_week

**Medium Password**
pwd_strength_indicator_medium

**Strong Password**
pwd_strength_indicator_strong

rafaelx:
**Bright Theme**
pwd_strength_indicator_week_bt

**Dark Theme**
pwd_strength_indicator_week_dt

ronys:
Nice.
I've always had a problem with these meters not checking for dictionary words, but in this case "better is the enemy of good enough".

It would be nice, though, to add the meter to where the master passphrase is set and changed.

src/core/PWCharPool.cpp
```diff
@@ -542,3 +544,40 @@ bool CPasswordCharPool::CheckMasterPassword(const StringX &pwd, StringX &error)
...
+  if (hasUpper) poolSize += 26; // A-Z
+  if (hasDigits) poolSize += 10; // 0-9
+  if (hasSpecial) poolSize += 32; // Special characters (approximation)
+
+  // Calculate entropy
+  double entropy = password.size() * std::log2(poolSize);
+
+  // Normalize entropy to a range of 0 to 100
+  double maxEntropy = 100.0; // Threshold for max entropy
+  double normalizedStrength = (entropy / maxEntropy) * 100.0;
```
dskrvk, line 576: Nit:
```suggestion
  double normalizedStrength = (entropy / maxEntropy) * maxEntropy;
```
Same on next lines

src/core/PWCharPool.cpp
```diff
@@ -542,3 +544,40 @@ bool CPasswordCharPool::CheckMasterPassword(const StringX &pwd, StringX &error)
...
+    return 0.0; // Empty password has no strength
+  }
+
+  // Determine the character pool size
+  std::unordered_set<char> charSet(password.begin(), password.end());
+  int poolSize = 0;
+
+  bool hasLower = false, hasUpper = false, hasDigits = false, hasSpecial = false;
+  for (char c : charSet) {
+    if (std::islower(c)) hasLower = true;
```
dskrvk, line 560: A relatively easy improvement would be to keep separate sets for each class of characters and calculate the entropy based on the number of distinct characters in each set rather than the maximum set sizes.

src/ui/wxWidgets/StrengthMeter.h
```diff
@@ -0,0 +1,61 @@
...
+
+#ifndef STRENGTH_METER_H
+#define STRENGTH_METER_H
+
+#include <wx/wx.h>
+
+class StrengthMeter : public wxControl {
+public:
+  static const int STRENGTH_MIN;
+  static const int STRENGTH_WEEK;
```
dskrvk, line 20:
```suggestion
  static const int STRENGTH_WEAK;
```

src/ui/wxWidgets/StrengthMeter.h
```diff
@@ -0,0 +1,61 @@
...
+  void SetLabels(const wxString& weak, const wxString& medium, const wxString& strong);
+
+private:
+  void OnPaint(wxPaintEvent& event);
+  void OnSize(wxSizeEvent& event);
+  wxColor GetColorForStrength() const;
+  wxString GetLabelForStrength() const;
+
+  bool HasStrength() const { return m_strength > STRENGTH_MIN; }
+  bool IsWeekStrength() const { return m_strength < STRENGTH_WEEK; }
```
dskrvk, line 47:
```suggestion
  bool IsWeakStrength() const { return m_strength < STRENGTH_WEAK; }
```

src/ui/wxWidgets/StrengthMeter.h
```diff
@@ -0,0 +1,61 @@
...
+  bool IsMediumStrength() const { return m_strength < STRENGTH_MEDIUM; }
```
dskrvk, line 48:
```suggestion
  bool IsMediumStrength() const { return m_strength >= STRENGTH_WEAK && m_strength < STRENGTH_MEDIUM; }
```
so that this check can be used independently of others (no assumption that the caller has already checked `IsWeekStrength`)

src/ui/wxWidgets/StrengthMeter.h
```diff
@@ -0,0 +1,61 @@
...
+ */
+
+#ifndef STRENGTH_METER_H
+#define STRENGTH_METER_H
+
+#include <wx/wx.h>
+
+class StrengthMeter : public wxControl {
+public:
+  static const int STRENGTH_MIN;
```
dskrvk, line 19: Nit: why not assign the values here, since these are constants?

src/core/PWCharPool.cpp
rafaelx: I removed this, because I think this is code from experimenting with the entropy calculation. The calculation result can essentially be used directly and doesn't really need to be converted, scaled, or normalized. The return value is limited to 100, which also isn't really necessary. However, this is done with the knowledge that at the UI level, the strength meter uses a value range from 0 to 100. According to various sources, an entropy value of approximately 70 or more already considers a password strong. Therefore, I thought that values ​​greater than 100 no longer provide any added value.

src/core/PWCharPool.cpp
rafaelx: I've aligned the implementation with that of CPasswordCharPool::CheckMasterPassword, which uses additional criteria for password validation. CheckMasterPassword could also use the result of CalculatePasswordStrength. A comment in this function indicates the use of an entropy calculation.

src/ui/wxWidgets/StrengthMeter.h
rafaelx: That's funny, although there's actually nothing wrong with having a strong week. :-)

src/ui/wxWidgets/StrengthMeter.h
rafaelx: Renamed

src/ui/wxWidgets/StrengthMeter.h
rafaelx: Changed as proposed.

src/ui/wxWidgets/StrengthMeter.h
rafaelx: Changed as proposed. - Thanks for the review.

rafaelx:
> Nice. I've always had a problem with these meters not checking for dictionary words, but in this case "better is the enemy of good enough".
> 
> It would be nice, though, to add the meter to where the master passphrase is set and changed.

I will add this to the corresponding dialogs.

src/ui/wxWidgets/StrengthMeter.cpp
```diff
@@ -16,10 +16,10 @@ wxBEGIN_EVENT_TABLE(StrengthMeter, wxControl)
...
 wxEND_EVENT_TABLE()
 
-const int StrengthMeter::STRENGTH_MIN = 0;
-const int StrengthMeter::STRENGTH_WEEK = 33;
-const int StrengthMeter::STRENGTH_MEDIUM = 66;
-const int StrengthMeter::STRENGTH_MAX = 100;
+// const int StrengthMeter::STRENGTH_MIN = 0;
+// const int StrengthMeter::STRENGTH_WEAK = 33;
+// const int StrengthMeter::STRENGTH_MEDIUM = 66;
+// const int StrengthMeter::STRENGTH_MAX = 100;
```
dskrvk, line 22: Remove?

dskrvk (approved):

src/ui/wxWidgets/StrengthMeter.h
```diff
@@ -44,8 +44,8 @@ class StrengthMeter : public wxControl {
   wxString GetLabelForStrength() const;
 
   bool HasStrength() const { return m_strength > STRENGTH_MIN; }
-  bool IsWeekStrength() const { return m_strength < STRENGTH_WEEK; }
-  bool IsMediumStrength() const { return m_strength < STRENGTH_MEDIUM; }
+  bool IsWeekStrength() const { return m_strength < STRENGTH_WEAK; }
+  bool IsMediumStrength() const { return m_strength >= STRENGTH_WEAK && m_strength < STRENGTH_MEDIUM;; }
```
ronys, line 48: extra ';' at the end should be removed

src/ui/wxWidgets/StrengthMeter.h
```diff
@@ -44,8 +44,8 @@ class StrengthMeter : public wxControl {
   wxString GetLabelForStrength() const;
 
   bool HasStrength() const { return m_strength > STRENGTH_MIN; }
-  bool IsWeekStrength() const { return m_strength < STRENGTH_WEEK; }
-  bool IsMediumStrength() const { return m_strength < STRENGTH_MEDIUM; }
+  bool IsWeekStrength() const { return m_strength < STRENGTH_WEAK; }
```
ronys, line 47: s/IsWeekStrength/IsWeakStrength/g

ronys (commented):
Aside from the typo comments:
1. If the password make the entry an alias ([g:t:u]), then I think that the meter shouldn't be displayed at all, as it's meaningless and possibly confusing.
2. If I understand the code correctly, non-English alphabet characters are treated as 'special'. Can we do better than this?

nobugshere:
It's a nit-pick, but the strength meter doesn't line-up with the right end text box.
macOS M1Max Sequoia 15.4.1,  Xcode 16.3,    wx 3.2.4 (local build)

Screenshot 2025-05-05 at 10 35 00 PM

ronys:
Looks fine to me on Debian.
@rafaelx , @nobugshere - what do you prefer: Merging this as-is and addressing the Mac nit separately, or addressing this before the merge? I'm afraid I can't help, as I don't have access to a Mac.

nobugshere:
I'm okay either way.  I'll take a closer look at it sometime soon.

nobugshere:
I also suggest adding it to the "Set Master Password" dialog.  (When creating a new DB.)

rafaelx:
Some hints in case someone wants to optimize the alignment of the strength meters with the text input field.
The password input field and the eye icon are combined into one control element. If the strength meter is added to the layout, it expands to the width of the password input field plus the eye icon.

The eye icon has a fixed width of 29 pixels, the other horizontal spacing appears to vary slightly between platforms.
The current solution adds a fixed width spacer to the right of the strength meter. The width is calculated as 29 pixels for the eye icon plus 10/11 pixels for the spacing and additional 10/9 pixels for the Mac version.

nobugshere:
It looks better now.  
Screenshot 2025-05-07 at 2 39 37 PM

rafaelx:
Is the width in the Set Master Password dialog also correct?

The background color of the strength meter is very dark, looking black. My goal was to match the color of the text input fields for a dark desktop theme. Perhaps there's a way to determine this color in wx.

rafaelx:
One more thought: Is the layout and size correct even with the Yubikey elements present?

nobugshere:
> Is the width in the Set Master Password dialog also correct?
> One more thought: Is the layout and size correct even with the Yubikey elements present?

Yes and yes.  But, I'm not sure your patch for the background color is quite what you were looking for.  Here are a few screenshots with #f9848e1 applied.  Hmmm...I just noticed the Yubikey buttons are a different size on the "Change Master Password" dialog.  It's not that way on a build from master. (They both are the smaller size.) I'm not sure why that would have changed.  

Screenshot 2025-05-08 at 6 22 32 PM
Screenshot 2025-05-08 at 6 22 59 PM
Screenshot 2025-05-08 at 6 23 28 PM
Screenshot 2025-05-08 at 6 24 06 PM
Screenshot 2025-05-08 at 6 24 22 PM

rafaelx:
I have activated 5 pixel vertical and horizontal spacing in the flex grid sizer and I am not sure how this affects the width of the strength meter on the Mac. Perhaps the additional spacing recently added for the Mac needs to be adjusted?!

nobugshere:
The strength bar looks the same.  And now the Yubikey buttons match.
Screenshot 2025-05-11 at 11 05 48 AM

rafaelx:
> The strength bar looks the same. And now the Yubikey buttons match. Screenshot 2025-05-11 at 11 05 48 AM

Unfortunately, the strength meter does not have the same background color as under Linux, but I still find it visually acceptable.

rafaelx:
> Aside from the typo comments:
> 
> 1. If the password make the entry an alias ([g:t:u]), then I think that the meter shouldn't be displayed at all, as it's meaningless and possibly confusing.
> 2. If I understand the code correctly, non-English alphabet characters are treated as 'special'. Can we do better than this?

1. The last commit refers to the first comment.
2. I think this (the character set to be considered) depends on the locale (LC_CTYPE) set. For a more precise calculation of the entropy, the locale might also have to be taken into account, since the maximum number (see pool size in CPasswordCharPool::CalculatePasswordStrength) of possible alphabetic characters can vary depending on it.

nobugshere:
@rafaelx 
>The background color of the strength meter is very dark, looking black. My goal was to match the color of the text input fields for a dark desktop theme. Perhaps there's a way to determine this color in wx.

A little late, but I got around to playing with the background color.  I think I get closest to what you want by not setting the brush background at all in dark mode (commented dc.SetBrush).  It's a little darker than the text box, but lighter than the window background. 
Screenshot 2025-05-12 at 7 09 57 PM

On Ubuntu that made little if any difference.  It looked about the same as the grayed-out auth code box.

Issue #1499, opened 2025-05-03
Passkey support - followup


ronys:
The following should be done as a continuation of #1495 

1. Since all fields 0x2a..0x2f are required for passkeys, it would be nice to check for all present when opening a database. That is, either all are present or none.
2. A new filter for selecting entries that ether have or don't have passkeys
3. A graphic representation (new icon? new color?) for entries with passkeys
4. Support for import/export of XML
5. Support for import/export of text.

rafaelx:
Just a note: Compiling ItemData.cpp now causes the following warning.

```
/pwsafe/src/core/ItemData.cpp: In member function ‘bool CItemData::HasIncompletePasskey() const’:
/pwsafe/src/core/ItemData.cpp:2566:36: warning: comparison of integer expressions of different signedness: ‘int’ and ‘std::array<int, 6>::size_type’ {aka ‘long unsigned int’} [-Wsign-compare]
 2566 |     return !(numSet == 0 || numSet == fields.size());
      |                             ~~~~~~~^~~~~~~~~~~~~~~~

```

ronys:
Warning squashed in commit b20607417

PR #1500, opened 2025-05-03
Addresses first item of #1499

huven:
The code detects invalid (i.e. incomplete) passkeys, does not try to fix them.

For most missing fields, the only way to fix is to clear all passkey fields, which is probably better left to the user.


src/core/PWScore.cpp
```diff
@@ -2528,6 +2545,18 @@ bool PWScore::Validate(const size_t iMAXCHARS, CReport *pRpt, st_ValidateResults
         pRpt->WriteLine(cs_Error);
       } );
     }
+
+    // Passkey-related issues
+    if (st_vr.num_incomplete_passkeys > 0) {
```
dskrvk, line 2550: Could we define a separate method to write passkey issues to `pRpt`?

src/core/PWScore.cpp
```diff
@@ -2262,6 +2267,18 @@ bool PWScore::Validate(const size_t iMAXCHARS, CReport *pRpt, st_ValidateResults
       }
     }
 
+    // Test if passkey is complete
+    bool hasAllPasskeyFields = ci.IsPasskeyCredentialIDSet() && ci.IsPasskeyRelyingPartyIDSet()
```
dskrvk, line 2271: Could we define separate methods to calculate hasAllPasskeyFields and hasNonePasskeyFields? In general, `Validate` is quite long already, and it would be good to extract some of the logic.

huven: I've moved the logic to CItemData, see updated commit.

Better like this?

src/core/ItemData.cpp
```diff
@@ -2543,6 +2544,21 @@ void CItemData::SetPasskeySignCount(const uint32 sign_count) {
     CItem::SetField(PASSKEY_SIGN_COUNT, buf, 4);
 }
 
+bool CItemData::HasIncompletePasskey() const {
+    std::array<int, 6> fields = {
```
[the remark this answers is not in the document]
huven: Good point. I'll add a constexpr too, just for completeness.
https://lemire.me/blog/2023/04/12/consider-using-constexpr-static-function-variables-for-performance/

src/core/PWScore.cpp
huven: Yeah, will look at that tomorrow, separating sounds like a good idea.

huven: I've moved the report formatting to Report.cpp.

Let me know if you like it.

ronys (approved):
Looks good to me. If no one else has any comments, I'll merge this in tomorrow.

dskrvk (approved):

PR #1501, opened 2025-05-08
[wx] Disable menu and toolbar items when database is locked

### LINKED ISSUE #1227: "Write error" with "Unkown Error" text when database is locked and Manage | Options settings are changed.
Password Safe v1.9-pre from master branch on Ubuntu 23.10.

1. Manage | Options | Security.
2. Check the "Lock password database after" and type in 1 into "minutes idle" input box and close dialog with OK.
3. Wait a minute. Password database is automatically locked down as expected.
4.  Manage | Options | Security and change the same setting to 2 min and press OK button. What happens, pretty strange error: "Write Error" with text "/home/user/.pwsafe/database.psafe3. "Unknown Error" with OK button.

image

I am not sure why this error? Is because database is locked no options can be saved into locked database? If yes, then probably if database is locked Manage | Options should be greyed out until database is unlocked.

### LINKED ISSUE #1228: "Write error" with "Unkown Error" text when database is locked and File | Properties name and descriptions are changed
Password Safe v1.19-pre from master build on Ubuntu 23.10

1. Click Lock button to lock database.
2. File  | Properties.
3. Click three dots next to Name or three dots next to Description.
4. Click on Close button. Strange error appears: "Write Error" with "Unknown Error" appears.

It looks like if database is locked then File | Properties can not be changed. Both three buttons should be greyed out if database is Locked.

rafaelx:
Resolves #1227 
Resolves #1228


src/ui/wxWidgets/PasswordSafeFrame.cpp
```diff
@@ -2215,219 +2229,236 @@ void PasswordSafeFrame::OnUpdateUI(wxUpdateUIEvent& evt)
...
     case ID_RESTORE:
+    case ID_GENERATEPASSWORD:
 #ifndef NO_YUBI
     case ID_YUBIKEY_MNG:
 #endif
+      evt.Enable(isUnlocked && m_core.IsDbFileSet());
+      break;
+
+    case wxID_PROPERTIES:
       evt.Enable(m_core.IsDbFileSet());
```
nobugshere, line 2264: I wonder if properties should also be disabled if locked?  I see the name and description are disabled, would anything else be considered sensitive?  The DB ID, perhaps?

src/ui/wxWidgets/PasswordSafeFrame.cpp
```diff
@@ -122,6 +122,8 @@ BEGIN_EVENT_TABLE( PasswordSafeFrame, wxFrame )
   EVT_MENU( wxID_EXIT,                  PasswordSafeFrame::OnExitClick                   )
 
   // Update of menu items
+  EVT_UPDATE_UI( wxID_NEW,              PasswordSafeFrame::OnUpdateUI                    )
+  EVT_UPDATE_UI( wxID_OPEN,             PasswordSafeFrame::OnUpdateUI                    )
```
nobugshere, line 126: We need an EVT_UPDATE_UI entry for ID_PASSWORDQRCODE (the OnUpdateUI function already handles it.) 

ID_TOOLBAR_CLASSIC and ID_TOOLBAR_NEW need both.

src/ui/wxWidgets/PasswordSafeFrame.cpp
```diff
@@ -2464,27 +2495,27 @@ void PasswordSafeFrame::OnUpdateUI(wxUpdateUIEvent& evt)
...
 
     case ID_SHOWHIDE_TOOLBAR:
-      evt.Check(GetMainToolbarPane().IsShown());
+      evt.Check(isUnlocked && GetMainToolbarPane().IsShown());
       break;
 
     case ID_SHOWHIDE_DRAGBAR:
-      evt.Check(GetDragBarPane().IsShown());
+      evt.Check(isUnlocked && GetDragBarPane().IsShown());
       break;
```
nobugshere, line 2519: I don't know if you noticed, these are Check() not Enable().  I think a calls to Enable(isUnlocked) should be added.

nobugshere (changes requested):

src/ui/wxWidgets/PasswordSafeFrame.cpp
rafaelx: I wasn't sure about this. I decided that at least some of the database properties should be visible to the user. I write "some properties" because when the database is locked, the statistics show that there are no entries in the database, which is incorrect. All other information seems to be correct. Perhaps access to the database properties should also be disabled to avoid confusing the user with incorrect information? Maybe that makes more sense?

The idea behind these changes is to prevent all functionality that allows changes to the database when it is locked.
As mentioned in #1228, changing the password while the database is locked can even result in data loss.

UI-only changes, such as showing/hiding the toolbar, that are saved in the configuration file can still be allowed.

I'll look at the rest of the comments.

nobugshere: > The idea behind these changes is to prevent all functionality that allows changes to the database when it is locked.

Fair point.  Maybe what needs to be concealed when the DB is locked should be a separate discussion.

> UI-only changes, such as showing/hiding the toolbar...

Another fair point.  However, the menu items I mentioned don't seem to be fully functional while the DB is locked. (At least, not on a Mac.)  I would disable them for that reason (or make them work.)

rafaelx: I've removed the newly introduced code that disables the buttons for editing the database name and description in the properties dialog. The menu item for accessing the properties dialog is disabled when the database is locked, thus preventing access to the dialog.

src/ui/wxWidgets/PasswordSafeFrame.cpp
rafaelx: I've added the ID. Thanks for the hint.

src/ui/wxWidgets/PasswordSafeFrame.cpp
rafaelx: I have corrected this so that the menu items work again and deactivating/activating the corresponding bar is possible again.

nobugshere (approved):
Works well on a Mac now.  Thanks!

rafaelx:
Even if certain functionality can no longer be executed because the corresponding controls are disabled, it would still be interesting to find out why the database entries disappear if the password is changed when the database is locked. This may have uncovered a weakness in the core. Would this perhaps be worth a separate ticket/issue for investigation?

src/ui/wxWidgets/PasswordSafeFrame.cpp
```diff
@@ -925,6 +938,10 @@ void PasswordSafeFrame::RefreshToolbarButtons()
         toolbar->AddSeparator()->SetId(PwsToolbarButton.id);
     }
     else {
+      if ((PwsToolbarButton.id == ID_LOCK_SAFE) && m_sysTray->IsLocked())
+        continue; // When database is locked don't add the lock icon...
+      if ((PwsToolbarButton.id == ID_UNLOCK_SAFE) && !m_sysTray->IsLocked())
+        continue; // accordingly, don't add the unlock icon when database is unlocked
```
dskrvk, line 944: Not sure I see the point of an Unlock button since the menubar is inaccessible behind the unlock modal. Alternatively, we could change the button's behavior to be a toggle and have it change the text according to current DB state.

src/ui/wxWidgets/PasswordSafeFrame.cpp
```diff
@@ -2750,6 +2777,7 @@ void PasswordSafeFrame::UnlockSafe(bool restoreUI, bool iconizeOnCancel)
   }
 
   CreateMenubar(); // Recreate menubar to replace menu item 'Unlock' by 'Lock'
+  RefreshToolbarButtons();
```
dskrvk, line 2780: It seems strange to me that we have to refresh the toolbar right after creating it. Why can't create take into account the lock status?

src/ui/wxWidgets/PasswordSafeFrame.cpp
```diff
@@ -2215,219 +2233,228 @@ void PasswordSafeFrame::OnUpdateUI(wxUpdateUIEvent& evt)
...
-      evt.Enable(!isFileReadOnly && m_core.IsDbFileSet());
+      evt.Enable(isUnlocked && !isFileReadOnly && m_core.IsDbFileSet());
 #endif
       break;
 
     case ID_PROTECT:
-      evt.Enable(!isFileReadOnly && pci && !pci->IsShortcut());
+      evt.Enable(isUnlocked && !isFileReadOnly && pci && !pci->IsShortcut());
       evt.Check(pci && pci->IsProtected());
       break;
```
dskrvk, line 2459: We can also use the `isUnlocked` variable below on lines 2468 and 2472 instead of `!m_sysTray->IsLocked()`.

src/ui/wxWidgets/PasswordSafeFrame.cpp
rafaelx: We have a Lock/Unlock menu item in the menu bar, and we have a Lock/Unlock button/tool ​​item in the toolbar, which is located below the menu bar.
When the database is locked, the application is minimized. In a Wayland environment, the application can be restored or brought to the foreground without the unlock dialog appearing. In this case the user should be able to unlock the database via the menu as well as the toolbar.
Depending on the database status (locked/unlocked), the corresponding menu item and toolbar item should have the appropriate label.

src/ui/wxWidgets/PasswordSafeFrame.cpp
rafaelx: _CreateMenubar_ creates/recreates the menu items of the menu bar, while _RefreshToolbarButtons_ updates the toolbar items. With the changes, not only the menu is now updated, but also the toolbar. (Please see also my [comment](https://github.com/pwsafe/pwsafe/pull/1501#discussion_r2085372554) above.)

src/ui/wxWidgets/PasswordSafeFrame.cpp
rafaelx: I have changed this as suggested.

dskrvk (approved):

PR #1502, opened 2025-05-08
Prevent crash on reading invalid safe data

huven:
This PR fixes two crash reports, so this does happen in the real world.

### DIFF
--- a/src/core/ItemData.cpp
+++ b/src/core/ItemData.cpp
@@ -2131,6 +2131,8 @@ bool CItemData::SetField(CItem::FieldType ft, const unsigned char* data, size_t 
       {
         uuid_array_t uuid_array;
         ASSERT(len == sizeof(uuid_array_t));
+        if (data == nullptr || len < sizeof(uuid_array_t))
+          return false;
         for (size_t i = 0; i < sizeof(uuid_array_t); i++)
           uuid_array[i] = data[i];
         SetUUID(uuid_array, ft);


src/core/ItemData.cpp
```diff
@@ -2131,6 +2131,8 @@ bool CItemData::SetField(CItem::FieldType ft, const unsigned char* data, size_t
       {
         uuid_array_t uuid_array;
         ASSERT(len == sizeof(uuid_array_t));
+        if (data == nullptr || len < sizeof(uuid_array_t))
```
ronys, line 2134: Why not check for len != sizeof(uuid_array_t) instead of len < sizeof(uuid_array_t) ?

This might prevent a crash, but we'll still end up with a database with a missing UUID. Worth another Validation check?

Any idea how we got this? The fix addresses the crash, but not the root cause...

huven: It was the minimal change to prevent the crash, I didn't want to drop data in case len > 16.

Extending the validation to try to recover the entry as good as possible is definitely a good idea. At this stage I just wanted to prevent the crash, as that locks the user out completely. Having one entry dropped is inconvenient, but better than being locked out completely.

I have no sample unfortunately, just the crash reports pointing to this code.

huven: The reports I received indicate that data==NULL btw for those two cases, so the len check is for future events.

PR #1505, opened 2025-05-11
Improve passkey doc

huven:
This documents how the stored value must be interpreted, to make sure all platforms work alike.

Reusing a value likely results in a login failure, so it is important that all implementations use the same logic.

### DIFF
--- a/docs/formatV3.txt
+++ b/docs/formatV3.txt
@@ -663,6 +663,8 @@ of the contents.
 
     Passkey Sign Count, encoded as unsigned 32 bit little-endian value
     https://www.w3.org/TR/webauthn-2/#sctn-sign-counter
+    The value stored in the database is the last-used value, so the next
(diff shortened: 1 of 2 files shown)


ronys:
Hm.

Do you think it makes sense to save the pwsafe file whenever Passkey Sign Count changes then?
What happens if the app crashes before the new value is saved?

huven:
Yes, the database needs to be saved on each usage of the passkey. In the iOS version I first save the database to disk and only if that is successful the app signs the request.

When there is synchronizations or offline usage in play, it can still happen that a count is re-used. In that case the other side will decide how to handle it (allowing access or not). Not a disaster, as the next attempt will keep increasing the counter.

huven:
The [referenced URL](https://www.w3.org/TR/webauthn-2/#sctn-sign-counter) explains the counter in a bit more detail.

PR #1506, opened 2025-05-13
[wx] Do not set background color explicitly for MAC

rafaelx:


rafaelx:
This change is inspired by the suggestion of @nobugshere in the [comment](https://github.com/pwsafe/pwsafe/pull/1498#issuecomment-2874923447) of #1498 

The background color was also changed slightly for Linux environments.
Currently, it is set to the background color of text input controls.
image

Now the background color is slightly lighter than that of the text input controls and slightly darker than the background color of the dialog. As a result, the strength meter no longer looks like a control element with which the user can interact directly and at the same time it still stands out slightly from the background of the entire dialog.
image

See screenshot from @nobugshere in [comment](https://github.com/pwsafe/pwsafe/pull/1498#issuecomment-2874923447) for Mac version.

PR #1508, opened 2025-05-17
Mac: Fallback to default locale instead of throwing

huven:
Throwing when setlocale fails is not really helping the user. This PR addresses that by providing fallback logic, inspired by the unix implementation.

Different from the unix implementation, there is no output written to the console, as I don't think many Mac users will look at that, and also, that's not something a core lib should be doing.

Should we store the failure in the Startup global so the application can query that and provide an alert to the user after launching?

### DIFF
--- a/src/os/mac/utf8conv.cpp
+++ b/src/os/mac/utf8conv.cpp
@@ -32,8 +32,17 @@ class Startup {
 public:
   Startup() {
     char *sl = setlocale(LC_ALL, "");
-    if (sl == NULL)
-      throw "Couldn't initialize locale - bailing out";
+    if (sl == NULL) {
+      // According to setlocale(3):
+      // An argument of "" will determine the name of the new locale taking into account the environment variables LANG and LC_*.  If these environment variables yield a locale that is invalid, NULL will be returned and the current locale will remain unchanged.
+      // Apparently these environment variables are setup incorrectly, let's try some sane defaults
+      sl = setlocale(LC_ALL, "C.UTF-8");
+      if (sl == NULL)
+        sl = setlocale(LC_ALL, "en_US.UTF-8");
+    }
+    // If s1 is still NULL, all three setlocale calls failed.
+    // According to setlocale(3): By default, C programs start in the "C" locale.
+    // Continue using the "C" locale. UTF8 will not work.
   }
 };
 


ronys:
Throwing an exception is probably a bad idea, as this is called before main(). Behavior in this case is "implementation defined".

I think it's better to fail completely if we can't set the default locale, using system calls that don't depend on runtime initializations to perform correctly. The assumption is that the system is so badly misconfigured that nothing can be assumed to work correctly.

If you think that a fatal error or severe warning can be displayed via UI in this case, we can try it, but I'd keep the write(STDER_FILENO, ...) code in there as a failsafe way to help debugging.

huven:
Most of the complexity arises due to initializing the locale (which can fail) in a static var.

Would it be a good idea to move this code to a pws_os::StartupLocale returning a bool to indicate success? Then this could simply be called from main() and in case of failure an error could be shown to the user in a dialog or on the console.

I see special init code for Windows in main, so that would be a good place to add this.

If you like this approach, I'll have a go at it. This would allow each application to choose how to respond to these errors instead of bailing out.

huven:
Forget my previous comment (I just deleted it), I need to think of something better..

huven:
The unix code looks interesting btw, it starts with

char *gl = setlocale(LC_ALL, nullptr);

so gl will point to the C locale, as that is the default libc starting point. Then, setlocale(LC_ALL, "") is tried and if that fails setlocale(LC_ALL, gl) is tried instead. And this last call will always succeed as gl points to the C locale, which was the default to start with, so it is probably just a successful no-op.

So, on Linux, the locale is either set to the user-preferred locale (assuming the call succeeds), or remains at C locale. The _exit(2) is not reached.

I copy/pasted the code to a mini-program and set corrupt environment variables on a Linux box to test and that confirms the above. The warning about UTF-8 is displayed in the end, but the program happily continues. To be sure I printed the value of sl at the end, which confirms that the C locale is active at that time

Screenshot 2025-05-18 at 22 35 30

huven:
I think adding a pws_os::SetLocale() returning a bool for success might be best, that can be added to utf8conv.h just for Mac/Unix, there is already a "#ifndef _WIN32" in there.

The implementation for this can be almost (if not fully) the same for both platforms, just trying the setlocale(LC_ALL, "") and return false if setlocale returned NULL.

Then call this method from main() and do a regular console error print and exit there, or whatever action is appropriate, which will also allow other platforms to act accordingly.

What do you think?

huven:
Just checked the Windows os folder, the issue is in there too, so maybe even better to forget about pws_os:: and put the setlocale(LC_ALL, "") in main() and act on it there? That would seem to work for all platforms with very little code.

ronys:
Looking at PWSafeApp::OnInit(), I see there's already __WXMAC__ -specific code for setting the locale, so perhaps this PR is redundant?

huven:
That definitely looks like some code is redundant. So what do you suggest, just dropping the Startup from utf8conv.cpp and adding a setlocale(LC_ALL, "") to main.cpp for the cli so it matches the logic in PWSafeApp::OnInit()?

nobugshere:
I think, to maximize portability, I wouldn't throw in static Startup nor try to use setlocale() before entering main(). I don't know any of the history that led to this, maybe people had to deal with broken platforms at various times.  As @huven pointes out, the Unix version seems overly complicated.

As for PWSafeApp::OnInit(), I just ran it in the Xcode debugger:  The setlocale() call in Startup() returns "C", as I would expect, but by the time it reaches OnInit, the first call (LC_CTYPE) returns "UTF-8".  Changing it to get LC_ALL returns "C/UTF-8/C/C/C/C".  So, wxWidgets is doing its own thing before calling OnInit.

huven:
How about

- Removing Startup for all platforms
- Introduce "Status pws_os::SetupLocale()" returning success/failure or possibly a value signaling UTF-8 not being supported, allowing all invocations to act appropriately (exit / alert / ignore)
- Call this method in all gui/cli/test applications in main/OnInit or wherever early in the launch phase

This keeps the logic centralized in the core lib making it easy to improve, and requiring future os's to provide an implementation.

ronys:
The whole idea behind Startup is to have one less thing to remember to call in init/main. I'd rather keep it.

PR #1509, opened 2025-05-19
macOS: Slight changes to the Options Prop Sheet

nobugshere:
I'd like to request some changes to the Options dialog on macOS.  
1. I've been doing some work involving the System properties and the horizontal scroll bar is getting annoying. 
- Set a fixed width big enough for all the icons.
2. In dark mode, the dark background around the page title doesn't look very good.
- Use the system default background.  That has the slight advantage of picking-up the desktop color tint. (Matching the rest of the dialog background.) It's a Mac thing.

The change is ifdef'ed, so it only affects macOS.

Current image:
Screenshot 2025-05-19 at 7 31 34 AM

Proposed version:
Screenshot 2025-05-19 at 7 13 20 AM
Screenshot 2025-05-19 at 7 12 42 AM

### DIFF
--- a/src/ui/wxWidgets/OptionsPropertySheetDlg.cpp
+++ b/src/ui/wxWidgets/OptionsPropertySheetDlg.cpp
@@ -277,6 +277,9 @@ void OptionsPropertySheetDlg::CreateControls()
 
 wxPanel* OptionsPropertySheetDlg::CreateHeaderPanel(wxWindow* parent, const wxString& title)
 {
+#ifdef __WXMAC__
+  auto headerPanel = new wxPanel(parent, wxID_ANY, wxDefaultPosition, wxSize(605, -1));
+#else
   auto headerPanel = new wxPanel(parent, wxID_ANY, wxDefaultPosition, wxDefaultSize);
   auto bgColor = headerPanel->GetBackgroundColour();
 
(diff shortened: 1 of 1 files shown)


ronys:
Odd. On Linux, the icons are displayed vertically, on the side, and also suffer from the annoying scrollbar:
image

I'd be glad to have this fixed for both Mac and Linux.

Likewise, perhaps skipping the color adjustment in dark mode can be done at runtime, leaving it in place for light mode (and making this change for all platforms, that is, without the ifdef)?

rafaelx:
Removing the heading would also be worth considering. The idea behind the heading is to still indicate to the user which option group is currently displayed, even if the corresponding icon is no longer visible in the list view of icons due to the size of the dialog. The goal was to visually display the heading in large letters on a field that stands out from the background or the underlying area containing all the controls.

For example, something like the following to also make it look more modern.
image

In the dark environment, it also pretty much matches the idea, but could be slightly darker on Linux.
image

Recently I made some color adjustments to the strength meter depending on the environment (dark / light). Maybe this could be useful for this.

The heading font doesn't look nice on the Mac and doesn't appear vertically centered.

nobugshere:
@ronys 
I noticed that different Linux distros and desktop environments have slight variations in look and color.  The only reason I ifdef'ed it was so it wouldn't cause issues for the (very) many combinations I don't have to test.  WX has several layout options available.  I took a quick look at them and this one looked the least bad :-)

@rafaelx 
Are you suggesting removing the heading entirely ?   If not, I can try to pick a better font.

I'll take another look at the various options in a few days.  If given a choice, would the consensus be for the having the icons on the top or the side?

ronys:
No strong preference for top vs. side. Windows has them on the side, FWIW.

PR #1512, opened 2025-05-23
Mfc strength meter

ronys:
Windows version, tracking @rafaelx 's work on #1498


rafaelx:
Is it not necessary to distinguish between a light and dark desktop theme for selecting the background color in the Windows version?

dskrvk (approved):

PR #1514, opened 2025-06-02
Improve access denied detection

huven:
This will show a more helpful error message ("can't open file" instead of "failure") to the user in case the file cannot be opened for whatever reason (could be permission issue for example).

Note: use a release build to test this change, otherwise an assert stops the program in an earlier stage.

### DIFF
--- a/src/core/PWSfile.cpp
+++ b/src/core/PWSfile.cpp
@@ -23,6 +23,7 @@ PWSfile *PWSfile::MakePWSfile(const StringX &a_filename, const StringX &passkey,
                               Asker *pAsker, Reporter *pReporter)
 {
   PWSfile *retval = nullptr;
+  FILE *fd = nullptr;
 
   if (mode == Read && !pws_os::FileExists(a_filename.c_str())) {
     status = CANT_OPEN_FILE;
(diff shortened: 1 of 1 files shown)


src/core/PWSfile.cpp
```diff
@@ -64,6 +65,14 @@ PWSfile *PWSfile::MakePWSfile(const StringX &a_filename, const StringX &passkey,
         ASSERT(0);
         //[[fallthrough]];
       case UNKNOWN_VERSION:
+        // Return more meaningful status code if read access to the file is denied
+        fd = pws_os::FOpen(a_filename.c_str(), _T("rb"));
```
dskrvk, line 69: I wonder if we can instead directly check for file readability, similar to `[[ -r $file ]]` in shell?

huven: That probably uses access(2) to perform the test. Doing that here would probably require OS-specific code, and catch just permission related errors.

 The current PR keeps it simple, does a bit more work than just access(2) which is why I put it inside the unhappy flow instead of replacing the "exists" test at the start of the method. With the added benefit that it really catches all situations where the file cannot be opened (link an intermittent network error).

PR #1518, opened 2025-06-11
[wx] Replace variable length array with unique_ptr

nobugshere:
This warning started to appear with Apple Clang 17 (Xcode 16):
```
.../Xcode/pwsafe/src/ui/wxWidgets/wxUtilities.cpp:558:24: warning: variable length arrays in C++ are a Clang extension [-Wvla-cxx-extension]
  558 |   unsigned char buffer[size];
      |                        ^~~~
.../Xcode/pwsafe/src/ui/wxWidgets/wxUtilities.cpp:558:24: note: read of non-const variable 'size' is not allowed in a constant expression
.../Xcode/pwsafe/src/ui/wxWidgets/wxUtilities.cpp:552:8: note: declared here
  552 |   auto size = itemAttachment.GetContentSize();
      |        ^
1 warning generated.
```

If I understand smart pointers properly, this should maintain the original intent while being standards compliant (c++11 and up)

### DIFF
--- a/src/ui/wxWidgets/wxUtilities.cpp
+++ b/src/ui/wxWidgets/wxUtilities.cpp
@@ -555,9 +555,9 @@ bool ImagePanel::LoadFromAttachment(const CItemAtt& itemAttachment, wxWindow* pa
     return false;
   }
 
-  unsigned char buffer[size];
+  std::unique_ptr<unsigned char[]> buffer(new unsigned char[size]);
 
-  if (!itemAttachment.GetContent(buffer, size)) {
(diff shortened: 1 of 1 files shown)


src/ui/wxWidgets/wxUtilities.cpp
ronys, line 1: Might as well make size in line 552 const auto instead of auto, while we're at it...

nobugshere: Good point...done.

PR #1519, opened 2025-06-12
[wx] Expand the text field on resize of AddEditPropSheetDlg

nobugshere:
I noticed this problem crept into 1.21: If I vertically stretch the add/edit dialog, it grows in the wrong place.

Version 1.21 and current master:
Screenshot 2025-06-12 at 12 28 55 PM
Version 1.20 and after this PR is applied to master:
Screenshot 2025-06-12 at 12 37 26 PM


rafaelx:
Oops, this probably crept in when the password strength meter and the text input field for the authentication code were added. Someone must have forgotten to update the stretchable row. :)

rafaelx:
The Additional Tab would be another potential candidate because of the added text input field for the authentication secret, but it doesn't seem to be the case.

nobugshere:
I'v gone through the other tabs and several other dialogs, both horizontal and vertical. So far, I haven't found any other problems. (I only checked on macOS.)

Issue #1516, opened 2025-06-07
Compilation seems to break with wxWidgets 3.3


frankspace:
I am using Artix linux, which seems to have jumped the gun on upgrading to wxWidgets 3.3.  Unfortunately, that led me to discover that attempting to compile pwsafe with wxWidgets 3.3 results in the following failure:

```
In file included from /usr/include/wx-3.3/wx/intl.h:18,
                 from /build/passwordsafe/src/pwsafe-1.21.0/src/core/StringX.cpp:19:
/usr/include/wx-3.3/wx/translation.h: In instantiation of ‘wxString wxUnderscoreWrapper(T) [with T = const wchar_t*]’:
/build/passwordsafe/src/pwsafe-1.21.0/src/core/StringX.cpp:159:7:   required from ‘void LoadAString(T&, int) [with T = std::__cxx11::basic_string<wchar_t>]’
  159 |   s = _(core_st[id]).c_str();
      |       ^
/build/passwordsafe/src/pwsafe-1.21.0/src/core/StringX.cpp:223:45:   required from here
  223 | template void LoadAString(stringT &s, int id);
      |                                             ^
/usr/include/wx-3.3/wx/translation.h:394:20: error: static assertion failed: See https://wxwidgets.org/help/msgid-literals or LITERALS-MSGID comment above.
  394 |     static_assert(!sizeof(T), "See https://wxwidgets.org/help/msgid-literals or LITERALS-MSGID comment above.");
      |                    ^~~~~~~~~
/usr/include/wx-3.3/wx/translation.h:394:20: note: the comparison reduces to ‘(8 == 0)’
make[2]: *** [src/core/CMakeFiles/core.dir/build.make:508: src/core/CMakeFiles/core.dir/StringX.cpp.o] Error 1
make[2]: *** Waiting for unfinished jobs....
make[1]: *** [CMakeFiles/Makefile2:393: src/core/CMakeFiles/core.dir/all] Error 2
make: *** [Makefile:166: all] Error 2
```

I looked at the link, but, unfortunately, I didn't understand any of it.  I tried a lot of things at random, none of which worked.

frankspace:
Artix is known sometimes to push updates that are still in testing on Arch, and I'm afraid they did actually replace it:

![Image](https://github.com/user-attachments/assets/912fd070-5172-47b4-b188-18ec766bbc24)

[...]

PR #1520, opened 2025-06-14
[wx] wxWidgets 3.3.0 compatibility 

nobugshere:
Related to #1516 

This is a first draft to fix some wx3.3.0 compatibility issues. The story so far...

1. Removed inappropriate use of _().  This fixed failures in coretest and pwsafe-cli-test.
2. Fixed a crash in PasswordSafeFrame::CreateMenubar().  It was removing items from a wxList while iterating over a reference to the same object.  I'm not sure why it works in 3.2 but not 3.3, but it's probably bad form in any case.
3. The default path style for the recent file list changed.  It turns out that in 3.2 the default didn't match the documentation, in 3.3 it does.

I'm marking it draft for now because I have only tested it on macOS with 3.3.0 and 3.2.8.  I wanted to put it out for review and more testing on Linux and with other languages.  Also, the first commit needs to be reverted before merging so 3.3.0 won't be used for production.

@frankspace: Do you still have a 3.3.0 package to build against?

### DIFF
--- a/src/ui/wxWidgets/MenuViewHandlers.cpp
+++ b/src/ui/wxWidgets/MenuViewHandlers.cpp
@@ -226,7 +226,7 @@ void PasswordSafeFrame::RunShowReport(int iAction)
   }
   else {
     wxString tcAction = CReport::ReportNames.find(iAction)->second;
-    wxMessageBox(_(tcAction) + _(L" file \'") + rpt.GetFileName() + _(L"' not readable"), _("View Report"), wxOK|wxICON_ERROR);
+    wxMessageBox(tcAction + _(L" file \'") + rpt.GetFileName() + _(L"' not readable"), _("View Report"), wxOK|wxICON_ERROR);
   }
 }
 
--- a/.github/workflows/macos-latest.yml
+++ b/.github/workflows/macos-latest.yml
@@ -29,7 +29,7 @@ jobs:
       # "WX_DEBUG_FLAG: --enable-debug --enable-debug_info"
       # "XCODE_CONFIG:  Debug"
       # "XCODE_CONFIG:  Debug-no-yubi"
-      WX_VERS:          3.2.4
+      WX_VERS:          3.3.0
       WX_SRC_DIR:       ${{ github.workspace }}/wxWidgets
       WX_BUILD_DIR:     ${{ github.workspace }}/wxWidgets/static-release
       WX_ARCH:          "arm64,x86_64"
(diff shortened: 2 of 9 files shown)


frankspace:
I'll give it a test later today!

pm-kan:
> Removed inappropriate use of _()

Looks like, that most of these changes should be reverted or implemented in other way. After these changes items in combo boxes in settings, messages in status bar, field names in dragbar tooltips, ... become untranslated.

nobugshere:
@pm-kan is correct.  Commit 4e7ea02 should be reverted for wx 3.2.x, but will still need to be re-worked for 3.3.x.  It may be as simple as replacing it with explicit calls to [wxGetTranslation()](https://docs.wxwidgets.org/3.3.0/group__group__funcmacro__string.html#ga49dcdc17d82a627b7a50d401f1a37d69), but I clearly need to do more research to ensure compatibility.

I believe the other commits in this PR are still correct.

I apologize for not understanding the ramifications better.

src/ui/wxWidgets/MenuViewHandlers.cpp
```diff
@@ -226,7 +226,7 @@ void PasswordSafeFrame::RunShowReport(int iAction)
   }
   else {
     wxString tcAction = CReport::ReportNames.find(iAction)->second;
-    wxMessageBox(_(tcAction) + _(L" file \'") + rpt.GetFileName() + _(L"' not readable"), _("View Report"), wxOK|wxICON_ERROR);
+    wxMessageBox(tcAction + _(L" file \'") + rpt.GetFileName() + _(L"' not readable"), _("View Report"), wxOK|wxICON_ERROR);
```
dskrvk, line 229: Concatenating localizable strings is an anti-pattern. This should use a template string like most other UI captions.

ronys:
> @pm-kan is correct. Commit [4e7ea02](https://github.com/pwsafe/pwsafe/commit/4e7ea029cec8bf2da7d4f2e0a813d28544f32e92) should be reverted for wx 3.2.x, but will still need to be re-worked for 3.3.x. It may be as simple as replacing it with explicit calls to [wxGetTranslation()](https://docs.wxwidgets.org/3.3.0/group__group__funcmacro__string.html#ga49dcdc17d82a627b7a50d401f1a37d69), but I clearly need to do more research to ensure compatibility.
> 
> I believe the other commits in this PR are still correct.
> 
> I apologize for not understanding the ramifications better.

No need to apologize - this is a learning experience for all of us.

Please submit a PR reverting the commit, as I've just released a Windows version, and would like to follow up with a new wx version in a week or so.

Thanks to @pm-kan for catching this!

src/ui/wxWidgets/MenuViewHandlers.cpp
ronys: Indeed. I'm certain we have more silly concatenations like this in the code.
We should really move to [positional formatting](https://www.geeksforgeeks.org/cpp/cpp-20-std-format/), which was added in C++20.

PR #1529, opened 2025-06-30
Revert "Remove _() where it's not appropriate"

nobugshere:
This reverts commit 4e7ea029cec8bf2da7d4f2e0a813d28544f32e92.
Partial reversion of PR #1520 

This will need to be revisited before migrating to wxWidgets 3.3.0 or later.

### DIFF
--- a/src/core/StringX.cpp
+++ b/src/core/StringX.cpp
@@ -156,7 +156,7 @@ template<class T> void LoadAString(T &s, int id)
   cs.LoadString(id);
   s = cs;
 #else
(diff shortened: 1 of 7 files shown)


nobugshere:
@ronys,
I don't know why the macOS workflow has been failing here, it works fine for me.

ronys:
Yes, the macOS workflow has been timing out on my builds as well. Copilot thinks that it's because create-dmg is called incorrectly in a headless environment, but I haven't had time to look into this yet.

PR #1545, opened 2025-07-15
[wx GH954] Fix input focus on unlock

### LINKED ISSUE #954: Input Focus in Linux does not return to the password List after autolock

rhendric:
Fixes #954.


ronys:
Thanks @rhendric

PR #1547, opened 2025-07-15
[wx] Flag text for translation

### LINKED ISSUE #1543: Searching - status message not translated (wxWidgets only)

When Searching finds more than 1 match, the word "matches" is not translated.
Example: 1/4 matches
This seems to affect all non-English languages.

The other possible results are translated correctly:
msgid "No matches found"
msgid "1 match"

I've tracked the bug down to src/ui/wxWidgets/PasswordSafeSearch.cpp file, line 753:
wxT(" matches");
perhaps it should be:
_(" matches");

Attached you can find some screenshots for German, hope it helps.

Image

Image

nobugshere:
Resolves #1543 

Questions:
1. Should the ".  " inserted at line 755 be moved to the strings passed as prefix arguments, so it can be accounted for in their translation?
2. Should the string arguments to wxCHECK_RET, and other debug macros, be flagged for translation?

### DIFF
--- a/src/ui/wxWidgets/PasswordSafeSearch.cpp
+++ b/src/ui/wxWidgets/PasswordSafeSearch.cpp
@@ -706,7 +706,7 @@ SearchPointer& SearchPointer::operator++()
     m_currentIndex++;
     if (m_currentIndex == m_indices.end()) {
       m_currentIndex = m_indices.begin();
-      PrintLabel(_("Search hit bottom, continuing at top").c_str());
+      PrintLabel(_("Search hit bottom, continuing at top.").c_str());
     }
     else {
       PrintLabel();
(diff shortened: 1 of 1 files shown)


ronys:
Answers:
1. I don't think the period makes a difference in the translation.
2. No. These are meant for developers, which are assumed to be proficient in technical English.

src/ui/wxWidgets/PasswordSafeSearch.cpp
```diff
@@ -750,7 +750,7 @@ void SearchPointer::PrintLabel(const TCHAR* prefix /*= 0*/)
     // need a const object so we get both args to distance() as const iterators
     const SearchIndices& idx = m_indices;
     m_label.Clear();
-    m_label << std::distance(idx.begin(), m_currentIndex)+1 << '/' << m_indices.size() << wxT(" matches");
+    m_label << std::distance(idx.begin(), m_currentIndex)+1 << '/' << m_indices.size() << _(" matches");
```
dskrvk: There's still an issue here. This should be a template string instead of concatenation so the translators can see the whole thing. Different languages have a different word order - it may not always be as simple as "x/y [matches]".

dskrvk (approved):

src/ui/wxWidgets/PasswordSafeSearch.cpp
nobugshere: I re-worked the "match" messages

PR #1570, opened 2025-07-27
[wx] [GH1559] Save xml and export reports to correct report filename

### LINKED ISSUE #1559: "Export to XML" performs correct export but the Report is "Export to Text" (wxWidgets)

When selecting "Save to Disk" the Report summary is appended to "Export Text_Report.txt" file instead of creating a "Export XML_Report.txt" file.
That's why the Report is not available under View > Reports > Export XML.

I've tracked the bug down to src/ui/wxWidgets/MenuFileHandlers.cpp, line 822:
rpt.StartReport(IDSC_RPTEXPORTTEXT, sx_temp.c_str());
The issue is that the StartReport function uses "IDSC_RPTEXPORTTEXT" parameter when exporting both to Text and XML.

Suggested change: Create switch for the ExportType template parameter when invoking the StartReport function.
Something like this:

1.
struct ExportFullText {
->
struct ExportFullText {
  static constexpr int TypeId = 1;

2.
struct ExportFullXml {
->
struct ExportFullXml {
  static constexpr int TypeId = 2;

3.
rpt.StartReport(IDSC_RPTEXPORTTEXT, sx_temp.c_str());
->
        switch (ExportType::TypeId) {
            case 1:
                rpt.StartReport(IDSC_RPTEXPORTTEXT, sx_temp.c_str());
                break;
            case 2:
                rpt.StartReport(IDSC_RPTEXPORTXML, sx_temp.c_str());
                break;
            default:
                <dummy_command>;
        }

Image

Image

Image

Image

ronys:
Resolves #1559

### DIFF
--- a/src/ui/wxWidgets/MenuFileHandlers.cpp
+++ b/src/ui/wxWidgets/MenuFileHandlers.cpp
@@ -647,6 +647,10 @@ struct ExportFullText
   static wxString GetTaskWord() {
     return _("export");
   }
+
+  static int GetReportNameId() {
+    return IDSC_RPTEXPORTTEXT;
+  }
 };
 
 struct ExportFullXml {
@@ -689,6 +693,10 @@ struct ExportFullXml {
   }
   static wxString GetTaskWord() {
     return _("export");
+  }
+
+    static int GetReportNameId() {
+    return IDSC_RPTEXPORTXML;
   }
 };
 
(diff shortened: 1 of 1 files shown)


src/ui/wxWidgets/MenuFileHandlers.cpp
```diff
@@ -820,7 +828,7 @@ void PasswordSafeFrame::DoExportText()
           newfile = fd.GetPath().c_str();
           CReport rpt;
 
-          rpt.StartReport(IDSC_RPTEXPORTTEXT, sx_temp.c_str());
+          rpt.StartReport(ExportType::GetReportNameId(), sx_temp.c_str());
           rpt.WriteLine(tostdstring(wxString(_("Exporting database: ")) << towxstring(sx_temp) << _(" to ") << newfile<< wxT("\r\n")));
```
dskrvk, line 833: Side note: another example of concatenating translatable strings which only works for languages with the same word order as English. This should've been a template string.

dskrvk (approved):

src/ui/wxWidgets/MenuFileHandlers.cpp
ronys: Added a FIXME comment so as not to forget this.

PR #1574, opened 2025-07-29
[wx] Various dialog help updates

### LINKED ISSUE #1565: Help topic missing for some dialogs (wxWidgets)

1. SafeCombinationChangeDlg dialog does not have the Help button.
Help topic for the dialog is defined in HelpMap.h, but the "Help" button needs to be added.
Such Help button is available when setting up Master password for a new DB (SafeCombinationSetupDlg).
The dialog is self-explanatory when YubiKey is not present.
But the info in Help would be particularly useful when having YubiKey and confused about how to set up Master password/YubiKey combinations.

2. CreateShortcutDlg dialog does have the Help button but it fails to display Help.
HelpMap.h needs to be updated.

3. AdvancedMergeOptions dialog does have the Help button but it fails to display Help.
HelpMap.h needs to be updated.

Image

Image
Image

nobugshere:
Resolves #1565

### DIFF
--- a/help/default/html/file_menu.html
+++ b/help/default/html/file_menu.html
@@ -97,11 +97,12 @@ correctly formatted text file, XML, or a KeePass V1 password
 database into the current database. See <a href="import.html">here</a> for
 more information on importing text files.</p>
 
-<h3>Merge</h3>
+<h3><a name="Merge"></a>Merge</h3>
 
 <p>Merge a <b>Password Safe</b> database or backup file (.bak) into the current database.
-Disabled in read-only mode.  Merge uses a Wizard interface. You may also select the Advanced
-field selection in order to restrict which entries you merge into your current database.</p>
+Disabled in read-only mode.  You may also select Advanced Marge Options in order to restrict 
+which entries are merged into your current database. To select a subset of entries to merge, 
+select a field, comparison mode, and enter the text to be compared. </p>
 
 <h3>Compare</h3>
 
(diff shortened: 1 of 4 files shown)


help/default/html/file_menu.html
```diff
@@ -97,11 +97,12 @@ <h3>Import From</h3>
 database into the current database. See <a href="import.html">here</a> for
 more information on importing text files.</p>
 
-<h3>Merge</h3>
+<h3><a name="Merge"></a>Merge</h3>
 
 <p>Merge a <b>Password Safe</b> database or backup file (.bak) into the current database.
-Disabled in read-only mode.  Merge uses a Wizard interface. You may also select the Advanced
-field selection in order to restrict which entries you merge into your current database.</p>
+Disabled in read-only mode.  You may also select Advanced Marge Options in order to restrict
```
ronys: Typo: s/Marge/Merge/
```suggestion
Disabled in read-only mode.  You may also select Advanced Merge Options in order to restrict 
```

ronys:
Hi @rafaelx 

The submitter of #1565 made a good point regarding the password change dialog:

> The dialog is self-explanatory when YubiKey is not present.
But the info in Help would be particularly useful when having YubiKey and confused about how to set up Master password/YubiKey combinations.

Of course, we could show the help button there only when a YubiKey is detected, but that's arguably less elegant/consistent.

rafaelx:
I should read #1565 first. The pull request immediately reminded me of the changes to the smaller dialogs.
I can imagine that this process of changing the YubiKey master password combination can lead to uncertainty about doing something wrong. The ability to learn about the procedure can give users a sense of confidence.

help/default/html/file_menu.html
nobugshere: Rats.  One always seems to get by...Thanks.

ronys:
I'm ambivalent about the Advanced Merge Options dialog. On the whole, I don't think the five or so people who actually reach that dialog will care much one way or the other...

PR #1578, opened 2025-07-31
[wx] Add a warning about clipboard history utilities

nobugshere:
Lately I've been thinking about clipboard history utilities.  Specifically, clearing the clipboard does not remove information from them.  I can't think of anything we can do about that, so I added a warning to the pop-up confirmation message and the help file.  I also changed the default for the DontAskQuestion preference so a new user, or one who has never changed that option, will see the message at least once.

Note: The meaning of DontAskQuestion is backward.  The code treats it as "Do Ask Question".  I didn't change the code or the parameter so as to remain compatible with existing config files.

Screenshot 2025-07-30 at 5 24 37 PM

### DIFF
--- a/help/default/html/edit_menu.html
+++ b/help/default/html/edit_menu.html
@@ -99,7 +99,8 @@ in reverse order, from last to earliest.</p>
 <p>Clear the clipboard's memory. This clears the clipboard of
 anything that was pasted there by <b>Password Safe</b> (password, username, notes, etc.).
 If the current contents of the clipboard did not come from <b>Password Safe</b>, then
-this will have no effect.</p>
+this will have no effect. <b>WARNING: Password Safe</b> can only clear the main clipboard.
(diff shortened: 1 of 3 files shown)


ronys (changes requested):
The Windows code treats the preference as it's name implies, so changing the default value to true breaks the "show first time" semantics on Windows.

Also, I'm not sure what the user is supposed to do with the information displayed in the warning - it's not "actionable", at least, not for the non-technical user.
Ideally, PasswordSafe would check (perhaps upon startup, or first copy?) if the clipboard history is kept or not and give a more focused warning accordingly.

nobugshere:
> The Windows code treats the preference as it's name implies, so changing the default value to true breaks the "show first time" semantics on Windows.

Oops.  Well, that complicates things.  

> Ideally, PasswordSafe would check....

I don't know a portable way to check for that or, for that matter, any platform specific ways.  I'll keep thinking about it.

PR #1580, opened 2025-08-01
Feature/non windows help

ronys:
Implements suggested change in #1579

PR #1581, opened 2025-08-01
[wx] Added wx-specific Slovak help files from Milan

ronys:
This works with PR #1580


help/pwsafeSK_wx/html/images/abase_exp.png
dskrvk, line 1: Wondering whether we could share non-translatable images between languages instead of copying them.

dskrvk (approved):

PR #1583, opened 2025-08-02
[macOS] CFStringGetCStringPtr can return NULL

### LINKED ISSUE #1582: App crashes on macOS when attaching an image other than JPG (wx)
Version: Password Safe 1.22 on macOS
Database format: 4.01

The app silently crashes on importing image in these formats: GIF/PNG/BMP (maybe others).
Other file formats like Text or a binary file work OK.
Tested on the latest macOS 12 and 15, there is no issue on Linux.
Strangely enough, if I change the extension of a GIF/PNG/BMP file to JPG or JPEG, it gets imported without any issue and a preview is available.

Attached you can find some screenshots, hope it helps (PNG image with .jpg extension).

Image
Image

nobugshere:
Resolves #1582

### DIFF
--- a/src/os/mac/media.cpp
+++ b/src/os/mac/media.cpp
@@ -41,7 +41,14 @@ stringT pws_os::GetMediaType(const stringT & sfilename)
         return unknown_type;
 
     // Convert mime_type to stringT
-    stringT retval = pws_os::towc(CFStringGetCStringPtr(mime_type, kCFStringEncodingUTF8));
+    CFIndex maxSize = CFStringGetMaximumSizeForEncoding(CFStringGetLength(mime_type), kCFStringEncodingUTF8) + 1;
+    auto utf8str = new char[maxSize];
+    auto success = CFStringGetCString(mime_type, utf8str, maxSize, kCFStringEncodingUTF8);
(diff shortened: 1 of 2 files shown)


src/os/mac/media.cpp
```diff
@@ -41,7 +41,14 @@ stringT pws_os::GetMediaType(const stringT & sfilename)
         return unknown_type;
 
     // Convert mime_type to stringT
-    stringT retval = pws_os::towc(CFStringGetCStringPtr(mime_type, kCFStringEncodingUTF8));
+    CFIndex maxSize = CFStringGetMaximumSizeForEncoding(CFStringGetLength(mime_type), kCFStringEncodingUTF8) + 1;
+    auto utf8str = new char[maxSize];
+    auto success = CFStringGetCString(mime_type, utf8str, maxSize, kCFStringEncodingUTF8);
+    if (success == false)
```
ronys: As @huven pointed out, mime_type and utf8str need to be released if !success.

nobugshere: Ouch! Bad on me :-(   But, I did find one other spot for a release.

PR #1587, opened 2025-08-05
[NSIS] Fix CLI uninstallation and updated Slovak translation

froggie2523:
1. Minor fixes to Slovak translation for NSIS installer.
2. pwsafe-cli.exe is not deleted during uninstallation on Windows, when selected to install via NSIS. This also prevents the "C:\Program Files\Password Safe" folder from being deleted from disk. See attached screenshots.

pwsafe_3 69-bug-CLI-01
pwsafe_3 69-bug-CLI-02


ronys:
Good catch - thanks!

PR #1588, opened 2025-08-05
[GH1557] Sanitize report texts before display

### LINKED ISSUE #1557: Blank window when viewing reports on Linux

After I finish an action like "Export to text", the report is displayed correctly and I can save it to disk.
When I try opening the report later under View > Reports, the report is not displayed on Linux, it works on macOS though.
I can delete the file from there, I can copy the text to clipboard and see it in a text editor.
Tested on the latest Debian/Fedora/Kubuntu, using Wayland/X.Org.

Image

Image

Image

ronys:
Resolves #1557

### DIFF
--- a/src/ui/wxWidgets/ViewReportDlg.cpp
+++ b/src/ui/wxWidgets/ViewReportDlg.cpp
@@ -37,7 +37,29 @@ ViewReportDlg::ViewReportDlg(wxWindow *parent, CReport* pRpt, bool fromFile) :
 
   wxBoxSizer* dlgSizer = new wxBoxSizer(wxVERTICAL);
 
-  wxTextCtrl* textCtrl = new wxTextCtrl(this, wxID_ANY, towxstring(pRpt->GetString()),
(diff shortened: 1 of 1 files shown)


src/ui/wxWidgets/ViewReportDlg.cpp
```diff
@@ -37,7 +37,29 @@ ViewReportDlg::ViewReportDlg(wxWindow *parent, CReport* pRpt, bool fromFile) :
...
+  try {
+    reportText = towxstring(reportString);
+  } catch (...) {
+    reportText.Clear();
+  }
+    
+
+  if (!reportText.IsEmpty()) {
+    wxString filteredText;
+    for (auto ch : reportText) {
```
rafaelx: Perhaps std::replace_if could be a one-line alternative to the classic for loop, which also highlights the intent of the filter through the name.

ronys: Of course! Thanks.

PR #1589, opened 2025-08-06
[i18n] view-report-fix (wx)

froggie2523:
After PR #1588 that fixed issue #1557 (Blank window when viewing reports on Linux), the Reports get displayed now but the program treats "non-English" chars incorrectly and replaces them by space.
This PR is to fix that by using locale-aware functions that are part of wxWidgets.
See attached screenshots.

Left image is OK. Right: German "ä ü ß" and Slovak "Ú č é" disappeared.

pwsafe_1 22-wxWidgets-bug-View_Report-04

pwsafe_1 22-wxWidgets-bug-View_Report-05

### DIFF
--- a/src/ui/wxWidgets/ViewReportDlg.cpp
+++ b/src/ui/wxWidgets/ViewReportDlg.cpp
@@ -50,7 +50,7 @@ ViewReportDlg::ViewReportDlg(wxWindow *parent, CReport* pRpt, bool fromFile) :
 
   if (!reportText.IsEmpty()) {
     std::replace_if(reportText.begin(), reportText.end(),
-                   [](wxUniChar ch) { return !isprint(ch) && !isspace(ch); },
+                   [](wxUniChar ch) { return !wxIsprint(ch) && !wxIsspace(ch); },
                    wxT(' '));
   }
   


ronys:
@rafaelx  - is your comment referring to the text before or after this PR?

src/ui/wxWidgets/ViewReportDlg.cpp
```diff
@@ -50,7 +50,7 @@ ViewReportDlg::ViewReportDlg(wxWindow *parent, CReport* pRpt, bool fromFile) :
 
   if (!reportText.IsEmpty()) {
     std::replace_if(reportText.begin(), reportText.end(),
-                   [](wxUniChar ch) { return !isprint(ch) && !isspace(ch); },
+                   [](wxUniChar ch) { return !wxIsprint(ch) && !wxIsspace(ch); },
```
rafaelx, line 53: Is the whitespace check by wxIsspace actually necessary if, according to the [isprint ](https://en.cppreference.com/w/c/string/byte/isprint) documentation, whitespace is already taken into account?

rafaelx: Oh, now I see that [isspace](https://en.cppreference.com/w/cpp/string/byte/isspace.html) also takes newline characters into account, if that was the intention.

ronys: > Oh, now I see that [isspace](https://en.cppreference.com/w/cpp/string/byte/isspace.html) also takes newline characters into account, if that was the intention.

Indeed that was the intention.

ronys:
> @ronys My comment refers to the general translation, regardless of the changes made by this PR.
> 
> Perhaps some textual aspects could be improved. For example, the title and button labels, but also the text of the report.
> 
> * Title "Show Report" - "Report Summary"?
> * Button "Remove from Disk" - "Remove"?
> * Button "Save to Disk" - "Save"?
> * Button "Copy to Clipboard" - "Copy"
> 
> "Disk" sounds a bit old-fashioned and somehow suggests a classic rotating hard disk. Maybe "disk" could be replaced as "file". In case of the shortened version of the button labels, these could have tooltips for more information.

Good points, worth a separate ticket - thanks

PR #1593, opened 2025-08-08
Import Text Report - Fix unknown headers formatting

froggie2523:
This PR is to fix formatting Report for the Import Text function, in case there are unknown headers in the import file.
The list of unknown columns is not formatted before displaying (no delimiter), so we need to add CRLF after each item as well.
Tested on Windows and Linux, it should work on macOS too.

See attached screenshots how it gets improved.
pwsafe_3 69-bug-ImportText-01
pwsafe_3 69-bug-ImportText-02


froggie2523:
Report after the change:
pwsafe_3 69-bug-ImportText-03

nobugshere (approved):

PR #1608, opened 2025-09-08
[wx] Add Tooltip to Options when System Tray support is unavailable

froggie2523:
This PR is to add Tooltip with details for the System Tray settings in Options when the feature is unavailable, providing hints about how to resolve the issue.

Like the AutoType feature, System Tray support in Linux/UNIX does not work on Wayland.
Additionally, System Tray icon is not available in Desktop Environments like Gnome on X11, unless user installs an extension that adds the feature back to the system (but who knows about this?).
While settings for AutoType shows a tooltip with details ("Not supported by Wayland"), the settings for System Tray provides no info on why the feature is unavailable.

Todo:
.PO files need to be updated with a translated message:
"Not supported by the current Windowing System (e.g. Wayland), or you may need to install a Desktop Environment extension to enable System Tray support."

Tested in:
-Debian/FreeBSD/OpenBSD with Gnome where X11 is the default one or it can be selected instead of Wayland on the login screen.
  Gnome extensions tested: "AppIndicator and KStatusNotifierItem Support" and "Tray Icons: Reloaded".
  Password Safe does not need to be restarted after the extension is installed and enabled.
-Mint with Cinnamon on Wayland/X11
-Kubuntu with Plasma on Wayland/X11

pwsafe_1 22-wxWidgets-feat-SystemTray-01

ToolTip added - Debian on Wayland:

pwsafe_1 22-wxWidgets-feat-SystemTray-02

ToolTip added - Mint on Wayland:

pwsafe_1 22-wxWidgets-feat-SystemTray-03

Debian on X11 with a Gnome extension for System Tray:

pwsafe_1 22-wxWidgets-feat-SystemTray-04

Mint on X11:

pwsafe_1 22-wxWidgets-feat-SystemTray-05

### DIFF
--- a/src/ui/wxWidgets/wxUtilities.cpp
+++ b/src/ui/wxWidgets/wxUtilities.cpp
@@ -355,6 +355,15 @@ void wxUtilities::DisableIfUnsupported(enum Feature feature, wxWindow* window)
   }
 }
 
+void wxUtilities::NotifyIfUnsupported(enum Feature feature, wxWindow* window)
+{
+#if defined(__linux__) || defined(__FreeBSD__) || defined(__OpenBSD__)
+  if (feature == SystemTray && !IsTaskBarIconAvailable()) {
+    window->SetToolTip(_("Not supported by the current Windowing System (e.g. Wayland), or you may need to install a Desktop Environment extension to enable System Tray support."));
+  }
+#endif
+}
+
 // Wrapper for wxTaskBarIcon::IsAvailable() that doesn't crash
 // on Fedora or Ubuntu
 bool IsTaskBarIconAvailable()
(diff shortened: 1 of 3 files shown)


src/ui/wxWidgets/wxUtilities.cpp
```diff
@@ -355,6 +355,15 @@ void wxUtilities::DisableIfUnsupported(enum Feature feature, wxWindow* window)
   }
 }
 
+void wxUtilities::NotifyIfUnsupported(enum Feature feature, wxWindow* window)
+{
+#if defined(__linux__) || defined(__FreeBSD__) || defined(__OpenBSD__)
```
ronys, line 360: Aren't the conditionals around IsTaskBarIconAvailable() sufficient for this?

nobugshere: Given that, might it also be better to move this specialized tool tip message to OptionPropertySheetDlg.cpp, keeping the more generic IsTaskBarIconAvailable() in wxUtilities?

froggie2523: As I see it, the conditionals in IsTaskBarIconAvailable() are for a specific issue with some version of wxWidgets, then wxTaskBarIcon::IsAvailable() is called.
The conditionals in NotifyIfUnsupported are about not displaying the ToolTip in macOS, I've seen this combo somewhere else in the code. 
Yes, I could use __APPLE__ or similar but I don't have full understanding of the macOS related conditionals, so this looks more appropriate to me.

Issue #1612, opened 2025-09-14
[doc] Update README.LINUX.md with instructions to enable System Tray support in Linux


froggie2523:
**I would like to update documentation with details on how to make the System Tray work on Linux.**
Since the Help files are copied from Windows, I think it is most appropriate to put such info into the README.LINUX.md file.

Issue: In many distros, Password Safe does not appear in the System Tray. The source code shows that Password Safe relies on wxWidgets to detect whether System Tray support is available in the OS. Once the feature is enabled at the system level, the "Put icon in System Tray" checkbox becomes available under Options > System.
Usually, this feature is not supported by the windowing system such as Wayland, or the user needs to install a desktop environment extension to enable System Tray support. I researched System Tray support in Linux, and it can be enabled even when running on Wayland or using Flatpak. On Wayland, it basically revolves around making Password Safe use the X11/Xorg backend through the XWayland compatibility layer.

**My question is: Taking into account the Security model of Wayland vs. X11 and the implications it might have, is it OK to publish such workaround?**

Tested in:
-Debian with Gnome on Wayland/X11
-Ubuntu with Gnome on Wayland/X11
-Ubuntu with Cinnamon on X11 (Wayland support is experimental - System Tray not working)
-Kubuntu with KDE Plasma on Wayland/X11
-Mint with Cinnamon and MATE on X11 (Wayland support is experimental - System Tray not working)
-Fedora with Gnome on Wayland/X11
-Fedora with Xfce on X11
-FreeBSD with Gnome on X11
-FreeBSD with KDE Plasma on Wayland
-OpenBSD with Gnome on X11

**Attached you can find the instructions (screenshots), how it would look like once added:**

Current README.LINUX.md:

<img width="1153" height="360" alt="Image" src="https://github.com/user-attachments/assets/2d29e4c9-f506-4836-85ae-287ae5a4d5b1" />

New README.LINUX.md:

<img width="1158" height="1149" alt="Image" src="https://github.com/user-attachments/assets/5a9fb405-bb08-4583-8922-71948204a2d0" />
<img width="1165" height="1089" alt="Image" src="https://github.com/user-attachments/assets/6a18fa07-8345-4805-aca6-fa2aa4989368" />

froggie2523:
File updated via PR #1614, ok to close.

PR #1614, opened 2025-09-16
[doc] Update README.LINUX.md with instructions to enable System Tray

froggie2523:
In many distros, Password Safe does not appear in the System Tray. Password Safe relies on wxWidgets to detect whether System Tray support is available in the OS. Once the feature is enabled at the system level, the "Put icon in System Tray" checkbox becomes available under Options > System.
I researched System Tray support in Linux, and it can be enabled in many distros, even when running on Wayland or using Flatpak.
This PR adds details on how to make System Tray work.

Tested in distros for which a Password Safe package is available:
-Debian with Gnome on Wayland/X11
-Ubuntu with Gnome on Wayland/X11
-Ubuntu with Cinnamon on X11 (Wayland support is experimental - System Tray not working)
-Kubuntu with KDE Plasma on Wayland/X11
-Mint with Cinnamon on X11 (Wayland support is experimental - System Tray not working)
-Fedora with Gnome on Wayland/X11
-Fedora with Xfce on X11

### DIFF
--- a/README.LINUX.md
+++ b/README.LINUX.md
@@ -80,6 +80,70 @@ $ flatpak run org.pwsafe.pwsafe
 ```
 See https://flathub.org/setup to get started using flatpak.
 
+## Enabling System Tray support in Password Safe
+Issue: Password Safe is running but not appearing in the System Tray.
+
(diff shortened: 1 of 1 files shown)


rafaelx:
Closes #1612

Issue #1339, opened 2024-07-14
[Wayland] Lock button and Alt+Tab and Unlock dialog does not appear automatically


igor2x:
Password Safe v1.19.1 flatpak on Ubuntu 24.04.

SCENARIO 1 - Wayland (the problem)
1. Make sure you have not changed any flatpak settings (like changing some _Flatseal_ setting or changing .desktop icon or starting flatpak from terminal with some parameters).
2. Start Password Safe and type in master password.
3. Click _Lock_ icon or from menu _File | Lock_.
4. Alt+Tab to some other program and Alt+Tab back to Password Safe. Problem appears. I expect _Unlock Password Database_ dialog automatically appears (like it does on Windows), but what appears is _Lock_ button is greyed out and main program is displayed without any group or entry (it looks similar to password database opened). In order to unlock the database you have to use _File | Unlock_.

SCENARIO 2 - X-window (works fine)
1. Disable "Wayland".
First close down _Password Safe_.
 a) Or start _Password Safe_ from terminal: `flatpak run --nosocket=wayland org.pwsafe.pwsafe`
 b) Or open /var/lib/flatpak/exports/share/applications/org.pwsafe.pwsafe.desktop file and after `/usr/bin/flatpak run ` add `--nosocket=wayland `
 c) Or use _Flatseal_ GUI flatpak and disable "Wayland" and start _Password Safe_.
![image](https://github.com/user-attachments/assets/f9ef07c1-6c37-4b9a-b74f-d4092bb76515)
3. Click _Lock_ icon or _File | Lock_.
4. Alt+Tab to some other program and then Alt+Tab back to _Password Safe_ and _Unlock Password Database_ automatically appears (that is how Windows version is working) and it makes sense to me.

--- 

It looks like another _Password Safe_ issue with Wayland.

Note: If you login into Linux with X-Window session, you probably get to the "SCENARIO 2" and never experience "SCENARIO 1" issue.

Issue #1322, opened 2024-06-29
[Wayland] Unlock Password dialog does not appear


nobugshere:
Using local builds from the current master: 
1. Open a file, then lock it.  The window minimizes.
2. Click the icon and an empty main window appears but the Unlock Password Database dialog does not.
3. Using File|Unlock or Ctrl-I brings-up the dialog and you can proceed as normal.

I've only seen this happen on Wayland based desktops:
Fedora 40 ARM64 VMware, KDE/Wayland, gcc 14.1.1, wx 3.2.4, GTK 3.24.42
Ubuntu 24.04 ARM64 VMware, Gnome/Wayland, gcc 13.2.0, wx 3.2.4, GTK 3.24.41

Specifically, it works as expected on these platforms:
macOS M1Max Sonoma 14.5, Xcode 15.4, wx 3.2.4 and 3.2.5 (local debug build)
Fedora 40 ARM64, Xfce/X11, gcc 14.1.1, wx 3.2.4, GTK 3.24.42

igor2x:
This is an old issue I have never reported, but I know it has been present for a long time. For example I can reproduce a problem in Password Safe v1.18.2 using flatpak. But I am quite sure this problem existed even before. I used to think this is just one of the feature not yet implemented on Linux, because there are many thinks not working (auto-type, virtual keyboard... but there are all Wayland specific).

---

Additionally it looks like password database was closed down or that somehow program crashed, because there is this automatic minimize on Lock.

![image](https://github.com/pwsafe/pwsafe/assets/10895030/baa2c49a-2539-4fee-bc3f-435688b9e2f9)

I used to think this was some issue that appears time to me (but it was auto-lock after time expiration) and I used to think the program has somehow closed down the database (because no groups/entries are visible) and when I clicked on Open and select a database it returned error the database is already opened. This all looks like weird behaviour and strange UX, but then somehow I have found out the solution was File | Unlock.

I have also lost some passwords, because I have unchecked auto-password saving and when this issue appeared I closed down the program and there used to be strange Yes/No questions and easily miss correct option.

---

I think fixing problems like this would massively improve UX. On Windows when user clicks on Lock button program minimizes automatically, but when clicked on program again to restore from minimize Unlock dialog is automatically displayed.

nobugshere:
Hi @igor2x,
[...]

Issue #1093, opened 2024-02-05
Password dialog for unlocking doesn't always show (on Linux)


atleta:
I have auto lock enabled and sometimes when I maximize the locked app the unlock dialog just won't show so I'm stuck with an unusable instance that won't react to the mouse or keyboard and has to be killed explicitly. (It seems it doesn't receive the keyboard focus, so even after trying to focus it, e.g. clicking with the mouse, the keyboard events go to other windows.)

![image](https://github.com/pwsafe/pwsafe/assets/6267837/fdf039cd-b7f0-4de5-ac5c-a1af0670ea61)

igor2x:
Can you select from menu File | Unlock Safe.
If yes, then this issue is similar (but not the same as) bug https://github.com/pwsafe/pwsafe/issues/1088

atleta:
Unfortunately not. When this happens, the app doesn't react to any input. (Including clicking the menubar.)

igor2x:
Sorry, I have misread your first post. Mouse and keyboard does not respond on click. It looks like password dialog was hidden somewhere behind some other window or similar.

1. What is your Linux distribution name and version?
2. What is your Password Safe version? From Help | About
3. How did you install the program?
3.1 Using version from distribution repository.
3.2. Downloading deb or rpm package from Github or Sourceforge.
3.3. Downloaded source code and compiled it yourself?
3.4 Downloaded manifest file and created and installed flatpak version.
3.5. Running Password Safe inside Wine, Bottles or similar. Which one.
3.6. Something else. What?
4. What is the setting that triggers auto-lock in your case? I have tested some of the options on Password Safe v1.18.2 on Ubuntu 23.10 and I see from menu Manage | Options | Security icon on left site
a) Lock password database after "1" minute. WORKS FINE: It locks database after a minute.
b) Lock password database on minimize. DOESN'T WORK: After minimize and click to restore from minimize and database is still unlocked.
c) Lock password database on workstation lock. DOESN'T WORK: Pressing Windows+L to lock screen. Login into Ubuntu again and password is still unlocked.
d) Do you use some other option. If yes, which one?
5. Does your problem of unable to unlock the database appears randomly or does it happen something that you suspect it may be the cause of the problem?
6. What is the reason why you maximize locked window and not unlock it before maximizing?

atleta:
> Sorry, I have misread your first post. Mouse and keyboard does not respond on click. It looks like password dialog was hidden somewhere behind some other window or similar.

[...]

PR #1616, opened 2025-09-21
[wx] Fix minimize to System Tray on Linux

froggie2523:
This PR provides a fix to make the program's window minimize to Task Bar (so it’s Alt+Tab-able) in case System Tray becomes not available.

Issue: In many Linux distros, when the System Tray support becomes unavailable (e.g. due to migration to Wayland), Password Safe's window on locking the database behaves like it disappears:
In this case, there’s no Tray or Task Bar icon for user to click to restore the window.
Since the minimized window is unmapped/hidden, it also doesn’t show up in Alt+Tab or when using the compositor's task switcher.
Then, user can only kill existing process of Password Safe and start the program again.
A confused user would most probably launch Password Safe again trying to open the same database, only to be prompted with an error message: "Lock is done by yourself".

Some scenarios when this can happen:
1. Some distros provide a Wayland or Xorg session to choose from on the login screen. A user using pwsafe with System Tray logs out of Xorg session and logs in to Wayland session.
2. The OS gets upgraded with Wayland as the default display server, and an Xorg session may not be available anymore as it's deprecated.
3. The desktop environment extension providing support for System Tray is disabled or uninstalled.
4. User migrates from OS with System Tray support to OS without it, including the pwsafe config file.
5. New version of Password Safe gets installed, overwriting the application's .desktop file where a workaround for System Tray was configured (as described in README.LINUX.md).
  In all cases, the UseSystemTray=1 settings is saved in pwsafe.cfg.
  On startup, Password Safe assumes System Tray is still available, based on settings in pwsafe.cfg.
  Database is locked manually, or after the timeout specified in Options > Security.

Somewhat related issues on Github:
https://github.com/pwsafe/pwsafe/issues/1322
https://github.com/pwsafe/pwsafe/issues/1339
https://github.com/pwsafe/pwsafe/issues/1093
https://github.com/pwsafe/pwsafe/issues/1223

Attached you can find some screenshots (Debian on X11 -> Wayland).

pwsafe_1 22-wxWidgets-bug-SysTrayLock-01
pwsafe_1 22-wxWidgets-bug-SysTrayLock-02
pwsafe_1 22-wxWidgets-bug-SysTrayLock-03
pwsafe_1 22-wxWidgets-bug-SysTrayLock-04

After the fix:

pwsafe_1 22-wxWidgets-bug-SysTrayLock-05

### DIFF
--- a/src/ui/wxWidgets/PWSafeApp.cpp
+++ b/src/ui/wxWidgets/PWSafeApp.cpp
@@ -585,8 +585,14 @@ void PWSafeApp::FinishInit() {
   } else {
     SetTopWindow(m_frame);
   }
-  if (PWSprefs::GetInstance()->GetPref(PWSprefs::UseSystemTray))
-    m_frame->ShowTrayIcon();
+  if (PWSprefs::GetInstance()->GetPref(PWSprefs::UseSystemTray)) {
+    if ( IsTaskBarIconAvailable() ) {
+      m_frame->ShowTrayIcon();
+    }
+    else {
+      PWSprefs::GetInstance()->SetPref(PWSprefs::UseSystemTray, false, false);
+    }
+  }
 
   m_frame->Register(); // register modal dialog hook
   m_initComplete = true;


src/ui/wxWidgets/PWSafeApp.cpp
```diff
@@ -585,8 +585,14 @@ void PWSafeApp::FinishInit() {
...
     SetTopWindow(m_frame);
   }
-  if (PWSprefs::GetInstance()->GetPref(PWSprefs::UseSystemTray))
-    m_frame->ShowTrayIcon();
+  if (PWSprefs::GetInstance()->GetPref(PWSprefs::UseSystemTray)) {
+    if ( IsTaskBarIconAvailable() ) {
+      m_frame->ShowTrayIcon();
+    }
+    else {
+      PWSprefs::GetInstance()->SetPref(PWSprefs::UseSystemTray, false, false);
```
dskrvk, line 593: So we've updated the pref, but where is it read again, causing the desired change in behavior?

rafaelx: @dskrvk The desired behavior results from changes made to the PWSafeApp::FinishInit function. If the environment does not support the system tray, the setting is simply reset, and the system tray icon will not be shown. On subsequent launches of the application under this environment, the system tray icon will also not be shown. It will only be shown during application startup if both the environment supports this and the user setting allow it.

ronys:
Nice - thanks!

PR #1627, opened 2025-10-15
V3 att windows support

ronys:
Support attachments in V3 databases under Windows using @huven 's V3 format addition (Att* field types).

The approach I've taken is to keep the code as agnostic as possible to the attachment's source, V3 or V4.

### DIFF
--- a/docs/ReleaseNotes.md
+++ b/docs/ReleaseNotes.md
@@ -6,6 +6,12 @@ In the following, SFxxxx refers to Bug Reports, Feature Requests and Service Req
 
 PasswordSafe 3.70.0pre Release ???
 ==========================================
+New Features in 3.70.0pre
+-------------------------
+* Password Safe now supports adding a single attachment to an entry. Previously this was only available in the experimental V4 database format. With this release, the ability to add, update, remove and download a single attachment to an entry is now available via a new "attachment" tab in the Add/Edit Entry window.
+
+Bugs Fixed in 3.70.0pre
+-----------------------
 * "Always on top" now persists correctly across restarts when "Start Password Safe at Login" is selected.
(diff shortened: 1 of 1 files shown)


src/core/ItemData.h
```diff
@@ -373,7 +373,7 @@ class CItemData : public CItem
   void SetAttModificationTime(time_t t)          { CItem::SetTime(DATA_ATT_MTIME, t); }
   void SetAttContent(const unsigned char *content, size_t clen) { CItem::SetField(DATA_ATT_CONTENT, content, clen); }
 
-  bool HasAttachment() const            { return IsAttMediaTypeSet();            }
+  bool HasAttachment() const            { return IsAttMediaTypeSet() || HasAttRef(); }
```
huven, line 377: When adding V4 here, does it make sense (for semantic consistency) to clear AttRef in ClearAttachment() too? That would allow code like

```
if (HasAttachment())
  ClearAttachment();
```

For v4, the attachment record still needs to be dropped separately of course.

ronys: Interesting suggestion.
I don't think symmetry/consistency is warranted here, since HasAttachment() affects display behavior regardless of V3/V4 (e.g., popup menu, clip icon). ClearAttachment() for V4, however, requires the core to drop the attachment reference, as you've noted.

So either we complicate ClearAttachment by passing a reference to the core, or we rename it to ClearV3Attachment.

I'm leaning towards the latter, at least for now. What say?

huven: ClearV3Attachment sounds good to me 👍, I think it's better not to introduce a dependency on core in the item code.

ronys: Thanks! Done in commit 32ebed8

PR #1629, opened 2025-10-19
Fix report names

pm-kan:
Previously there was confusion between localized and fixes report names: 
- `CReport::ReportNames` contains non-localizable items with spaces and these names were used as file names;
- on other hand, localizable items in `core_st` contains underscores instead of spaces (these items were used as first placeholder for report descriptions `"Report of %ls function started at %ls"`);
- also MFC version incorrectly used translatable names when checking for presence of report file.

In this PR files names were changed to use underscores and spaces were added to strings used in report's content.

Also new files were added to Code::Blocks project.

### DIFF
--- a/CodeBlocks/core/core.cbp
+++ b/CodeBlocks/core/core.cbp
@@ -387,6 +387,8 @@
 		<Unit filename="../../src/core/UnknownField.h" />
 		<Unit filename="../../src/core/Util.cpp" />
 		<Unit filename="../../src/core/Util.h" />
+		<Unit filename="../../src/core/Validate.cpp" />
+		<Unit filename="../../src/core/Validate.h" />
 		<Unit filename="../../src/core/VerifyFormat.cpp" />
 		<Unit filename="../../src/core/VerifyFormat.h" />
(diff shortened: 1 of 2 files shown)


src/ui/wxWidgets/MenuViewHandlers.cpp
```diff
@@ -225,8 +225,13 @@ void PasswordSafeFrame::RunShowReport(int iAction)
...
   }
   else {
-    wxString tcAction = CReport::ReportNames.find(iAction)->second;
-    wxMessageBox(_(tcAction) + _(L" file \'") + rpt.GetFileName() + _(L"' not readable"), _("View Report"), wxOK|wxICON_ERROR);
+    stringT title;
+    stringT reportName;
+    LoadAString(reportName, iAction);
+    Format(title, _("View Report: %ls"), reportName.c_str());
+    stringT message;
+    Format(message, _(L"File \'%ls\' not readable"), rpt.GetFileName().c_str());
```
dskrvk, line 233: Thank you! These format strings will be much easier to translate.

Issue #1626, opened 2025-10-14
After search (Strg+F) found entry is covered by horizontal scroll bar


therob84:
**System**: Linux Mint 21.3
**PW**: passwordsafe-ubuntu22-1.22-amd64.deb

**Issue**:
If I search for an entry (mostly Strg+F) and an entry is found at a position not already on screen the marked entry is shown as the last entry on the screen. When I try to click it, instantly the horizontal scroll bar appears (it was hide out before), and I can't see or click the entry anymore.
In fullscreen/maximized window it does not happen, as the window is large enough for no need of the horizontal scroll bar at all.

**Workaround**: 
I need to manualle scroll up down to click the entry.

**Solution**:
show found entries in the middle of the window after hitting search/enter, not at the bottom.

rafaelx:
I tried to reproduce this on Linux Mint and Ubuntu, but couldn't observe the described effect.

I assume it occurs in list view, since a horizontal scrollbar only appears there. Here are some screenshots of Linux Mint with Cinnamon.

The test db contains the following list of entries:
<img width="641" height="295" alt="Image" src="https://github.com/user-attachments/assets/3f3214e4-4ece-4118-b6ec-0166d8de948d" />

Minimized window to hide some of the last entries:
<img width="641" height="174" alt="Image" src="https://github.com/user-attachments/assets/9b8ccfac-f771-492d-8bb1-11a087759149" />

Result when searching for entry 'h':
<img width="641" height="174" alt="Image" src="https://github.com/user-attachments/assets/9ac33e44-cac4-4474-b639-41f5547e4a86" />

The scroll bar only appears when the mouse pointer is moved there:
<img width="641" height="174" alt="Image" src="https://github.com/user-attachments/assets/b9a20270-68eb-46bd-ad04-2c84cb310c50" />

Search result for last entry 'i':
<img width="641" height="174" alt="Image" src="https://github.com/user-attachments/assets/f03f6ff8-6b92-4eaa-a472-6ba67affea16" />

<img width="641" height="174" alt="Image" src="https://github.com/user-attachments/assets/e36170fe-98d1-40e5-bfee-47319939ad0b" />


There always seems to be enough space between the found and highlighted search entry and the edge of the list view, so that the scrollbar never obscures an element. - Any suggestions on how to reproduce this?

rafaelx:
Thanks for the additional information.

[...]

PR #1632, opened 2025-11-02
[wx] Allow higher search result positioning via environment variable

rafaelx:
Related to #1626

### DIFF
--- a/src/ui/wxWidgets/TreeCtrl.cpp
+++ b/src/ui/wxWidgets/TreeCtrl.cpp
@@ -948,11 +948,30 @@ void TreeCtrlBase::SelectItem(const CUUID & uuid)
       Expand(parent);
     }
     ::wxSafeYield();
-    if (wxUtilities::IsAdvancedScrollEnabled()) {
-      auto sibling = searchSibling(id);
-      EnsureVisible(sibling.IsOk() ? sibling : id);
+    // Determine an item's height based on its surrounding rectangle.
+    // To obtain a surrounding rectangle, the element must be visible.
+    auto firstVisibleItem = GetFirstVisibleItem();
+    if (firstVisibleItem.IsOk()) {
+      wxRect boundingRect;
+      if (GetBoundingRect(firstVisibleItem, boundingRect)) {
+        auto itemHeight = boundingRect.GetHeight();
+        auto viewHeight = GetClientSize().GetHeight();
+
+        // Scroll the tree view up one more position (to the next sibling element)
+        // if there is sufficient space (at least two items) in the tree view.
+        if (2 /*visible items*/ * itemHeight < viewHeight) {
+          auto sibling = searchSibling(id);
+          EnsureVisible(sibling.IsOk() ? sibling : id);
+        }
+        else {
+          // If there is not enough space in the tree view, then scroll only
+          // to the search result instead of to the next sibling element.
+          EnsureVisible(id);
+        }
+      }
     }
     else {
+      // By default, scroll directly to the search results.
       EnsureVisible(id);
     }
     ::wxSafeYield();
--- a/src/ui/wxWidgets/wxUtilities.cpp
+++ b/src/ui/wxWidgets/wxUtilities.cpp
@@ -364,41 +364,6 @@ void wxUtilities::NotifyIfUnsupported(enum Feature feature, wxWindow* window)
 #endif
 }
 
-/**
- * Checks if the environment variable PWS_ADVANCED_SCROLL
- * is set to a value greater than zero.
- * 
- * Using this variable, the user can activate the option to
- * position the search result higher up in the tree view and,
- * in the case of an individual configuration of the scrollbar
- * size, to prevent it from being obscured by the scrollbars.
- */
-bool wxUtilities::IsAdvancedScrollEnabled()
-{
-  static long advancedScrollEnabled = -1L;
-
-  // Get the env. variable only once
-  if (advancedScrollEnabled == -1L) {
-    wxString PWS_ADVANCED_SCROLL = wxEmptyString;
-    if (wxGetEnv(wxT("PWS_ADVANCED_SCROLL"), &PWS_ADVANCED_SCROLL)) {
-
-      if (!PWS_ADVANCED_SCROLL.IsEmpty() && PWS_ADVANCED_SCROLL.ToLong(&advancedScrollEnabled)) {
-        if (advancedScrollEnabled < 0) { // Don't call wxGetEnv() more than once per process if value negative
-          advancedScrollEnabled = 0L;
-        }
-        // The user has specified a value of zero or greater,
-        // which is taken into account when returning the result.
-      }
-      else { // Don't call wxGetEnv() more than once per process if value bad
-        advancedScrollEnabled = 0L;
-      }
-    } else { // Don't call wxGetEnv() more than once per process if not set
-      advancedScrollEnabled = 0L;
-    }
-  }
-  return advancedScrollEnabled > 0;
-}
-
 // Wrapper for wxTaskBarIcon::IsAvailable() that doesn't crash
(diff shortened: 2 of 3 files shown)


src/ui/wxWidgets/TreeCtrl.cpp
```diff
@@ -947,9 +947,43 @@ void TreeCtrlBase::SelectItem(const CUUID & uuid)
...
+      Searches for a subsequent or sibling element (password element / group element)
+      in the tree view at the same level. If no element is found, the search continues
+      one level higher until the root element is reached.
+      The goal is to find the next visual element below the target element so that it
+      appears visually positioned slightly higher. See also the documentation for
+      wxUtilities::IsAdvancedScrollEnabled().
+      itemId refers either to the found sibling element or to an intermediate result,
+      therefore the original element from which the search begins should be stored.
+    */
+    std::function<bool(wxTreeItemId&)> searchSibling = [&](wxTreeItemId& itemId) -> bool {
```
dskrvk: Why not make this a private method of the class? I think anonymous functions are meant to be compact, otherwise they don't really help break down the code into easy-to-read chunks (since they're still part of the parent function's body).

dskrvk: Also, not sure it's a good practice to use the same parameter for both input and output. Why not return the value (or some sentinel if not found) and keep the input unmodified? The function name seems to indicate it does only reads.

src/ui/wxWidgets/wxUtilities.cpp
```diff
@@ -364,6 +364,41 @@ void wxUtilities::NotifyIfUnsupported(enum Feature feature, wxWindow* window)
...
+/**
+ * Checks if the environment variable PWS_ADVANCED_SCROLL
+ * is set to a value greater than zero.
+ * 
+ * Using this variable, the user can activate the option to
+ * position the search result higher up in the tree view and,
+ * in the case of an individual configuration of the scrollbar
+ * size, to prevent it from being obscured by the scrollbars.
+ */
+bool wxUtilities::IsAdvancedScrollEnabled()
```
dskrvk: Why use an env variable rather than a UI config option or even make this behavior apply always?

src/ui/wxWidgets/wxUtilities.cpp
```diff
@@ -364,6 +364,41 @@ void wxUtilities::NotifyIfUnsupported(enum Feature feature, wxWindow* window)
...
+{
+  static long advancedScrollEnabled = -1L;
+
+  // Get the env. variable only once
+  if (advancedScrollEnabled == -1L) {
+    wxString PWS_ADVANCED_SCROLL = wxEmptyString;
+    if (wxGetEnv(wxT("PWS_ADVANCED_SCROLL"), &PWS_ADVANCED_SCROLL)) {
+
+      if (!PWS_ADVANCED_SCROLL.IsEmpty() && PWS_ADVANCED_SCROLL.ToLong(&advancedScrollEnabled)) {
```
dskrvk: For a boolean-like variable we shouldn't care what the value is as long as it's defined/non-empty. Easier for users that way.

rafaelx:
wxWidgets sometimes seems to take enlarged scrollbars into account when the application's main window has been resized or resized to a specific size, potentially resulting in elements being partially or completely obscured, or when the scrollbars were displayed before the search was performed. I can't reliably reproduce this behavior; I only noticed it by chance while preparing the pull request.

Normally, a search result that isn't visible is placed at the bottom of the view or the last row of the tree view within the visible area. This pull request moves the search result up one position, if possible. However, if wxWidgets also correctly takes the scrollbar size into account, the search result is moved up another position, for a total of two positions. If a user has set the height of the main window to a particularly small size, a search result could be pushed out of the visible area. I suspect this is a rare occurrence, but it's possible.

I currently have concerns about this pull request if it might lead to inconsistent or undesirable user interface behavior. Therefore, I think this functionality shouldn't be enabled by default and should be explicitly activated by the user. Whether this is done via an environment variable or user settings is still open for discussion. The original problem of #1626 also lies somewhat in the characteristic of dynamically displayed scrollbars, their size and a particularly small height of the application window, or perhaps simply in wxWidgets or the underlying Gtk.

@dskrvk I'll take a look at the other comments as well. Thanks for your feedback.

src/ui/wxWidgets/TreeCtrl.cpp
rafaelx: I liked the lambda, which initially served as a playground for experimentation, but the arguments were too convincing. :)

src/ui/wxWidgets/wxUtilities.cpp
rafaelx: If the variable is documented, wouldn't it be more logical to describe activating the functionality with "PWS_ADVANCED_SCROLL=1" instead of "PWS_ADVANCED_SCROLL must be set to any character"?

If the documentation refers to the value 1 or any value greater than 0 to activate the functionality, doesn't that automatically create the expectation that deactivation with a specific value is also possible?

If someone uses something similar to the following...
```
PWS_ADVANCED_SCROLL=1 pwsafe&
```
...then a quick or temporary deactivation can be achieved as follows...
```
PWS_ADVANCED_SCROLL=0 pwsafe&
```
...without having to remove the variable from the script and then add it back in.

rafaelx:
I doubt that a feature "hidden" behind an environment variable will be used by many people and will generate significant feedback. Since the height of a tree element and the visible height of the tree view can be determined, the additional scrolling of the tree view is no longer statically enabled, but dynamically at runtime. This approach thus also corresponds to @dskrvk 's suggestion and takes into account when additional scrolling is allowed and when it is not.

I find the result acceptable, so the draft status has been removed.

Linux Mint with the Cinnamon desktop was used as the test environment with modified scrollbar size.

#### Scrollbar Configuration
Settings > Themes > Advanced Settings > Settings

- Use overlay scrollbars
- Scrollbar size 19px or any other setting

#### Test Database
- [pwsafe11.psafe3](https://github.com/user-attachments/files/23422928/pwsafe11.psafe3.zip)
- Password: 11

#### Scenarios that were considered when visualizing the search results

- With or without a horizontal scroll bar due to very long element names and small frame width
- Different frame sizes, especially if the frame height is very low
- Different font sizes for tree view, which influences the item's height

#### A few special cases

**Unusual behavior occurs when the last tree element is a group to which the tree is supposed to scroll**

Steps

1. Open the test database
2. Open search panel
3. Type x and click on 'Find Next' button or hit return on the keyboard

Result
The group p0 is correctly found as the next sibling element of x, but the tree is not scrolled to p0.
The tree's IsVisible method returns "true" for p0, meaning that p0 is visible.
Once the tree is manually scrolled to p0, p0 becomes visible in subsequent searches for x.
If the tree is not manually scrolled to p0, p0 can still be made visible by triggering the search again.
This seems to occur only with the last group in the tree. The same combination somewhere in the middle of the tree does not appear to exhibit this behavior.
It is also noticeable that in both cases (whether the tree has been scrolled to p0 or not), the available client space is the same. Therefore, there is no indication that the tree was not scrolled correctly.

**Additional scroll by wxWidgets**

Steps

1. Open the test database
2. Scroll to item with long name 'hTTTT...'
3. Reduce the width of the frame so that the element is partially hidden and the scroll bar appears when the mouse pointer is moved to the scroll bar position.
4. Scroll to first item so that item e is not visible
5. Search for item e
6. Increase the width of the frame until the view of the tree is wider than its longest element.
7. Search for item e

Result
During the first search the scroll bars are not taken into account, and the tree is only moved by one position by the PR code.
In case of the second search the tree is scrolled by wxWidgets and additionally by the code of this PR

Further steps, just for info

- Reduce the height of the frame by a few pixels
- Search for item e again
- Repeat reducing the frame's height and the search for item e and check whether the search result is always visible

PR #1636, opened 2025-11-12
Wx v3 attachments

ronys:
This brings support for attachment in the V3 file format to wx, equivalent to the support added to the Windows version in 3.70.

Once this is merged, 1.23 can be released.

### DIFF
--- a/src/core/PWScore.cpp
+++ b/src/core/PWScore.cpp
@@ -3499,23 +3499,18 @@ void PWScore::RemoveAtt(const pws_os::CUUID &attuuid)
 
 AttList::size_type PWScore::GetNumAtts() const
 {
-  switch (GetReadFileVersion()) {
-    case PWSfile::V30:
-    {
-      AttList::size_type retval = static_cast<AttList::size_type>(
-        std::count_if(m_pwlist.begin(), m_pwlist.end(),
-          [](const auto &p) {
-            return p.second.HasAttachment();
-          })
+  if (GetReadFileVersion() == PWSfile::V40)
+    return m_attlist.size();
+  else {
+    AttList::size_type retval = static_cast<AttList::size_type>(
+      std::count_if(m_pwlist.begin(), m_pwlist.end(),
+        [](const auto &p) {
+           return p.second.HasAttachment();
+         } )
       );
-      return retval;
-    }
-    case PWSfile::V40:
-      return m_attlist.size();
-    default:
-      return 0;
-    }
+    return retval;
   }
+}
 
 std::set<StringX> PWScore::GetAllMediaTypes() const
 {


src/ui/wxWidgets/TreeCtrl.cpp
```diff
@@ -570,13 +570,18 @@ wxString TreeCtrlBase::ItemDisplayString(const CItemData &item) const
   }
 
   if (item.IsProtected()) { 
-    disp += wxT(" #");
-#ifdef NOTYET
-    wxUniChar padlock(0x1f512);
-    disp +=padlock;
-#endif
+    // disp += wxT(" #");
```
rafaelx: Is the comment still needed?

ronys: Keeping it for historical reference, added a bit of clarification.

ronys:
>There doesn't seem to be an indicator in the list view, does there?
>
Fixed, 
> From a user's perspective, it's somewhat strange that an entry in the "AddEdit" dialog is always recognized as changed after an undo operation, even though the user only viewed the data and didn't make any changes. However, this seems to be independent of these changes.
> 
Not fixed (for now)

> Does the V3 format not support a creation time for the attachment?
>
No. We decided to leave this out for simplecity.
 
> If the AddEdit dialog is closed in a V4 database without any changes to an entry (even without an attachment), a message will always appear indicating that there are changes (to the attachment).
Not fixed (for now)

ronys:
> After quick build and test on macOS, I have a few observations. (Sorry if some of these have already been discussed.) macOS M1Max Sequoia 15.7.2, Xcode 16.4, wx 3.2.8 (local build)
> 
> 1. The new symbols look good on the Mac.
> 2. On the Attachment tab: The Title text box is somewhat taller than it needs to be.
Changed the size to default, should fix this.
> 3. As @rafaelx pointed out, the change detection logic needs work: AddEditPropSheetDlg::NewAddEntryCommand() and AddEditPropSheetDlg::GetChanges()
Didn't fix this yet.
> 4. I noticed the Windows version does not accept plain text files as attachments, the Mac build does.
The windows version accepts all file type, you just need to specify "All Files" in the file selector.
> 5. Attachment size is rounded up to a multiple of 8 bytes and, when exported, the file is padded with zeros.  The Windows version does this as well.  The format spec specifies the size limit, but I don't see any discussion of how the original file size is ment to be handled.  Seems like this could be a problem for some file types.  Even if not, I think most users would expect an exported file to be identical to the original.
Fixed (this was the most sever bug, IMHO).
> 6. The feature is not yet documented in the Help text.
Didn't fix this yet.

ronys:
> > > 5. Attachment size is rounded up to a multiple of 8 bytes and, when exported, the file is padded with zeros.  The Windows version does this as well.  The format spec specifies the size limit, but I don't see any discussion of how the original file size is ment to be handled.  Seems like this could be a problem for some file types.  Even if not, I think most users would expect an exported file to be identical to the original.
> > >    Fixed (this was the most severe bug, IMHO).
> 
> Looks good for new attachments, but an existing attachment still has the wrong length. Maybe not a problems for the WX version itself, but the Windows 3.70 version is already released. I'm not sure what to suggest.

Existing attachments are stored incorrectly, so there's not much that can be done. That's why I pushed 3.70.1 out so soon.

PR #1638, opened 2025-11-16
[wx] Show confirmation dialog before clearing list of recently opened databases

### LINKED ISSUE #1337: [wx] Confirmation dialog is missing when clearing Clear Recently Opened List
File | Clear Recently Opened List.
- On Windows there is [confirmation dialog](https://github.com/pwsafe/pwsafe/issues/1127).
- On Linux there is no confirmation dialog. List just gets deleted.

I think behaviour on Windows and Linux should be the same unless there is some operating system specifics (that I can't see in this case).

rafaelx:
Resolves #1337

### DIFF
--- a/src/ui/wxWidgets/MenuFileHandlers.cpp
+++ b/src/ui/wxWidgets/MenuFileHandlers.cpp
@@ -278,11 +278,19 @@ void PasswordSafeFrame::OnUnlockSafe(wxCommandEvent&)
  * wxEVT_COMMAND_MENU_SELECTED event handler for ID_MENU_CLEAR_MRU
  */
 
-void PasswordSafeFrame::OnClearRecentHistory(wxCommandEvent& evt)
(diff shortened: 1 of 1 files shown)


ronys:
Nice.

@igor2x has taught me that OK/Cancel options are potentially confusing. In this case, I think it would be better to put "Clear" instead of "OK", so that the choices would be "Clear" and "Cancel".

(Like the change we've made to the modified entry warning: "One or more values have been changed. Do you want to discard the changes? Discard/Cancel")

rafaelx:
This is what the notification dialog looks like under Linux Mint, following the latest changes to various dialogs initiated by @igor2x.

grafik


The label on the "OK" button should be replaced with "Clear" by the following line of code.

```
dialog.SetOKLabel(_("Clear"));
```

Do you still see the "OK" label on your system? If so, what environment are you using?

ronys:
My bad - missed the call to SetOKLabel().

Apologies.

PR #1653, opened 2025-11-30
[wx] Improve implementation and design of database properties dialog

### LINKED ISSUE #1229: Simplify File | Properties dialog
Password Safe v1.19-pre on master branch on Ubuntu 23.10.

Getting into [issue 1228](https://github.com/pwsafe/pwsafe/issues/1228) it looks like "Close" button is not innocent, it actually saves the data to database. Close button should not be used to save data of any kind. I think "Close" button should be renamed.

Or... probably better, simplified solution is to just make Properties dialog more useful with two fields.

image

rafaelx:
Resolves #1229

### DIFF
--- a/src/ui/wxWidgets/PropertiesDlg.cpp
+++ b/src/ui/wxWidgets/PropertiesDlg.cpp
@@ -64,16 +64,31 @@ void PropertiesModel::Init()
   std::vector<stringT> allGroups;
   m_Core.GetAllGroups(allGroups);
 
+  auto allgroups = allGroups.size();
   auto numgroups = wxString::Format(_("%d"),
-    static_cast<int>(allGroups.size()));
+    static_cast<int>(allgroups));
   m_PropertiesDb.numgroups = tostringx(numgroups);
 
+  auto emptygroups = m_Core.GetEmptyGroups().size();
   auto numemptygroups = wxString::Format(_("%d"),
-    static_cast<int>(m_Core.GetEmptyGroups().size()));
+    static_cast<int>(emptygroups));
   m_PropertiesDb.numemptygroups = tostringx(numemptygroups);
 
-  m_View.SetNumOfGroups(wxString::Format(_("%ls (%ls empty)"),
-    numgroups, numemptygroups));
+  wxString groupsinfo;
+  if (allgroups == 0) {
+    groupsinfo = wxT("0");
+  }
+  else if (emptygroups <= 1) {
+    groupsinfo = wxString::Format(
+      _("%ls (%ls of which is empty)"), numgroups, numemptygroups
+    );
+  }
+  else {
+    groupsinfo = wxString::Format(
+      _("%ls (%ls of which are empty)"), numgroups, numemptygroups
+    );
+  }
+  m_View.SetNumOfGroups(groupsinfo);
 
   /////////////////////////////////////////////////////////////////////////////
   // Property: Number of entries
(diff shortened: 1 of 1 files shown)


rafaelx:
This PR picks up the suggestions from #1229. To simplify the dialog, the two buttons that open a new dialog for editing were removed. To edit the database label/name and the database description, two text input fields were added to the dialog. Both fields are initially in read-only mode, so that the dialog is purely informative. To edit the name or description, the fields must be switched to edit mode by double-clicking.

#### Dialog Design
PropertiesDlg

- The close button checks for unsaved changes so that the user is informed accordingly and can decide how to proceed.
- The save button is only enabled when there are changes and will accordingly save them if any changes are present.
- The tooltips of the text input fields provide the user with corresponding guidance on how to use them.

#### Implementation Design
Since the functionality of this dialog is quite simple, it was used experimentally to separate the user interface and application logic, without strictly applying the typical design patterns for this purpose. It would be interesting to expand the experiment to develop a shared codebase for the different platforms.

The original idea was to have only C++ code based on the Standard Library in the introduced PropertiesModel. For a comfortable creation of formatted strings, switching to C++20 would have been very helpful. However, the question arises here of how the strings to be translated should be handled. At the moment, PropertiesModel is wx and thus Linux/Mac specific due to the use of wx functions, the creation of wxStrings, and macros for the strings to be translated. But perhaps that is also the insight from this experiment: that a platform-specific PropertiesModel would be needed for each platform.

The introduction of PropertiesModel and comparison and assignment operators has made handling the data significantly easier. Checking for changes was very easy to implement, and thus enabling the Save button. An Apply button, based on this concept, might also be easier to implement in the AddEdit dialog. In the Properties dialog, the two text input fields are monitored for changes to detect user modifications. In this dialog, there were only two controls, whereas in the AddEdit dialog, there are significantly more controls through which changes can occur. The AddEdit dialog would therefore be a much greater challenge.

rafaelx:
@nobugshere My understanding of this dialog is that it is primarily intended to provide some information about the database that has just been opened. Adding a name/label and a description is an optional feature. Therefore, the text input fields are not editable by default. The main intention of the Add/Edit dialog is different. This dialog is aimed at creating an item or editing an existing one, so all controls are in "edit mode". However, I am also open to displaying the two text input fields as editable by default, without requiring an explicit switch to edit mode. Name and description could be edited immediately, and if there are differences to the information stored in the database, the save button would be activated immediately, signaling to the user right away that something has been modified.
The disadvantage of the current solution is that a double mouse click is required, meaning that pure operation via keyboard, accelerator keys and Tab key does not work to edit the text input fields.

nobugshere:
Looks good.  There seems to be some cosmetic Mac/wxWidgets issues that don't happen on Linux; I can look into them later.  Functionally, it's good.

Mac only issues:
- The font is not italic
- Bright border around the multiline text control in dark mode.
- A little quirky when I add text to the description and try to switch to the title. (But, not the other way around.)

Screenshot 2025-12-03 at 10 33 56 AM

src/ui/wxWidgets/PropertiesDlg.cpp
```diff
@@ -402,6 +402,10 @@ void PropertiesDlg::CreateControls()
   flexGridSizer->Add(itemStaticText17       , 0, wxALIGN_RIGHT|wxALL        , 5);
   flexGridSizer->Add(m_dbDescriptionTextCtrl, 1, wxALIGN_LEFT|wxALL|wxEXPAND, 5);
 
+  auto *itemStaticTextHint = new wxStaticText(this, wxID_STATIC, _("Double click the text entry fields to edit."), wxDefaultPosition, wxDefaultSize, 0);
+  itemStaticTextHint->SetFont((itemStaticTextHint->GetFont()).Italic());
```
nobugshere: Duplicating this "SetFont" line makes it work every time on macOS.

rafaelx:
PropDlgSimplified

I think the simplified version of the dialog feels somehow right now. Simple, clear, with an easily accessible editing option and no need for additional explanations.

ronys:
Odd. The attached database has one group with one entry. However, the Properties dialog insists that it's an empty group:
image
image


Perhaps it has to do with how I created this: I created the group via right-click Add Group, and then I edited the "when" entry, setting its group in that Add/Edit dialog.

[pr1653.zip](https://github.com/user-attachments/files/24060619/pr1653.zip) - zipped because GH doesn't support uploading .psafe3 files. Password Sample123

ronys:
Update: The Windows version reports the correct numbers (1 group / 0 empty), but release 1.23 also reports 1 / 1, So I'm merging this PR, and opening a new ticket on the issue.

PR #1654, opened 2025-12-02
Allow usage of custom logger

huven:
This PR enables third-party code using the core library to register a custom (subclassed) PWSLog instance, allowing capturing all logging in one location.

### DIFF
--- a/src/core/PWSLog.h
+++ b/src/core/PWSLog.h
@@ -12,6 +12,7 @@
 #include "../os/typedefs.h"
 #include "../os/logit.h"
 
+#include <cassert>
 #include <deque>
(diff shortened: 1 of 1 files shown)


src/core/PWSLog.h
```diff
@@ -12,6 +12,7 @@
 #include "../os/typedefs.h"
 #include "../os/logit.h"
 
+#include <assert.h>
```
ronys: I think this should be 
`<cassert>` instead of `<assert.h>`
for consistency

huven: That's better indeed, changed now.

PR #1655, opened 2025-12-02
[wx] Fix sorting of columns by timestamp

### LINKED ISSUE #1646: Sorting by date/time sorts alphabetically
Hello, thanks for your great work!

I noticed that in the "Flattened List" view in the Linux version (v1.23), sorting by any of the date/time columns doesn't actually sort by date but alphabetically (by the formatted timestamp), which starts with the weekday and day of month in my case.
I.e. all Fridays come first, then Mondays etc., so "Tue 17 Oct 2023" comes before "Tue 18 Sep 2018", which in turn comes before "Wed 12 Sep 2018".

The expected behavior would be sorting by the actual date value, i.e. "Wed 12 Sep 2018" before "Tue 18 Sep 2018" before "Tue 17 Oct 2023".

rafaelx:
Resolves #1646

### DIFF
--- a/src/ui/wxWidgets/GridCtrl.cpp
+++ b/src/ui/wxWidgets/GridCtrl.cpp
@@ -18,6 +18,7 @@
 #endif
 
 ////@begin includes
+#include <wx/regex.h>
 ////@end includes
 
 #include "GridCtrl.h"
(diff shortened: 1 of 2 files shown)


src/ui/wxWidgets/GridCtrl.cpp
```diff
@@ -670,6 +708,118 @@ void GridCtrl::RearrangeItems(ItemsCollection& collection, int column)
...
+    }
+    /*
+      The case if a date-time combination is expected.
+    */
+    else if (!dateOnly && dt.ParseDateTime(datetime, &end)) {
+      collection.insert(std::pair<wxDateTime, const CItemData*>(dt, GetItem(row)));
+    }
+    /*
+      The case if a date without a time is expected.
+      wxDateTime::ParseDate interprets 01/02/25 as mm/dd/yy, but it's actually dd/mm/yy,
```
dskrvk, line 740: Isn't the format locale-dependent?

src/ui/wxWidgets/GridCtrl.cpp
```diff
@@ -670,6 +708,118 @@ void GridCtrl::RearrangeItems(ItemsCollection& collection, int column)
...
+              collection.insert(std::pair<wxDateTime, const CItemData*>(wxDateTime::Today(), GetItem(row)));
+              continue;
+          }
+        }
+        else {
+          pws_os::Trace(L"Failed to parse item's date (day, month or year) string: %ls ; using: %ls", datetime.wc_str(), wxDateTime::Today().FormatISOCombined().wc_str());
+          collection.insert(std::pair<wxDateTime, const CItemData*>(wxDateTime::Today(), GetItem(row)));
+          continue;
+        }
+        collection.insert(std::pair<wxDateTime, const CItemData*>(dt, GetItem(row)));
```
dskrvk, line 804: Wouldn't it have been easier to serialize the dates in the "YYYYMMDDHHMM" format and sort lexicographically?

src/ui/wxWidgets/GridCtrl.cpp
rafaelx: I tested it with three different locale settings, which were set system-wide (all LC-related variables such as LC_TIME were set) to one of the following settings:

- en_US.UTF-8
- en_GB.UTF-8
- de_DE.UTF-8

The date selection field in the "Date and Time" tab of the "Add/Edit" dialog always displays the date in the format DD/MM/YY, which also appears in the "Password Expiration Date" column of the list view. It therefore seems to be independent of the language setting.

src/ui/wxWidgets/GridCtrl.cpp
rafaelx: Interesting suggestion. The current solution uses the "wxDateTime" class, which provides several helpful methods. Some methods are used to parse a string for date and time information, others for comparison. I see no advantage in creating a new, lexicographically sortable string from a "wxDateTime" object when the "wxDateTime" object already provides everything necessary for sorting. The concept of comparing "wxDateTime" objects has also proven useful for sorting the "Password Expiration Date" column in the list view. The only advantage I can think of might be that the two structures "DateTimeIsEarlierThan" and "DateTimeIsLaterThan" would no longer be needed, which only provide one operator for sorting.

dskrvk: Looks like we could use [ParseFormat](https://docs.wxwidgets.org/latest/classwx_date_time.html#a5d01d56c310d505a093f1a7f25086a1b) instead of parsing manually.

Another option, since we're already using regex, is to recombine the day/month/year strings into YYMMDD without going through an intermediate wxDateTime representation.

PR #1656, opened 2025-12-03
[wx] Fix check of attachment title

rafaelx:
When the Add/Edit dialog is closed without any changes being made, a message appears indicating that there are unsaved changes. - This should be corrected by this PR.

If an attachment has no title, "N/A" is displayed in the text input field. If the user clicks in the text input field, they must first remove the "N/A" before they can enter the desired title. This is an unnecessary extra step.

A potentially rare scenario: The user clicks in the input field, adds the desired title to the existing "N/A", and saves the changes. Now they might want to remove the title and wonder how. Should I delete my input up to the "N/A" or do I have to remove everything? The correct procedure is to remove everything, as "N/A" has no effect.

The separate handling of "N/A" in the user interface also reflects the separate handling in the source code. Furthermore, it would be difficult to distinguish between an intentionally entered "N/A" by the user and an internal initialization with "N/A" by the application.

Writing something by default into the text input field therefore conflicts somewhat with user input.
Therefore, I believe that if an attachment has no title, an empty text input field should simply be displayed.

Another observation is that when an attachment is removed from a V40 database and the Add/Edit dialog is closed, the password entry is no longer visible. This may only be an update issue, as the entry reappears after switching between views.

### DIFF
--- a/src/ui/wxWidgets/AddEditPropSheetDlg.cpp
+++ b/src/ui/wxWidgets/AddEditPropSheetDlg.cpp
@@ -1165,7 +1165,7 @@ void AddEditPropSheetDlg::DisableAttachmentControls()
   m_AttachmentButtonImport->Disable();
   m_AttachmentButtonExport->Disable();
   m_AttachmentButtonRemove->Disable();
-  m_AttachmentTitle->Enable();
+  m_AttachmentTitle->Disable();
 }
 
 void AddEditPropSheetDlg::OnImport(wxCommandEvent& WXUNUSED(event))


src/ui/wxWidgets/AddEditPropSheetDlg.cpp
```diff
@@ -1154,20 +1153,23 @@ void AddEditPropSheetDlg::EnableImport()
...
   m_AttachmentButtonRemove->Enable();
+  m_AttachmentTitle->Enable();
 }
 
 void AddEditPropSheetDlg::DisableAttachmentControls()
 {
   m_AttachmentButtonImport->Disable();
   m_AttachmentButtonExport->Disable();
   m_AttachmentButtonRemove->Disable();
+  m_AttachmentTitle->Enable();
```
nobugshere: Did you mean to call Disable here?

nobugshere:
>  I wonder if "N/A" is also displayed as a hint in the text input field on Mac.

Yes, the SetHint() works well.

nobugshere:
> Wouldn't it make more sense to simply inform the views via UpdateGUICommand::GUI_REFRESH_ENTRY to update the password element?

I haven't tested it, but that does make more sense to me.  We're not removing the entire entry, we just want it to update the paper clip icon.

src/ui/wxWidgets/AddEditPropSheetDlg.cpp
rafaelx: Yes, the text input field for the title should also be deactivated. - Thank you

nobugshere:
Other than the one edge case I described above, this all looks good to me.  I think that case is limited enough that, If you want to, you can merge all this and work that as a separate issue.

PR #1662, opened 2025-12-06
[macOS] Fix for media type filter crash

### LINKED ISSUE #1661: [MacOS] App crashes when filtering entries by Media Type
Issue: The app silently crashes on changing filter criteria for Attachment - Media Type.

Version: Password Safe 1.23 on macOS (Password Safe 1.22 too)
Database format: 4.01
OS: Tested on macOS 12 and 15, there is no issue on Linux.
Note: I was not able to test with database format 3.16 as filtering by Attachment fields is not working yet, see PR #1659 (the updated app from that PR providing this functionality crashes too with database format 3.16).

Steps to reproduce:
1. Attach a file to an entry.
2. Create a filter with this settings: Field=Attachment, click Criteria.
   In the Set Attachment Filters dialog specify this settings: Field=Media Type, Criteria=equals or contains a specific file format.
3. Apply the filter, everything works as expected, I can only see entries having attachment of that Media Type.
4. Go back to filters, select the filter, click Edit.
5. Try to change criteria by selecting another Media Type.
   Click the Criteria field once - nothing happens, click the Criteria field again - crash. 

Attached you can find a screenshot, hope it helps.

Image

nobugshere:
Resolves #1661 

The crash happens because on the second invocation it receives an extraneous EVT_TEXT event before the controls are created.  Apparently, PWFiltersStringDlg had a similar problem, so I borrowed its solution.

Tested with a v4 DB.  I'll test it with a v3 DB when #1659 is merged.

### DIFF
--- a/src/ui/wxWidgets/HelpMap.h
+++ b/src/ui/wxWidgets/HelpMap.h
@@ -68,6 +68,7 @@ DLG_HELP(pwFiltersTypeDlg,                 html/filters.html)
 DLG_HELP(pwFiltersDateDlg,                 html/filters.html)
 DLG_HELP(pwFiltersMediaDlg,                html/filters.html)
 DLG_HELP(pwFiltersPasswordDlg,             html/filters.html)
+DLG_HELP(pwFiltersMediaTypesDlg,           html/filters.html)
 
 DLG_HELP(SelectAliasDlg,                   html/aliases.html)
 


nobugshere:
Works with a v3.16 DB as well. Tested on:
macOS M1Max Sequoia 15.7.2, Xcode 16.4, wx 3.2.8 (local build)

PR #1672, opened 2025-12-13
[wx] Fix - Disable changing attachment of protected entries

froggie2523:
This PR disables Attachment tab controls for protected entries to match the Windows version's behavior.

Version: Password Safe 1.23
Issue: If an entry is protected, the Attachment tab controls are not disabled. User can delete the attachment, import a new one and change title. Naturally, all changes are lost after closing the dialog as the entry is protected.
Tested on Debian/FreeBSD/OpenBSD.

Attached you can find some screenshots from both Windows and Linux.

Windows:

pwsafe_1 23-wxWidgets-bug-Protected-Edit-Att-03

Linux:

pwsafe_1 23-wxWidgets-bug-Protected-Edit-Att-01

When fixed:

pwsafe_1 23-wxWidgets-bug-Protected-Edit-Att-02

### DIFF
--- a/src/ui/wxWidgets/AddEditPropSheetDlg.cpp
+++ b/src/ui/wxWidgets/AddEditPropSheetDlg.cpp
@@ -937,8 +937,13 @@ void AddEditPropSheetDlg::InitAttachmentTab()
         DisableAttachmentControls();
       }
       else {
-        // Attachment must be removed before a new one can be imported again.
-        DisableImport();
+        if (m_Item.IsProtected()) {
+          EnableExport(); // Disable all controls except the Export button
+        }
+        else {
+          // Attachment must be removed before a new one can be imported again.
+          DisableImport();
+        }
       }
     }
     else {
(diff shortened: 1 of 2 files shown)


src/ui/wxWidgets/AddEditPropSheetDlg.cpp
```diff
@@ -982,8 +987,13 @@ void AddEditPropSheetDlg::InitAttachmentTab()
         DisableAttachmentControls();
```
dskrvk, line 987: I'm wondering why we don't allow export of read-only entries. If we allowed it, then both r/o and protected cases could share the logic. If anything, it makes more sense to disable export for protected entries than read-only.

ronys: Enabling export of an attachment when read-only makes perfect sense. This is consistent with allowing export in general when read-only. This should be a separate PR though, for both Windows and wx.

ronys:
Nice catch - thanks!

Issue #1667, opened 2025-12-09
Empty groups count wrong in File->Properties


ronys:
The attached database has one group with one entry. However, under wx version 1.23, the Properties dialog insists that it's an empty group:

<img width="225" height="143" alt="Image" src="https://github.com/user-attachments/assets/e885ff8b-c6d7-4613-8624-c9d3daf56880" />

<img width="397" height="207" alt="Image" src="https://github.com/user-attachments/assets/58dbadf2-2331-4069-9e4d-9752c8e4ff5f" />

Perhaps it has to do with how I created this: I created the group via right-click Add Group, and then I edited the "when" entry, setting its group in that Add/Edit dialog.

[pr1653.zip](https://github.com/user-attachments/files/24060619/pr1653.zip) - zipped because GH doesn't support uploading .psafe3 files. Password Sample123

The Windows version displays the correct values for the same database.

rafaelx:
I see that the Edit command is missing the option to delete an empty group.

rafaelx:
If a password item leaves an empty group, how can this be determined using the Core in the Add/Edit dialog, which has no knowledge of the tree structure. Does the Core offer any helpful functions in this regard?

PR #1673, opened 2025-12-13
[GH1677] Fix handling of group changes via entry edit

ronys:


src/ui/wxWidgets/AddEditPropSheetDlg.cpp
```diff
@@ -2845,8 +2845,36 @@ Command* AddEditPropSheetDlg::NewEditEntryCommand()
...
+    m_Item.SetGroup(newGroup);
+    // if we're putting an entry into a previously empty group, it's no longer empty...
+    if (m_Core.IsEmptyGroup(newGroup)) {
+      commands->Add(DBEmptyGroupsCommand::Create(&m_Core, newGroup, DBEmptyGroupsCommand::EG_DELETE));
+    }
+    // Conversely, if we're moving an entry that was alone in a group, that group's now empty
+    if (!oldGroup.empty()) {
+      // There has to be a better way to do this...
+      int ngroup = 0;
+      for (auto iter = m_Core.GetEntryIter(); iter != m_Core.GetEntryEndIter(); ++iter) {
```
rafaelx, line 2866: I was hoping there was a simpler way than iterating over all the database entries. That's why I asked about it in issue #1667. Wouldn't this algorithm be a good candidate for the Core, so that other code locations and platforms could also benefit from it?

dskrvk: Would it be useful to maintain an in-memory map of group to list of elements? Maybe other places could benefit from this too.

PR #1675, opened 2025-12-14
[win] Fix - Clear list of recently opened databases

froggie2523:
Version: Password Safe 3.70.1

Options > System: Uncheck "Recent Databases on File Menu rather than as a sub-menu."
Command: Main menu > File > Clear Recently Opened list
Issue: Regardless of how many items are in the list, only the first item is removed. This issue does not occur when recent databases are displayed directly on the File menu.
Reason: The sub-menu is not referenced correctly.

pwsafe_3 70 01-bug-Clear-MRU-01
pwsafe_3 70 01-bug-Clear-MRU-02
pwsafe_3 70 01-bug-Clear-MRU-03

### DIFF
--- a/src/ui/Windows/ThisMfcApp.cpp
+++ b/src/ui/Windows/ThisMfcApp.cpp
@@ -1273,18 +1273,19 @@ void ThisMfcApp::ClearMRU()
   }
 
   // Recent databases are on the popup menu off the main File menu
-  CMenu *pPopupMenu = pFile_Submenu->GetSubMenu(FindMenuItem(pFile_Submenu, ID_MENUITEM_LOCK) + 2);
-  if (pPopupMenu != NULL)  // Look for MRU first entry
-    pos = FindMenuItem(pPopupMenu, ID_FILE_MRU_ENTRY1);
-  else
-    return;
-
-  if (pos > -1) {
-    // Recent databases are on a popup menu - remove them
-    for (int nID = numMRU; nID > 1; nID--)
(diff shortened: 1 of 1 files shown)


src/ui/Windows/ThisMfcApp.cpp
```diff
@@ -1273,7 +1273,7 @@ void ThisMfcApp::ClearMRU()
   }
 
   // Recent databases are on the popup menu off the main File menu
-  CMenu *pPopupMenu = pFile_Submenu->GetSubMenu(3);
+  CMenu *pPopupMenu = pFile_Submenu->GetSubMenu(FindMenuItem(pFile_Submenu, ID_MENUITEM_LOCK) + 2);
```
ronys: 1. I wish there was a more elegant/robust solution to finding the menuItem instead of assuming it's after the Lock menu item, but I can't think of one
2. Where does "+ 2" come from? It works, but I don't understand why "+ 1" doesn't work?
3. If FindMenuItem fails to find an item, it will return -1. This will mean GetSubMenu() will be called with -1 + 2 = 1, which will not be correct. Perhaps an ASSERT on FindMenuItem() would be better protection against future menu changes?

dskrvk: I think a slightly safer assumption is that the menu immediately precedes the "Clear Recently Opened List" entry.

froggie2523: OK, updated. As for the position evaluation, I've seen this code in ThisMfcApp::SetupMenu().
It's "+ 2" because there is also a separator on the menu.

PR #1676, opened 2025-12-14
[win] Fix - GetAttachmentSymbol() return value mismatch

froggie2523:
Version: Password Safe 3.70.1

Issue:
I noticed the GetAttachmentSymbol function currently uses "ProtectSymbol" variables, but based on the surrounding logic and naming, it appears "AttachmentSymbol" was intended. Most probably a copy&paste error.
The current implementation works because m_bProtectSymbolSupportedXY and m_bAttachmentSymbolSupportedXY are both True in practice, but this assumption is implicit. If m_bProtectSymbolSupportedXY and m_bAttachmentSymbolSupportedXY ever diverge, this function would silently behave incorrectly.


ronys:
Excellent catch - thanks!

PR #1678, opened 2025-12-15
[wx] Simplify sorting by expiration date

rafaelx:
Follow up of #1655 

@dskrvk Thanks for the hint about wxDateTime::ParseFormat. This significantly simplifies the implementation. What do you think about the new solution?


rafaelx:
@ronys Could you please tell me your locale settings? In my case, the date always appeared in the format dd/mm/yy, even with other locale settings.

PR #1680, opened 2025-12-18
[help] Updates to Authenticator Secret on Add/Edit - Additional tab

froggie2523:
Support for entering a TOTP secret on the Additional tab in the Add/Edit dialog is a great feature, and I think it deserves more detailed coverage in the Help so users can fully take advantage of it.
For example, some users might not know how to obtain an otpauth URI or which part of it to copy and paste into Password Safe, mainly because websites prefer to set it up using an Authenticator app on a phone.
Personally, I have only recently started using it on a daily basis, after figuring out its specifics.

pwsafe_3 70 01-help-TOTP-01

Help page now:

pwsafe_3 70 01-help-TOTP-02

Help page updated:

pwsafe_3 70 01-help-TOTP-03


ronys:
Thanks. The current description is indeed a bit terse, but I think the proposed text can be cleaned up a bit.

The main problem is that most sites use the QR code as the default way for users to get the key, which is problematic for a PC user.
In my experience, all sites have something like the "setup key" link in your example, which shows the key so that it can be copy/pasted to PasswordSafe.

I suggest rephrasing the text to use such a link where available, and use a phone to scan the QR code only as the last resort.

froggie2523:
Happy New Year :-)
I’ve rephrased the text, please review again.

ronys:
Thanks - happy new year!

PR #1681, opened 2025-12-25
[macOS] Typo in file type definition

nobugshere:
I don't think it caused any issues, but now it's "more correct".

### DIFF
--- a/Xcode/pwsafe-template.plist
+++ b/Xcode/pwsafe-template.plist
@@ -63,7 +63,7 @@
 			<key>UTTypeConformsTo</key>
 			<array>
 				<string>public.data</string>
-				<string>public.contents</string>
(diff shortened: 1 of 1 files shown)


ronys:
Can this cause any backwards compatibility issues when users upgrade?

nobugshere:
> Can this cause any backwards compatibility issues when users upgrade?

I don't think so.  Starting with a fresh VM (macOS 15 Sequoia), I tried upgrading and downgrading in various sequences and it "Just Worked".  These type associations are imported to a database that is updated whenever an app is installed or deleted.  To the best of my understanding, any new type definition needs to be traceable to one or more standard base types in the hierarchy.  I accidentally created a new base type instead of using the proper one, but we aren't really doing anything that relies on that particular association.

PR #1682, opened 2025-12-27
Switch to C++17, fix incremental backup across devices (Unix)

### LINKED ISSUE #1670: Can't create backups across different filesystems
From Fred:

> I installed PW Safe in MX Linux 25 and have one puzzling issue.
I have the physical hard drives… one for Linux, one for app data (such as My Safe) and one for “backup”.
I configure PW Safe to save on my app disk, no issue.
I “manually” have PW Safe to create a backup on my backup disk, no issue.
But PW Safe fails to create an .ibak backup to the same backup disk, same directory, as if there is a permissions issue?  But how can it be due to permissions when “manual” works?
Oh, and .ibak works fine if I use the same directory on my apps disk?

Looks like move/rename across filesystems is an issue on Linux.

ronys:
Under Unix, pws_os::RenameFile(), which is used in incremental backups, failed when the backup was set to be on another device. (resolves #1670)

The fix is to fallback to copy-then-delete. Since C++17 supports this in std::filesystem, it seems like a good opportunity to switch to this across the board.

Note that the original reason for staying with C++14 was due to a problem with wxWidgets 3.2.1, and the current (Debian 13) version is 3.2.8, so hopefully this is moot.

### DIFF
--- a/src/os/unix/file.cpp
+++ b/src/os/unix/file.cpp
@@ -22,6 +22,7 @@
 #include <cerrno>
 #include <filesystem>
 #include <system_error>
+#include <memory>
 
 #include <dirent.h>
 #include <fnmatch.h>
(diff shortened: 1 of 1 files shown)


ronys:
Let's try in a branch and see if it breaks anything on the major distros we currently support.

PR #1691, opened 2026-01-10
[GH1685] Drag & Drop support for attachments

### LINKED ISSUE #1685: Attachments are not transferred during Drag & Drop operations
I love the Drag & Drop feature, but it does not fully work in v3 Safes that have recently received support for attachments.

It appears to work only from DBv4 to DBv4 in the wxWidgets version. It does not work in the following scenarios:
From DBv4 to DBv4 (Windows version)
From DBv3 to DBv3 (both wx and Windows versions)
From DBv3 to DBv4 (both wx and Windows versions)
From DBv4 to DBv3 (both wx and Windows versions; a warning is displayed in the wx version - see the screenshot below)

Image

ronys:
Resolves #1685, at least partially.

Currently added support for V3 to V3 drag & drop with correct attachment handling

### DIFF
--- a/src/core/ItemData.cpp
+++ b/src/core/ItemData.cpp
@@ -2350,6 +2350,7 @@ void CItemData::SerializePlainText(vector<char> &v,
     unsigned char *pdata = nullptr;
     GetUnknownField(type, length, pdata, *vi_IterURFE);
     push(v, static_cast<char>(type), static_cast<uint32>(length), reinterpret_cast<char*>(pdata));
+    trashMemory(pdata, length);
     delete[] pdata;
   }
 


src/core/ItemData.cpp
```diff
@@ -2322,6 +2322,30 @@ void CItemData::SerializePlainText(vector<char> &v,
...
+  {
+    time_t att_mtime = 0;
+    CItem::GetTime(DATA_ATT_MTIME, att_mtime);
+    if (att_mtime != 0)
+      push(v, DATA_ATT_MTIME, att_mtime);
+  }
+  {
+    std::vector<unsigned char> content = GetAttContent();
+    if (!content.empty()) {
+      v.push_back(static_cast<char>(DATA_ATT_CONTENT));
```
dskrvk: Similar code is used below on lines 2357-61 and IIRC in other places that handle attachments. Would it make sense to define a helper method?

ronys: Thanks. Addressed in commit 755d12e

src/core/ItemData.cpp
```diff
@@ -2353,13 +2349,7 @@ void CItemData::SerializePlainText(vector<char> &v,
     size_t length = 0;
     unsigned char *pdata = nullptr;
     GetUnknownField(type, length, pdata, *vi_IterURFE);
-    if (length != 0) {
-      v.push_back(static_cast<char>(type));
-      push_length(v, static_cast<uint32>(length));
-      v.insert(v.end(), reinterpret_cast<char *>(pdata),
-               reinterpret_cast<char *>(pdata) + length);
-      trashMemory(pdata, length);
```
dskrvk, line 2337: Looks like this `trashMemory` got lost in the refactoring? (I originally thought it should be in the helper method, but looks like not all places need it.)

ronys: Nice catch - thanks!

PR #1694, opened 2026-01-21
Added field 0x30 (JSON object)

medo64:
Since PWSafe file format is used by other password managers, there are often custom fields defined (e.g. 0x20 QR Code, etc.). However, this means that every time some other application is expanded it needs to add entry to fields even though it's highly unlikely that PWSafe will ever implement it due to a different focus or philosophy.

 Possible alternative is using dedicated implementation-specific fields but these have no format specified and might contain different data between different applications leading to wrong data parsing or crashes.

I propose adding a new field (`JSON Object 0x30`) that would contain a valid JSON thus at least removing potential for crashes. In addition, even if application doesn't recognize the object, it could just show it to the user as-is and allow for cross-readability of custom fields.

For my application I already have three scenarios where this would fit without forcing a new field type:

1. Android lock pattern - e.g., `L` shaped unlock pattern could be described with
~~~json
{
    "app-id": {
        "type": "pattern",
        "pattern": "74123"
    }
}
~~~

2. Field with custom name:
~~~json
{
    "app-id": {
        "type": "text",
        "caption": "SWIFT Code",
        "text": "CHASUS01"
    }
}
~~~

3. Field with custom tooltip text:
~~~json
{
    "app-id": {
        "type": "text",
        "caption": "VPN PIN", "text": "1234",
        "tooltip": "Type it on second page after password has been confirmed"
    }
}
~~~

PS: Added namespace in examples as suggested by @huven.

### DIFF
--- a/docs/formatV3.txt
+++ b/docs/formatV3.txt
@@ -190,7 +190,7 @@ password change             0x13        time_t        Y              [5]
 End of Entry                0xff        [empty]       Y              [17]
 
 [1] The version number of the database format. For example, the value
-0x0310 is stored in little-endian format, that is, 0x10, 0x03.
+0x0311 is stored in little-endian format, that is, 0x11, 0x03.
 PasswordSafe V3.01 introduced Format 0x0300
 PasswordSafe V3.03 introduced Format 0x0301
 PasswordSafe V3.09 introduced Format 0x0302
@@ -705,6 +705,8 @@ For example, to encode the following custom fields:
 * WiFi Password: bobsyouruncle (sensitive field)
 One would use
     010009WiFi SSID020004Test00000001000DWiFi Password02000Cbobsyouruncle0300011
+
+This field was introduced in format version 0x0311
 
 4. Extensibility
 
--- a/docs/formatV4.txt
+++ b/docs/formatV4.txt
@@ -1,4 +1,4 @@
-PasswordSafe database format description version 4.01
+PasswordSafe database format description version 4.02
 -----------------------------------------------------
 
 Copyright (c) 2013-2026 Rony Shapiro <ronys@pwsafe.org>.
@@ -261,7 +261,7 @@ password change             0x13        time_t        Y              [5]
 End of Entry                0xff        [empty]       Y              [17]
 
 [1] The version number of the database format. For this version, the value
-is 0x0401 (stored in little-endian format, that is, 0x01, 0x04).
+is 0x0402 (stored in little-endian format, that is, 0x02, 0x04).
 PasswordSafe V3.69 introduced Format 0x0401
 
 [2] A universally unique database identifier is needed in order to
@@ -694,6 +694,8 @@ For example, to encode the following custom fields:
 * WiFi Password: bobsyouruncle (sensitive field)
 One would use
     010009WiFi SSID020004Test00000001000DWiFi Password02000Cbobsyouruncle0300011
+
+This field was introduced in format version 0x0402
 
 3.4 Attachments
 


huven:
If multiple applications start using this field, there is a risk of overwriting attributes of another application.

Could be solved by agreeing on namespaces like

```
{
  "app-id-1": { ... },
  "app-id-2": { ... },
  ...
}
```

with the "app-id-x" arbitrary but unique application identifiers.

ronys:
Initially I  liked this idea, but thinking abut it a bit more, I have a few reservations:
1. The added flexibility is confusing: I want to add a new field - do I use a record field like today or do I extend the JSON?
2. There are already fields that aren't used by all platforms, so this shouldn't be a factor.
3. As @huven pointed out, we need to manage JSON namespace collisions. More generally, we need to manage a JSON schema in addition to the current format.
4. One part that I really liked was the ability to define a "caption", but doing this at this level breaks the current i18n support.

So perhaps we can define some field values for "user-defined" fields, and store these in JSON. The user can then add fields with type, value, captions and tooltips in his/her preferred language.
By allowing the user to define these, or select from some predefined templates, I think we can get the best of both worlds.

What say?

huven:
I'll throw in a third variant

```
{
  "VPN PIN": { "value": "1234" },
  "WiFi Password": { "value": "bobsyouruncle" },
  "SSH Password": { "value": "nohesnot" }
}
```

also taking the field name as object key (which has the advantage of avoiding duplicate names), and storing the value inside an object allowing to extend it in the future with e.g. the "type" field to support dates etc., the "tooltip" mentioned earlier, or a "this is a password" boolean to allow the UI to hide the string with **** by default.

docs/formatV3.txt
```diff
@@ -669,20 +669,34 @@ of the contents.
...
-be set to true.
+[32] This is a custom text field that allows for configuring custom properties.
+Key/value pairs are stored in the following format:
+    PLVPLV...PLV S PLVPLV...PLV ... S PLVPLV...PLV
+where:
+    P = 2 hexadecimal digits indicating property ID
+    L = 4 hexadecimal digits indicating value length in bytes
+    V = value of the property whose interpretation depends on the P
+    S = 6 hexadecimal digits 000000 that serve as a separator between different
+        properties
```
huven: The term "property" is used for two different entities, like on line 678 it refers to one PLV, while on line 680 it refers to PLVPLV...PLV.

Maybe use "custom field" when referring to PLVPLV...PLV ?
The "property entry" used further would then also become "custom field".

In theory you could use "property entry" and "property" consistently to separate the two concepts, though I think that would be hard to read.

medo64: I agree. I reworded text it a bit to clarify.

huven: Thanks, that's better. There is some room left for improving naming consistency, like I see "custom properties" (line 672), "custom field entry" (line 694) and "custom entry" (line 698), which I assume all refer to the same concept as "custom field" so I would replace these with "custom field" to avoid any misinterpretation.

medo64: I agree. Please check now if it flows better.

docs/formatV3.txt
```diff
@@ -668,6 +669,33 @@ of the contents.
 
 These parameters were introduced in version 0x0310 (PasswordSafe V3.69).
 
+[32] Custom fields consist of property entries stored in the following format:
+    PLVPLV...PLV S PLVPLV...PLV ... S PLVPLV...PLV
+where:
+    P = 2 hexadecimal digits indicating property ID
+    L = 4 hexadecimal digits indicating value length in bytes
```
ronys: please add ", stored in little-endian format." for consistency with other such fields.

docs/formatV3.txt
```diff
@@ -668,6 +669,33 @@ of the contents.
...
+    V = value of the property whose interpretation depends on the P
+    S = 6 hexadecimal digits 000000 that serve as a separator between different
+        custom fields
+
+These custom fields repeat as many times as there are properties.
+
+Currently defined property IDs are
+    01 = name: property value is interpreted as a UTF-8 string; in application
+         this can be considered to be a title of a given value
```
ronys, line 688: Do we want to allow empty names, e.g., 010000?

docs/formatV3.txt
```diff
@@ -668,6 +669,33 @@ of the contents.
...
+    02 = value: property value is interpreted as a UTF-8 string; in application
+         this can be considered to be a value displayed to the user
+    03 = sensitivity flag: property value is interpreted as a boolean {0,1} with
+         literal 0 being false and literal 1 being true; in application this
+         might change how value is displayed (e.g., sensitive value could have
+         its characters masked)
+
+At minimum, each custom field must contain property name (01) and property value
+(02). Property names must not repeat between custom fields.
```
ronys: I propose "For a given custom text field, all property names must be unique." instead of the last sentence.

docs/formatV3.txt
```diff
@@ -668,6 +669,33 @@ of the contents.
...
+        custom fields
+
+These custom fields repeat as many times as there are properties.
+
+Currently defined property IDs are
+    01 = name: property value is interpreted as a UTF-8 string; in application
+         this can be considered to be a title of a given value
+    02 = value: property value is interpreted as a UTF-8 string; in application
+         this can be considered to be a value displayed to the user
+    03 = sensitivity flag: property value is interpreted as a boolean {0,1} with
```
ronys, line 691: Do we want to restrict the length of this field to 1?
What is "literal" here? '0', '1' (ASCII)?  0x0/0x1?
Suggest noting that any non-zero value will be treated as 'true'.

docs/formatV3.txt
huven: Perfect, my brain stopped complaining 😄 

I'm not sure if the 0x30 field should be marked Text or Binary, I'll leave that up to @ronys to decide.
With Text people might be tempted to store it in a String type variable which can break if V is allowed to contain binary data. The boolean (03) is already using 0 and 1 bytes.

docs/formatV3.txt
```diff
@@ -668,6 +669,33 @@ of the contents.
 
 These parameters were introduced in version 0x0310 (PasswordSafe V3.69).
 
+[32] Custom fields consist of property entries stored in the following format:
```
ronys, line 672: The terminology here's inconsistent: 
- Custom Text field
- Custom Field
- Property Entries

The first two should be unified, and the last should be defined, perhaps using an example?

docs/formatV3.txt
```diff
@@ -668,6 +669,33 @@ of the contents.
...
+    03 = sensitivity flag: property value is interpreted as a boolean {0,1} with
+         literal 0 being false and literal 1 being true; in application this
+         might change how value is displayed (e.g., sensitive value could have
+         its characters masked)
+
+At minimum, each custom field must contain property name (01) and property value
+(02). Property names must not repeat between custom fields.
+
+Any unknown property ID must be ignored and left unchanged for forward
+compatibility. Duplicate property IDs are not allowed within one custom field.
```
ronys: Custom field or property entry? I think you meant the latter.

docs/formatV3.txt
huven: Length = 1 makes sense to me. I read the sentence as encoding false to 0x0, feels most natural to me. Specifying it explicitly is a good idea.

docs/formatV3.txt
medo64: > Perfect, my brain stopped complaining 😄

The only question is whether it stopped complaining because I got it right or because you gave up on me :)

> I'm not sure if the 0x30 field should be marked Text or Binary

My thoughts here are to follow the PasswordHistory (`L  = 4 hexadecimal digit password length`) and mark it Text, but I am pretty much on the fence for this one.

docs/formatV3.txt
medo64: > Do we want to allow empty names, e.g., 010000?

I would think yes - just for simplification but I don't have any use-case on my side to justify it.

huven: Personally I would opt for non-empty names. The name is like an identifier for the custom field, and while technically the empty string is an identifier, it is a weird one that can only cause trouble. For instance in JS the identifier suddenly evaluates to false.

docs/formatV3.txt
```diff
@@ -668,6 +669,40 @@ of the contents.
...
+(02). For a given custom text field, all property names must be unique.
+
+Any unknown property ID must be ignored and left unchanged for forward
+compatibility. Duplicate property IDs are not allowed within one property entry.
+
+For example, to encode the following custom fields:
+* WiFi SSID: Test
+* WiFi Password: bobsyouruncle (sensitive field)
+One would use
+    010009WiFi SSID020004Test00000001000DWiFi Password02000Cbobsyouruncle0300011
```
huven, line 707: The length bytes (L in PLV) need to be swapped, so the string starts with
010900WiFi.....

And the last '1' was intended to be 01 I think, but anyway, will change to 0x30 probably (see other comment)

medo64: Actually, I think I made an error in text and in my head when I added little-endian statement. Based on password history entry [12], I believe hexadecimal values should always be big-endian, i.e. hex numbers. Endianness makes sense when discussing binary numbers. However, since we moved to hexadecimal (same as password history), I don't believe little-endian makes sense.

So, I would argue that example is correct - it's mention of endianess that should be removed.

medo64: > Please add at least one such property to the example too, can be just add the end, like for sensitive=false: 03010030

This is already one of those added to an example - it's 0300011 (true), since it's ASCII encoding, like other fields, `31` is not really needed as this would be hexadecimal notation.

docs/formatV3.txt
medo64: @ronys: Since this is hexadecimal representation, I don't believe it makes sense to have it little-endian.

I again borrowed from Password History [12] where there is parameter L (`L  = 4 hexadecimal digit password length (in TCHAR)`). I know from code that this is numeric hexadecimal and thus it is parsed as big-endian.

docs/formatV3.txt
huven: Yeah, my last statement about the 1 was a mistake, no hex-encoding for the boolean value, fully correct.

I'm not sure about the endianness of the hex numbers in the password policy, can't check right now. Consistency there would be best too.

docs/formatV3.txt
```diff
@@ -668,6 +669,40 @@ of the contents.
 
 These parameters were introduced in version 0x0310 (PasswordSafe V3.69).
 
+[32] Custom fields consist of property entries stored in the following format:
+    PLVPLV...PLV S PLVPLV...PLV ... S PLVPLV...PLV
+where:
+    P = 2 hexadecimal digits indicating property ID
```
huven: I think it is best to explicitly state that P=00 is not allowed to make the separator unique. It's obvious, but explicit is better.

docs/formatV3.txt
```diff
@@ -668,6 +669,40 @@ of the contents.
...
+    L = 4 hexadecimal digits value length in bytes
+    V = value of the property whose interpretation depends on the P
```
huven: As both strings and the boolean are now encoded as string, and the 0x30 is marked as Text, I think it makes sense to require _all_ V to be valid UTF8 as you suggested earlier. This also aligns with the password history field.

That will also simplify implementing this. So I suggest adding that requirement here.

docs/formatV3.txt
```diff
@@ -668,6 +669,40 @@ of the contents.
...
+    S = 6 hexadecimal digits 000000 that serve as a separator between different
+        custom fields
```
huven, line 680: Should we add a comment whether a separator is allowed (or not) at the end of the 0x30 field?

ronys: To me "between different custom fields" implies that it's not needed at the end.
Of course, the implementation should adhere to the robustness principle, and accept (and silently ignore) a separator at the end.

docs/formatV3.txt
ronys: I've check, and the endianness of hex values is a non-issue, no need to specify.

docs/formatV3.txt
huven: Agree. Still, I would document this explicitly to make other implementors aware, something like: A separator is allowed, but not required, after the last custom field.

medo64: Added

docs/formatV3.txt
medo64: All updated as suggested.

PR #1696, opened 2026-01-26
Added Custom Fields support

huven:
Implementation of #1694


src/core/ItemData.cpp
```diff
@@ -1599,6 +1604,12 @@ void CItemData::SetPWHistory(const StringX &PWHistory)
   CItem::SetField(PWHIST, pwh);
 }
 
+void CItemData::SetCustomFields(const CustomFieldList &fields)
+{
+  StringX data = fields;
```
ronys: Can be const or const &

src/core/CustomFields.h
```diff
@@ -0,0 +1,80 @@
...
+
+  const std::vector<CustomFieldProperty> &GetProperties() const { return m_props; }
+  bool Empty() const { return m_props.empty(); }
+
+private:
+  std::vector<CustomFieldProperty> m_props;
+};
+
+class CustomFieldList : public std::vector<CustomField>
+{
```
ronys, line 57: This is fine for serializing/deserializing the representation to/from the file.
I think it would be useful for the UI to be able to add/edit/delete a single custom field via member classes.
Add would check for an empty name and existing name.
Edit would replace the value of a given property, etc.

Since this would be common to Windows and wx (and cli?), this should be here, or in CItemData.
What say?

src/core/ItemData.cpp
huven: Changed into `const StringX &data`. The StringX will still be saved to a temporary, data will now reference the temporary instead of moving from it.

src/core/CustomFields.h
huven: Agree, I was still struggling with that API, in particular where to put the validations on setting new/updated data.

Will add the calls, they feel natural to have. They need to return bool to flag any errors so the application can respond to the error. So the application can either use these "checked" calls and rely on core to validate the data, or provide an updated CustomFieldList in which case the application is responsible for pre-validating the data.

src/core/ItemData.cpp
huven: Commit 691ee07f7

src/core/CustomFields.h
huven: Added in commit 110be1d

huven: Note: for now these calls always fail if GetCustomFields() reports an error, to avoid silently dropping malformed data.

ronys:
Have you considered deriving CustomFieldProperty from `std::pair<unsigned char, StringX>` and using `std::map<unsigned char, StringX>` for `m_props`?

huven:
For duplicate names, I'd normally just add #&NoBreak;1, #&NoBreak;2, ... until it's unique, leaving the user's original intent. More or less what most filesystems do with duplicate filenames.

src/core/CustomFields.cpp
```diff
@@ -0,0 +1,254 @@
...
+#include "CustomFields.h"
+#include "UTF8Conv.h"
+#include "StringXStream.h"
+
+#include <iomanip>
+#include <unordered_set>
+
+static bool ReadHexValue(const unsigned char *data, size_t len, unsigned int &value)
+{
+  StringX hexstr;
```
ronys, line 20: Why work via a StringX and not just use `std::basic_istringstream<unsigned char> iss(data)`?

src/test/ItemDataTest.cpp
```diff
@@ -270,6 +270,85 @@ TEST_F(ItemDataTest, PasswordHistory)
   EXPECT_EQ(emptyHeader, PWHistList::MakePWHistoryHeader(false, 0, 0));
 }
 
+TEST_F(ItemDataTest, CustomFields)
+{
+  const StringX raw = _T("010004Name020005Value0300012")
```
ronys, line 275: An interesting edge case would be one where the value and/or the name would be "000000", no?

huven: Sure, let me add both. The one you quote has sensitivity = '2', so that is why the normalized was added, since the new repair maps it to '1' now.

huven: Added... let's see how the checks go...

src/core/CustomFields.cpp
huven: Afaik basic_istringstream is not defined for unsigned char. I tried to match existing patterns in other code.

Happy to change of course if you want.

huven:
Squashed all commits into one, no other changes

src/core/CustomFields.cpp
ronys: Right. It appears that std::istringstream requires a copy to std::string or the like (StringX). It's possible to convert without copy, e.g., using std::from_chars, but I'm not sure it's worth the effort.
Let's leave it.

PR #1707, opened 2026-02-17
First shot at custom field support for Windows

ronys:
Still need to add code to prevent duplicate titles in Add and Edit.

### DIFF
--- a/docs/ReleaseNotes.md
+++ b/docs/ReleaseNotes.md
@@ -4,12 +4,22 @@ https://pwsafe.org/. Details about changes to older releases may be found in the
 
 In the following, SFxxxx refers to Bug Reports, Feature Requests and Service Requests in PasswordSafe SourceForge Project tickets, and GHxxxx refers to issues in the PasswordSafe GitHub project.
 
-PasswordSafe 3.71.0 Release ??? ?? 2026
-=======================================
+PasswordSafe 3.71.0pre Release March 20 2026
+============================================
 
-Bugs Fixed in 3.71.0
---------------------
+Bugs Fixed in 3.71.0pre
+-----------------------
+* [SF1623](https://sourceforge.net/p/passwordsafe/bugs/1623/) File paths may be enclosed in double-quotes in "Master Password Entry" and File->Compare dialogs.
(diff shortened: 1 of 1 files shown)


src/ui/Windows/AddEdit_Basic.cpp
```diff
@@ -1317,6 +1345,168 @@ void CAddEdit_Basic::OnSendEmail()
...
+  CPoint ptScreen;
+  GetCursorPos(&ptScreen);
+  menu.TrackPopupMenu(TPM_LEFTALIGN | TPM_LEFTBUTTON, ptScreen.x, ptScreen.y, this);
+  *pResult = 0;
+}
+
+void CAddEdit_Basic::OnNMDblclkCustomFieldsList(NMHDR *pNMHDR, LRESULT *pResult)
+{
+  OnCustomFieldsEdit();
+  *pResult = 0;
```
dskrvk, line 1490: Should we propagate the return value of OnCustomFieldsEdit into pResult?

src/ui/Windows/AddEdit_Basic.cpp
```diff
@@ -1317,6 +1345,168 @@ void CAddEdit_Basic::OnSendEmail()
...
+}
+
+int CAddEdit_Basic::GetCustomFieldListSelectedSubItem(NMHDR *pNMHDR, CPoint &pt)
+{
+  (void)pNMHDR;
+  m_customFieldsList.ScreenToClient(&pt);
+  LVHITTESTINFO hit = {};
+  hit.pt = pt;
+  m_customFieldsList.SubItemHitTest(&hit);
```
dskrvk, line 1499: Should we check the return value of SubItemHitTest like we do above?

src/ui/Windows/AddEdit_Basic.cpp
ronys: Don't see a reason to. The main change is to mark the entry as modified, and this is done there.

src/ui/Windows/AddEdit_Basic.cpp
ronys: Not sure I understand the question?

PR #1724, opened 2026-03-23
Copy custom value on click

ronys:
MrMe has pointed out that there's no easy way to copy a custom value to the clipboard. This PR addresses that by copying a value to the clipboard when the row is selected. This is the same behavior that's already implemented in the password history table in the 'Additional' tab.

Note that this PR also adds a new translated text: "Custom Value Copied", using machine translation.


froggie2523:
What is the src/ui/Windows/WZFinish-ai.cpp file? Looks like a draft of an enhanced version of WZFinish.cpp with Back button for only the export operations.

src/ui/Windows/I18N/pos/pwsafe_ru.po
```diff
@@ -3,22 +3,20 @@
 # This file is distributed under the same license as the PACKAGE package.
 # FIRST AUTHOR <EMAIL@ADDRESS>, YEAR.
 #
-# Translators:
-# ronys, 2025
-# Dmitr Y, 2025
-
+#, fuzzy
```
dskrvk, line 6: Why so many unrelated changes to this file?

ronys: Part of the release process runs the resource-to-po update tool to make sure there aren't any new strings that I missed.

This tool really needs to be fixed so as not to make such meaningless changes. Added to my to-do list...

dskrvk: Is there an issue for this already? With some pointers I might be able to look into it.

ronys: No ticket for this. The update tools (under src\Tools\Windows\I18N) were written / adopted from another project (TortoiseSVN!) by someone who's no longer with us. That person wrote some documentation that you can find at https://sourceforge.net/projects/passwordsafe/files/PreReleases/Translations%20for%20Password%20Windows%20MFC%20Version%20V1.02.rtf/download

A few years ago I fixed a painful performance issue, but since then haven't touched this area.

Let me know if you need more info.

PR #1736, opened 2026-04-03
Custom field fixes

ronys:
Fix a minor bug, add custom field support where it was missing from the 3.71.0 release.


src/ui/Windows/resource3.h
```diff
@@ -504,6 +504,8 @@
 #define IDS_NO_ENTRIES_SELECTED         5736
 #define IDS_NO_ENTRIES_PROCESSED        5737
 #define IDS_DCALONG                     5738
+#define IDS_HIDE_VALUE                  6075
+#define IDS_SHOW_VALUE                  6076
```
dskrvk: Shouldn't these entries be ordered by the numeric value?

src/core/XML/XMLFileHandlers.h
```diff
@@ -62,6 +63,13 @@ struct pwhistory_entry {
   StringX oldpassword;
 };
 
+struct custom_field_entry {
+  StringX name;
+  StringX value;
+  bool sensitive;
+  bool has_sensitive;
```
dskrvk, line 70: Why need both sensitive and has_sensitive?

ronys: This is because "sensitive" is defined as optional in the format and in the xsd schema.
'has_sensitive' means that the element was found in the XML file being read, and 'sensitive' is the actual value.
(if 'sensitive' isn't specified, it defaults to false)

src/ui/Windows/resource3.h
ronys: Indeed, Resolved in a530b96cd

src/core/PWSAuxParse.cpp
```diff
@@ -462,6 +497,12 @@ StringX PWSAuxParse::GetAutoTypeString(const StringX &sx_in_autotype,
         case TCHAR('2'):
           sxtmp += duplicateCharInString(sx_totpauthcode, L'\\');
           break;
+        case TCHAR('v'):
+          if (!ExpandCustomFieldValue(sx_autotype, n, pcustomfields, sxtmp)) {
```
dskrvk: Can we change this to
```
if (ExpandCustomFieldValue(sx_autotype, n, pcustomfields, sxtmp)) {
    break;
}
```
and let the default case handle putting back the `\v`?

src/core/PWSAuxParse.cpp
```diff
@@ -317,6 +320,37 @@ static bool GetSpecialCommand(const StringX &sx_autotype, size_t &n, WORD &wVK,
...
+    return false;
+
+  const StringX sxName = sx_autotype.substr(n + 2, iEndBracket - n - 2);
+  if (pcustomfields != nullptr) {
+    const auto it = std::find_if(pcustomfields->begin(), pcustomfields->end(),
+                                 [&sxName](const CustomField &cf) {
+                                   return cf.GetName() == sxName;
+                                 });
+    if (it != pcustomfields->end()) {
+      for (TCHAR c : it->GetValue()) {
```
dskrvk: Can we use `duplicateCharInString` here?

src/core/PWSAuxParse.cpp
```diff
@@ -565,7 +606,6 @@ StringX PWSAuxParse::GetAutoTypeString(const StringX &sx_in_autotype,
           sxtmp += L'\x1B';
           break;        // Also copy explicit control characters to output string unchanged.
         case TCHAR('a'): // bell (can't hear it during testing!)
-        case TCHAR('v'): // vertical tab
```
dskrvk, line 568: Would it still be valuable to support vtab, which is still distinguishable from a custom field spec (which must always contain the `{name}` part)?

src/test/AuxParseTest.cpp
```diff
@@ -94,3 +100,40 @@ TEST(AuxParseTest, testGetExpandedString)
...
+
+  const StringX expanded = PWSAuxParse::GetAutoTypeString(
+      L"prefix \\v{PIN} suffix \\v{Missing}",
+      L"", L"", L"", L"", L"", L"", L"", L"", L"",
+      &customFields, vactionverboffsets);
+
+  EXPECT_EQ(L"prefix 12\\\\34 suffix ", expanded);
+}
+
+TEST(AuxParseTest, testGetAutoTypeStringUsesItemCustomFields)
```
dskrvk, line 123: Could you also add tests for `\\v` and `\\v{}`?

src/core/PWSAuxParse.cpp
ronys: This will also require moving the case to above the default case.
OK

src/core/PWSAuxParse.cpp
ronys: nice catch

src/core/PWSAuxParse.cpp
ronys: I'd be very surprised if \v is used anywhere in 2026...

src/test/AuxParseTest.cpp
ronys: done

dskrvk (approved):

PR #1746, opened 2026-04-07
[wx] First hack at custom field support

ronys:
Getting started on release 1.24.
Note that the UI is significantly different from Windows: Here the user can toggle between the Notes and Custom Fields.
Otherwise the dialog would have been way too tall.
This can be improved:
1. If there's a custom field but no note, default to the custom fields.
2. Define some kind of shortcut to toggle between the 2.
3. Think of something other than a listbox to show the 2 options.


src/ui/wxWidgets/AddEditPropSheetDlg.cpp
```diff
@@ -3542,13 +3691,238 @@ void AddEditPropSheetDlg::OnUpdateUI(wxUpdateUIEvent& event)
...
+
+void AddEditPropSheetDlg::CopySelectedCustomFieldToClipboard()
+{
+  const int selectedIndex = GetSelectedCustomFieldIndex();
+  if (selectedIndex == wxNOT_FOUND || selectedIndex >= static_cast<int>(m_CustomFields.size())) {
+    return;
+  }
+
+  const CustomField &field = m_CustomFields[static_cast<size_t>(selectedIndex)];
+  auto *pwSafe = const_cast<PasswordSafeFrame*>(GetPwSafe());
```
rafaelx, line 3898: For this purpose, there is also 'PasswordSafeFrame* GetPasswordSafeFrame()' in PWSafeApp.h, which avoids a type conversion.

src/ui/wxWidgets/AddEditPropSheetDlg.cpp
```diff
@@ -3542,13 +3691,238 @@ void AddEditPropSheetDlg::OnUpdateUI(wxUpdateUIEvent& event)
...
+  if (m_BasicCustomFieldsListCtrl == nullptr) {
+    return wxNOT_FOUND;
+  }
+
+  return static_cast<int>(m_BasicCustomFieldsListCtrl->GetNextItem(-1, wxLIST_NEXT_ALL, wxLIST_STATE_SELECTED));
+}
+
+bool AddEditPropSheetDlg::EditCustomField(CustomField *field)
+{
+  CustomFieldEditDialog dlg(this, m_CustomFields, field);
```
rafaelx, line 3808: For modal dialogs, there is a 'DestroyWrapper' in wxUtilities.h in case the database is locked due to a timeout.

rafaelx:
Some feedback on the "first hack" regarding support for custom fields.

#### AddEditPropSheetDlg ####

1. I find it unusual to switch between different views in the user interface using a combo box. How about a tabbed view with one tab for "Notes" and another for "Custom Fields"?
2. The custom fields column could support sorting by name when the table header is clicked.
3. Currently, when a custom field is selected, its value is copied to the clipboard. This might surprise some users and be undesirable. How about copying the value to the clipboard by double-clicking, as with the password in the tree view?

#### CustomFieldEditDialog ####

1. In some dialogs, including Add/Edit, layout changes have been made so that the label is now positioned above a text input field. For consistency, the new dialog could also adopt this style.
2. The multi-line text input field for the value could also be vertically expandable to always fill the entire dialog. This is useful if the custom field contains a lot of text.

rafaelx:
Some further ideas:

#### AddEditPropSheetDlg ####

If it is preferred to switch between the notes view and the custom field view using a combobox/choicebox, then I would suggest extending the combobox/choicebox across the entire width of the dialog, visually emphasizing that this affects the view below. This would be similar to a title introducing a new paragraph/textual content and similar to the tabs in AddEditPropSheetDlg.

To avoid a long chain of modal dialog boxes, editing user-defined fields could be done directly in the table, which already provides all the necessary input fields (as cells). A double-click could switch to edit mode. A context menu could be used to delete an entry and copy the value to the clipboard. To mark sensitive entries, the context menu could include a checkbox, similar to the checkbox to protect password elements via the main view (main menu / context menu).

####  PasswordSafeFrame ####

To simplify copying custom fields for the user, an additional menu item (Copy custom field to clipboard) could be added to the main menu and context menu of the main view. This menu item could lead to a submenu with all the names of the custom fields assigned to a password element. This would allow the user to specifically select a value from one of the custom fields to copy to the clipboard. This would be consistent with the existing "copy to clipboard" functions, which can be executed from the main view for other data of a password element.
I would therefore not implement a copy mechanism (via single-click/double-click) in CustomFieldEditDialog. If the table in CustomFieldEditDialog is editable, then copying the value of a custom field is also possible, and additionally via a context menu.

src/ui/wxWidgets/AddEditPropSheetDlg.cpp
ronys: Addressed in 0e64085d8

src/ui/wxWidgets/AddEditPropSheetDlg.cpp
ronys: Addressed in 0e64085d8

rafaelx:
For anyone who wants to see what the changes look like.

grafik


grafik


grafik

nobugshere:
First quick-look on a Mac: 
macOS M1Max Sequoia 15.7.5, Xcode 16.4, wx 3.2.8 and 3.2.10 (local build)

So far, I don't see any Mac specific weirdness.
I tend to agree with @rafaelx comments.  In particular, given the potentially sensitive nature of the data, I don't think it should be copied to the clipboard just by selecting a row in the table.  There should be a specific action, i.e. a context menu item with a Ctrl-C shortcut.

dskrvk:
Instead of having a "Toggle Sensitive" button, could we add another column to the fields list showing a (writable) checkbox that indicates the sensitive state (and allows changing it)?

ronys:
- Copy is now only done via explicit Ctrl/Cmd-C. This is now also consistent with the password history table.
- Per @dskrvk 's suggestion, I've replaced the "Toggle Sensitive" button with a "Sensitive" checkbox column.
- For now, I'm keeping the Edit dialog for custom values, mainly because I think it's clearer / more obvious to users that this is how editing is supported.
- Adding the custom fields to the menus is a nice idea, but perhaps not for the first release of this feature in wx. Let's see what feedback we get from this field first.

nobugshere:
I just noticed a minor regression:
macOS M1Max Sequoia 15.7.5, Xcode 16.4, wx 3.2.8 and 3.2.10

In the Basic tab of the Add/Edit dialog, most of the controls only extend about 3/4 of the width.  None of the other tabs are affected.

Screenshot 2026-05-03 at 4 22 32 PM


This is the normal look using a build from the current master:
Screenshot 2026-05-03 at 4 30 53 PM

ronys:
I think that this is Good Enough (tm) to merge into master, as a "first hack", as I'd like to make a wx release over this weekend.

PR #1750, opened 2026-04-17
[wx] Enable virtual keyboard on Wayland

froggie2523:
Issue: In many modern distros having Wayland as the default display server, Password Safe does not show the virtual keyboard icon.

This PR is to enable virtual keyboard (xvkbd) on Wayland in Linux/BSD when the X11/Xorg backend through the XWayland compatibility layer is enabled.

Tested on Debian(GNOME)/Mint(Cinnamon)/Kubuntu/FreeBSD(KDE) with Wayland.

pwsafe_1 23-wxWidgets-xvkbd-Wayland-01


ronys:
Thanks!

PR #1752, opened 2026-04-18
[wx] Show vertical scrollbar when AddEdit Dialog resized

### LINKED ISSUE #1751: Restore scrollbars in password entry dialog
From Brett via email:

> The latest Linux version removes the scroll bars from the password entry editing sub-window. This is terrible for those with old machines with low-resolution screens because now there's no way to see the notes for a password entry.
Yes, they can (luckily) be copied by right-clicking on the password entry but it is still not the greatest experience.
Please re-enable scroll bars in the password entry window.
Version: 1.23 (heads/master-0-gf5084301c) - running on Ubuntu.

ronys:
Resolves #1751

### DIFF
--- a/docs/ReleaseNotesWX.md
+++ b/docs/ReleaseNotesWX.md
@@ -12,6 +12,7 @@ PasswordSafe 1.24.0 Release  ???
 
 Bugs fixed in 1.24.0
 --------------------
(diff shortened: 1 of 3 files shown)


froggie2523:
Nice feature, especially useful on Linux, where Wayland provides fractional display scaling.
Tested successfully on macOS, Debian (Wayland/X11), and Kubuntu.

PR #1753, opened 2026-04-19
Preserve passkey fields in CItemData plaintext serialization

huven:
## Summary

Include passkey entry fields in `CItemData` plaintext serialization so they
survive serialize/deserialize round-trips.

This fixes loss of passkey data in flows that use the plaintext entry
serialization path, including Windows cross-safe drag and drop.

## Changes

- serialize all six passkey fields in `CItemData::SerializePlainText()`
- keep the existing field-type handling in `DeSerializePlainText()`
- extend `ItemDataTest.PlainTextSerialization` to cover passkey fields

## Notes

- entries without passkey fields are unaffected
- binary, text, and 4-byte passkey field types are preserved as-is

### DIFF
--- a/src/core/ItemData.cpp
+++ b/src/core/ItemData.cpp
@@ -2559,6 +2559,35 @@ void CItemData::SerializePlainText(vector<char> &v,
     }
   }
 
+  // Passkey fields
+  {
+    const FieldType passkeyfts[] = {
+      PASSKEY_CRED_ID,
+      PASSKEY_RP_ID,
+      PASSKEY_USER_HANDLE,
+      PASSKEY_ALGO_ID,
+      PASSKEY_PRIVATE_KEY,
+      PASSKEY_SIGN_COUNT
+    };
+    std::vector<unsigned char> field;
+
+    for (auto ft : passkeyfts) {
+      if (ft == PASSKEY_RP_ID) {
+        if (IsFieldSet(ft))
+          push(v, ft, GetField(ft));
+        continue;
+      }
+
+      field.clear();
+      GetField(ft, field);
+      if (!field.empty()) {
+        push(v, ft, static_cast<uint32>(field.size()),
+             reinterpret_cast<char *>(field.data()));
+        trashMemory(field.data(), field.size());
+      }
+    }
+  }
+
   // Unknown fields
   for (auto vi_IterURFE = m_URFL.begin();
        vi_IterURFE != m_URFL.end();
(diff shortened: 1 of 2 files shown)


src/test/ItemDataTest.cpp
```diff
@@ -142,6 +142,13 @@ TEST_F(ItemDataTest, Getters_n_Setters)
 
 TEST_F(ItemDataTest, PlainTextSerialization)
 {
+  fullItem.SetPasskeyCredentialID(VectorX<unsigned char>(64, 1));
+  fullItem.SetPasskeyRelyingPartyID(_T("example.com"));
+  fullItem.SetPasskeyUserHandle(VectorX<unsigned char>(32, 2));
+  fullItem.SetPasskeyAlgorithmID(-7);
+  fullItem.SetPasskeyPrivateKey(VectorX<unsigned char>(512, 3));
+  fullItem.SetPasskeySignCount(0);
```
dskrvk: Why not set these inside SetUp, like other fields?

src/core/ItemData.cpp
```diff
@@ -2559,6 +2559,53 @@ void CItemData::SerializePlainText(vector<char> &v,
     }
   }
 
+  // Passkey fields
+  {
+    std::vector<unsigned char> field;
+
+    GetField(PASSKEY_CRED_ID, field);
```
dskrvk: This and the other 4 fields below are all handled the same way. Could we e.g. create a local vector var containing field names and process them in a loop (or create a helper method)?

src/test/ItemDataTest.cpp
huven: Good point. I’ve moved the passkey field setup into `SetUp()` so it follows the existing fixture pattern, and added the corresponding getter checks in `Getters_n_Setters`.

src/core/ItemData.cpp
huven: I changed that block to use a single ordered local `FieldType` list and a loop, keeping `PASSKEY_RP_ID` on the text path and the other passkey fields on the raw-byte path.

dskrvk (approved):

ronys:
After merging I compiled locally on Windows and got the following compiler warnings:
...\src\core\ItemData.cpp(2577,19): warning C4244: 'argument': conversion from 'CItem::FieldType' to 'char', possible loss of data
            push(v, ft, GetField(ft));
                    ^
...\src\core\ItemData.cpp(2584,17): warning C4244: 'argument': conversion from 'CItem::FieldType' to 'char', possible loss of data
          push(v, ft, static_cast<uint32>(field.size()),
                  ^

Also, passkeyfts can be a constexpr.

Making these changes in a separate commit.

dskrvk:
So compilation succeeds but produces warnings? Maybe we should add the `-Werror` compiler flag to treat warnings as errors?

huven:
There are several warnings on Linux and macOS, so these will need to be addressed first if you want to enable that cross-platform. Which I think is a good idea, btw.

dskrvk:
Right, I'll try to tackle some of those.

PR #1754, opened 2026-04-21
[wx] Enable AutoType on Linux/BSD under Wayland

froggie2523:
Issue: Many modern Linux distros are moving to Wayland as the default display server (protocol that is implemented by the desktop environment and its compositor).
For example, GNOME 50 (released March 2026) has officially dropped native X11 session support (Xwayland remains supported, allowing legacy X11 applications to run within a Wayland session). Some Wayland-only distros like Fedora 44 and Ubuntu 26.04 with GNOME 50 are scheduled to be released imminently.
Currently, Password Safe disables the AutoType feature when Wayland is detected. Users may encounter this change in several scenarios:
1. Some distros (still) offer a Wayland or Xorg session to choose from on the login screen. A user logs out of an Xorg session and logs into a Wayland session.
2. The OS gets upgraded with Wayland as the default display server, and an Xorg session may no longer be available.
3. A user migrates to a newer/different distro where Wayland is the default or the only available display server.

Fortunately, the AutoType feature works just fine on Wayland, with minimal one-time tweaking the one or two apps people use it typically with:
-inside web browsers
-Remote Desktop/VNC/SSH/VPN clients connecting to remote systems
-SMB/FTP connections within file managers
Note: Password Safe itself does not need to be started with Xwayland enabled (by setting the GDK_BACKEND=x11 environment variable).

This PR enables AutoType controls in Linux/BSD on Wayland and updates the tooltip in the Add/Edit dialog as well as in Options > Misc. It's up to the user to look for details in Help and search for solution for a specific app on the internet.

Tested on Debian (GNOME)/Mint (Cinnamon)/Fedora (KDE)/Kubuntu/FreeBSD (KDE) with Wayland, with the following apps:

Native X11 apps or apps using X11 by default:
Tor browser (set MOZ_ENABLE_WAYLAND=1 to use Wayland)
Double Commander (SMB/FTP connections)
Apps that can run in Xwayland mode:
Chrome: google-chrome --ozone-platform=x11
Firefox: env MOZ_ENABLE_WAYLAND=0 firefox (--nosocket=wayland --socket=x11 for flatpak version/env DISABLE_WAYLAND=1 for snap version)
Apps requiring Xwayland and the "Use alternate AutoType method" option enabled in Options > System:
another instance of Password Safe :-)
Remmina (RDP/VNC/SSH client)
Filezilla

Note: wxUtilities::DisableIfUnsupported should eventually be renamed to something like NotifyIfConditionallySupported.

As usual, some fancy screenshots:
pwsafe_1 23-wxWidgets-feat-AutoType-Wayland-01
pwsafe_1 23-wxWidgets-feat-AutoType-Wayland-02
pwsafe_1 23-wxWidgets-feat-AutoType-Wayland-03
pwsafe_1 23-wxWidgets-feat-AutoType-Wayland-04

### DIFF
--- a/src/ui/wxWidgets/wxUtilities.cpp
+++ b/src/ui/wxWidgets/wxUtilities.cpp
@@ -376,8 +376,8 @@ bool wxUtilities::IsVirtualKeyboardSupported()
 void wxUtilities::DisableIfUnsupported(enum Feature feature, wxWindow* window)
 {
   if (feature == Autotype && WhatWindowSystem() == Wayland) {
-    window->Disable();
-    window->SetToolTip(_("Not supported by Wayland"));
+//    window->Disable();
+    window->SetToolTip(_("Running on Wayland - make sure the target application is using Xwayland. For some applications, you may need to enable the 'Use alternate AutoType method' option in Options > System."));
   }
 }
 
--- a/src/ui/wxWidgets/wxUtilities.cpp
+++ b/src/ui/wxWidgets/wxUtilities.cpp
@@ -373,16 +373,11 @@ bool wxUtilities::IsVirtualKeyboardSupported()
 #endif
 }
 
-void wxUtilities::DisableIfUnsupported(enum Feature feature, wxWindow* window)
-{
-  if (feature == Autotype && WhatWindowSystem() == Wayland) {
-//    window->Disable();
-    window->SetToolTip(_("Running on Wayland - make sure the target application is using Xwayland. For some applications, you may need to enable the 'Use alternate AutoType method' option in Options > System."));
-  }
-}
-
 void wxUtilities::NotifyIfUnsupported(enum Feature feature, wxWindow* window)
 {
+  if (feature == Autotype && WhatWindowSystem() == Wayland) {
+    window->SetToolTip(_("Running on Wayland - make sure the target application is using Xwayland. For some applications, you may need to enable the 'Use alternate AutoType method' option in Options > System."));
+  }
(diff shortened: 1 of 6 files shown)


src/ui/wxWidgets/wxUtilities.cpp
```diff
@@ -376,8 +376,8 @@ bool wxUtilities::IsVirtualKeyboardSupported()
 void wxUtilities::DisableIfUnsupported(enum Feature feature, wxWindow* window)
 {
   if (feature == Autotype && WhatWindowSystem() == Wayland) {
-    window->Disable();
-    window->SetToolTip(_("Not supported by Wayland"));
+//    window->Disable();
```
dskrvk: Remove?

src/ui/wxWidgets/wxUtilities.cpp
```diff
@@ -376,8 +376,8 @@ bool wxUtilities::IsVirtualKeyboardSupported()
 void wxUtilities::DisableIfUnsupported(enum Feature feature, wxWindow* window)
```
dskrvk: This name is no longer accurate. Also, can we simply unconditionally set the tooltip, regardless of the windowing system? In other words, remove this method completely.

froggie2523:
Thanks for your feedback, I merged wxUtilities::DisableIfUnsupported and wxUtilities::NotifyIfUnsupported into 1 method.

dskrvk (approved):

PR #1758, opened 2026-04-28
[wx] Enable display scaling for virtual keyboard on Linux

froggie2523:
This PR is to enable display scaling for xvkbd on high-DPI displays in Linux/BSD, as it does not natively support DPI-aware scaling.
While xvkbd allows window resizing, the key labels do not automatically scale, making the font appear too small.
Tested on Linux (GNOME/Cinnamon/KDE) and BSD.
Also updated English Help - removed Windows specific text on virtual keyboard font.

pwsafe_1 23-wxWidgets-feat-xvkbd-scaling-01
<br />
pwsafe_1 23-wxWidgets-feat-xvkbd-scaling-02

### DIFF
--- a/help/default_wx/html/view_menu.html
+++ b/help/default_wx/html/view_menu.html
@@ -174,12 +174,8 @@
     <img alt="Change Fonts" src="images/change_font_menu.jpg">
 
     <p>Select a different font for displaying entries in the List or Tree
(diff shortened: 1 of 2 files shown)


ronys:
Nice touch - thanks!

PR #1759, opened 2026-04-29
[wx] Fix missing Run Command menu item and add some tooltips

froggie2523:
This PR is to add missing Run Command menu item and add tooltips for the Run Command and MFA fields in Add/Edit dialog.

Fix: Run Command menu item missing in Edit menu.
New: Added a tooltip for "Authentication code" field on the Basic tab in Add/Edit dialog (aligned with the Windows version).
New: Added tooltips for "Authentication Secret" (aligned with the Windows version) and "Run Command" fields on the Additional tab in Add/Edit dialog.
New: Display the default AutoType string on Additional tab in Add/Edit dialog, as configured in Options > Misc (consistent with the Windows version).

Notes:
Translations (POS) for "Authentication code" and "Authentication Secret" tooltips were taken from Windows POS files.
The tooltip and translations (POS) for "Run Command" were taken from the Help file (entering_pwd_add.html).

pwsafe_1 23-wxWidgets-feat-Tooltips-01
<br/>
pwsafe_1 23-wxWidgets-feat-Tooltips-02

### DIFF
--- a/help/default_wx/html/run_command.html
+++ b/help/default_wx/html/run_command.html
@@ -241,6 +241,11 @@ open ssh://${user}@${url}
 open smb://${user}@${url}/sharename
 </pre>
 </li>
+<li>On macOS, this will open a new instance of <b>Password Safe</b> with the specified database (use "\p\n" AutoType string to unlock it):
+<pre>
+open -n -a pwsafe ${dbdir}/database.psafe3
+</pre>
+</li>
 <li>On Linux, this will open a new instance of <b>Password Safe</b> with the specified database (Thanks to Dave for this one):
 <pre> ${autotype}(\p\n)"/usr/bin/pwsafe" \UNC\path\to\database.psafe3</pre></li>
 <ul>
--- a/help/default_wx/html/run_command.html
+++ b/help/default_wx/html/run_command.html
@@ -11,7 +11,7 @@
 <h1>Run Command</h1>
 
 <p>An entry's 'Run Command' field can contain a command that will
-be run when the Run Command action is selected. Anything that you can do on Linux in the "Run" dialog (Alt+F2) can be used here.</p>
+be run when the Run Command action is selected. On Linux, anything that you can do in the "Run" dialog (Alt+F2) can be used here.</p>
 
 <p>In addition, <b>Password Safe</b> is able to pass arguments to the command
 such as the entry's username, password, etc., as described below. This
@@ -226,12 +226,22 @@ variable.</p>
 
 <h3>Examples</h3>
 <ul>
-<li>The following Run Command will cause to connect to remote host via RDP giving the entry's username, password and URL:
+<li>On Linux, the following Run Command will cause to connect to remote host via RDP giving the entry's username, password and URL:
 <pre>
 remmina -c rdp://${user}:${password}@${url}
 </pre>
 </li>
-<li>This will open a new instance of <b>Password Safe</b> with the specified database (Thanks to Dave for this one):
+<li>On macOS, this launches Terminal and opens SSH connection to remote host using the entry's username and hostname (copy and paste the password to authenticate):
+<pre>
+open ssh://${user}@${url}
+</pre>
+</li>
+<li>On macOS, this connects to a remote Shared folder via SMB using the entry's username and URL (copy and paste the password to authenticate):
+<pre>
+open smb://${user}@${url}/sharename
+</pre>
+</li>
+<li>On Linux, this will open a new instance of <b>Password Safe</b> with the specified database (Thanks to Dave for this one):
 <pre> ${autotype}(\p\n)"/usr/bin/pwsafe" \UNC\path\to\database.psafe3</pre></li>
 <ul>
 <li>Make sure Manage&nbsp;&rarr;&nbsp;Options&nbsp;&rarr;&nbsp;System "Allow multiple instances" is checked for this to work</li>
--- a/src/ui/wxWidgets/AddEditPropSheetDlg.cpp
+++ b/src/ui/wxWidgets/AddEditPropSheetDlg.cpp
@@ -504,8 +504,7 @@ wxScrolledWindow* AddEditPropSheetDlg::CreateAdditionalPanel()
   StringX sx_dats = PWSprefs::GetInstance()->GetPref(PWSprefs::DefaultAutotypeString);
   if (sx_dats.empty())
     sx_dats = DEFAULT_AUTOTYPE;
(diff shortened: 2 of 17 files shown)


nobugshere:
PWSRun::runcmd() is not implemented on macOS :-(   
I don't know why it wouldn't be the same as the unix version.

There's also a OnCopyRunCmd (copy the run command to the clipboard) event handler that doesn't appear to be hooked-up.

froggie2523:
TBH, I haven't checked whether runcmd() is implemented on macOS. Should I disable the Run Command on macOS and change the tooltip to e.g. "Not Implemented"?
The "Not Implemented" phrase is part of some texts and translated so I can extract that and add to POS files...
The "Copy Run command to Clipboard" works for me on macOS; it's disabled if the "Run Command" field is empty and it will get empty once disabled :-)

ronys:
The unix implementation of PWSRun::runcmd() calls ::wxExecute(runcommand.c_str());

No reason why this wouldn't work on mac.

I suggest copying src/os/unix/run.cpp to src/os/mac.

src/ui/wxWidgets/AddEditPropSheetDlg.cpp
```diff
@@ -501,7 +501,11 @@ wxScrolledWindow* AddEditPropSheetDlg::CreateAdditionalPanel()
   auto *vBoxSizer = new wxBoxSizer(wxVERTICAL);
   mainSizer->Add(vBoxSizer, 1, wxEXPAND|wxALL, 10);
 
-  auto *itemStaticText41 = new wxStaticText(panel, wxID_STATIC, _("Autotype"), wxDefaultPosition, wxDefaultSize, 0);
+  StringX sx_dats = PWSprefs::GetInstance()->GetPref(PWSprefs::DefaultAutotypeString);
+  if (sx_dats.empty())
+    sx_dats = DEFAULT_AUTOTYPE;
+  wxString AutoTypeLabel = _("Autotype") + _T(' ') + _T('(') + _("default") + _T(':') + _T(' ') + towxstring(sx_dats) + _T(')');
```
dskrvk: Why not use a single template string with `_()` so that translators can see the whole message? Not all languages have the same word order as English, and it may not be appropriate to substitute individual words (e.g. "default" may have different translations in different contexts). It may seem like we're saving work by reusing parts of translated sentences, but it would just result in worse-quality localization.

nobugshere:
> The "Copy Run command to Clipboard" works for me on macOS;

Ah, I see it now on the context menu, I was looking for it on the main Edit menu.

nobugshere:
> The unix implementation of PWSRun::runcmd() calls ::wxExecute(runcommand.c_str());
> No reason why this wouldn't work on mac.
> I suggest copying src/os/unix/run.cpp to src/os/mac.

I just tried it and it does work.

Is it better to copy the file or just use the one under os/unix to reduce duplication? (a couple other files are already included that way in the Xcode project)

Either way, it requires a change to the Xcode project file to find the wx headers.  I can submit it as a separate Pwsafe PR, or as a PR to @froggie2523's branch.  Any preference?

ronys:
Hi @nobugshere 
- Reducing duplication is better. There are already some files that don't really need to be duplicated (e.g., sleep.cpp)
- In addition to the Xcode project, src/os/CMakeLists.txt should also reflect this.
- A separate PR is better, since this is a separate issue, with no conflicting changes.

froggie2523:
Now that the Run Command will work on macOS, I took the liberty of adding some examples to English Help.

help/default_wx/html/run_command.html
```diff
@@ -226,12 +226,22 @@ <h1>Run Command</h1>
...
+<pre>
+open ssh://${user}@${url}
+</pre>
+</li>
+<li>On macOS, this connects to a remote Shared folder via SMB using the entry's username and URL (copy and paste the password to authenticate):
+<pre>
+open smb://${user}@${url}/sharename
+</pre>
+</li>
+<li>On Linux, this will open a new instance of <b>Password Safe</b> with the specified database (Thanks to Dave for this one):
```
nobugshere, line 249: If "Allow multiple instances" is set in Options->System.

On macOS you can use:
open \path\to\database.psafe3
or just this, with or without a path:
open -a pwsafe

nobugshere (approved):

src/ui/wxWidgets/AddEditPropSheetDlg.cpp
ronys: I agree. This also just looks odd.

help/default_wx/html/run_command.html
```diff
@@ -241,6 +241,11 @@ <h3>Examples</h3>
 open smb://${user}@${url}/sharename
 </pre>
 </li>
+<li>On macOS, this will open a new instance of <b>Password Safe</b> with the specified database (use "\p\n" AutoType string to unlock it):
```
nobugshere, line 244: Does adding the "${autotype}(\p\n)" string, as in the Linux example below, work for you on macOS?  For me, it launches the app, but it doesn't autotype.  Using the "Perform Autotype" command, after the app is launched, does work.  (I tried it on Sequoia and Tahoe.)

If a separate "Perform Autotype" command is what you ment to use, maybe clarify the statement.  If it'a a bug, it can be a separate issue; no need to hold-up this PR.

dskrvk (approved):

help/default_wx/html/run_command.html
froggie2523: I came to the same conclusion after testing the examples again. I tested on Linux and suspected it might be related to Wayland or the “Use alternate AutoType method” option, but it appears this feature was never implemented on UNIX systems.
Anyway, I already have a fix ready, it needs some more testing to do. After PR #1760 was merged and your fix from PR #1761 is applied, it works as expected.
Note: On macOS, I sometimes need to remove and re-add pwsafe to the list in System Settings > Privacy & Security > Accessibility (not just re-enable it), as noted in “help/default_wx/html/autotype.html” in PR #1761. But in my case, this may be related to repeated recompilations or reinstallations during testing, or because I do it in a VM being suspended.

nobugshere: > ...I sometimes need to remove and re-add pwsafe to the list....

I've noticed this now too.  I think you might be right about recompile/reinstall being the cause.  I don't know how it identifies multiple copies, but their ad hoc signatures would be different.

I've also noticed that if I remove it from the list, the next time I try to use it I don't always get the prompt asking for permission to add it back, it just seems to be denied until I add it manually.  That too might be because of all the machinations.

Since you have another PR in the works, would you mind updating the note in autotype.html?  Thanks :-)

PR #1761, opened 2026-05-02
[macOS] Minor improvement to AutoType

nobugshere:
This makes "Perform Auto Type", from the context menu, work a little better.  Also added a note about permissions to the help file.

Question: In the context menu entry "Browse URL + Autotype", the Autotype part doesn't seem to be implemented.  All I see in the code is: "#ifdef NOT_YET".  I don't know what the intention was,  but couldn't it just be a call to DoAutotype() after DoBrowse() in OnBrowseUrlAndAutotype() ?  There's obviously some history here I don't get.

### DIFF
--- a/help/default_wx/html/autotype.html
+++ b/help/default_wx/html/autotype.html
@@ -52,6 +52,9 @@ for your machine.</p>
 command specified in the entry's <a href="run_command.html">Run Cmd</a> 
 field. This is a very convenient combination.</p> 
 
+<p><b>Note:</b> <i>On macOS, in order for the system to allow AutoType, <b>pwsafe</b> must be present and enabled in
+<b>System Settings->Privacy & Security->Accessibility</b>.</i></p>
+
 <p><b>Note:</b> <i>On Linux, AutoType functionality depends on the Display Server used, such as Wayland or X.Org/X11.
 Under Wayland, you may need to run the target application using the X11 backend via the Xwayland compatibility layer.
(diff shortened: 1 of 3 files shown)


ronys:
#ifdef NOT_YET in the code refer to the initial implementation of parts of the wx code that were copied over from the Windows version, and either wasn't implemented at the time, or I wasn't sure if it needed to be implemented under wx.

nobugshere:
> ...either wasn't implemented at the time, or I wasn't sure if it needed to be implemented under wx.

Okay.  Maybe I'll submit a separate issue to implement "Browse to URL + Autotype".

help/default_wx/html/autotype.html
```diff
@@ -52,6 +52,9 @@ <h2>AutoType</h2>
 command specified in the entry's <a href="run_command.html">Run Cmd</a> 
 field. This is a very convenient combination.</p> 
 
+<p><b>Note:</b> <i>On macOS, in order for the system to allow AutoType, <b>pwsafe</b> must be present and enabled in
+<b>System Settings->Privacy & Security->Accessibility</b>.</i></p>
```
ronys:
```suggestion
<b>System Settings->Privacy &amp; Security->Accessibility</b>.</i></p>
```

ronys (commented):
In html, '& a m p;' (spaced out to display correctly)is the correct way to display an ampersand sign.

help/default_wx/html/autotype.html
nobugshere: Fixed it.  Thanks!

PR #1762, opened 2026-05-05
vector.erase invalidates the iterator - found by cppcheck

carlbyington2:
I ran cppcheck which found this.

### DIFF
--- a/src/core/PWScore.cpp
+++ b/src/core/PWScore.cpp
@@ -268,7 +268,7 @@ bool PWScore::ConfirmDelete(const CItemData *pci, StringX sxGroup)
     if (!sxGroup.empty()) {
       UUIDVectorIter iter;
       size_t length = sxGroup.length(); // plus 1 to include trailling 0 byte
-      for (iter = dependentslist.begin(); iter != dependentslist.end(); ++iter) {
+      for (iter = dependentslist.begin(); iter != dependentslist.end(); ) {
         ItemListIter objiter = Find(*iter);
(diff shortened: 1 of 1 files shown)


ronys:
Thanks!

If that's all cppcheck found, we're in good shape :-)

Did you test this in the app to verify that the behavior is unchanged?

carlbyington2:
-----BEGIN PGP SIGNED MESSAGE-----
Hash: SHA512

On Mon, 2026-05-04 at 23:05 -0700, ronys wrote:

> If that's all cppcheck found, we're in good shape :-)

I think there are more, but this is the first one that I knew I could
fix. cppcheck has a lot of false positives. I will also run the clang
analyzer when I am done looking at the cppcheck output.

> Did you test this in the app to verify that the behavior is
> unchanged?

No, I don't have any aliases in my safe. It clearly needs testing.

BTW, thanks for maintaining this!


-----BEGIN PGP SIGNATURE-----

iHMEAREKADMWIQSuFMepaSkjWnTxQ5QvqPuaKVMWwQUCafoMKRUcY2FybEBmaXZl
LXRlbi1zZy5jb20ACgkQL6j7milTFsGdTQCfYPagJz37Bu1w+NESpUaWuQPXAFcA
nRnp2NeRQ0ZNCWtoJvx8OfFXhMxQ
=oldI
-----END PGP SIGNATURE-----

ronys:
Tested and works fine.
I've cleaned up a couple of comments while there...

PR #1764, opened 2026-05-06
ItemAtt.cpp leaked memory if GetContent returned false.

carlbyington2:
PWSfile.cpp leaked memory fread(ivthing failed.
GridCtrl.cpp did not protect GUI_REFRESH_ENTRY from a null pointer dereference.

clang-analyzer found these. I am not sure how to force test cases for these.

### DIFF
--- a/src/core/PWSfile.cpp
+++ b/src/core/PWSfile.cpp
@@ -430,6 +430,7 @@ bool PWSfile::Encrypt(const stringT &fn, const StringX &passwd, stringT &errmess
 
 bool PWSfile::Decrypt(const stringT &fn, const StringX &passwd, stringT &errmess)
 {
+  Fish *fish = nullptr;
   ulong64 file_len;
   bool status = true;
   unsigned char salt[SaltLength];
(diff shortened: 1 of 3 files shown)


src/core/PWSfile.cpp
```diff
@@ -502,6 +502,7 @@ bool PWSfile::Decrypt(const stringT &fn, const StringX &passwd, stringT &errmess
     ivthing = new unsigned char[BS];
 
     if (fread(ivthing, 1, BS, in) != BS) {
+      delete fish;
```
nobugshere: Instead of deleting fish in so many places, would it be better to initialize it to nullptr and delete it once after exit? (As it already does for ivthing.)

ronys: Or, better yet, rewrite this to use std::unique_ptr<>

nobugshere: Yes, sometimes the new ways are better :-)

dskrvk (approved):

PR #1765, opened 2026-05-06
[wx] Enable AutoType in Run Command

froggie2523:
Issue: Despite specifying an AutoType string in the Run Command field in the Add/Edit dialog (as described in Help), no action is triggered.
Users must manually invoke the AutoType command, which may use a different AutoType string defined for the entry or fall back to the global string in Settings > Misc.
Change: Enable the AutoType feature for the Run Command action on Linux and macOS.
Change: Update notes in help/default_wx/html/autotype.html.
Change: Update examples in help/default_wx/html/run_command.html.


nobugshere:
Works for me now, too.  Tested on Sequoia and Tahoe.

PR #1767, opened 2026-05-07
Add override for GetName()

nobugshere:
A new warning appeared because this is a virtual function defined in (WX) window.h.  I don't think it hurts anything, but a different name might be better, since it's conceptually a different function.


ronys:
I think it's better to rename GetName() to GetFieldName(), and rename GetValue() to GetFieldValue(), for consistency.

nobugshere:
Done.  I think it's right - one call to each function.  Other calls to GetName are for a different class type.

PR #1770, opened 2026-05-08
Improve fedora packaging with a proper .spec file. 

carlbyington2:
This probably breaks the old "cpack -G RPM .." method. The new procedure is "cd ../install/rpm; make". I don't know cmake/cpack, so don't know where to disable that old method of building rpms.


install/rpm/Makefile
```diff
@@ -1,58 +1,22 @@
...
+	sed 's/@VERSION@/$(VERSION)/g' <pwsafe.spec.tmpl >passwordsafe.spec
+	rpmbuild -ba -v passwordsafe.spec
+	rm passwordsafe.spec
+	find $(RPMBUILDDIR) -name passwordsafe*rpm | grep $(VERSION)
 
-clean:
-	@$(RM)  $(RPMBUILDDIR)
+tarball:
+	mkdir  $(RPMBUILDDIR)/SOURCES || /bin/true
+	[ ! -f ../../version.wx ] && echo "current directory should end with install/rpm" && /bin/false
```
ronys: I don't understand the purpose of this line or the next. In any case, it causes make to fail on my Fedora VM. Commenting this and the next out and make runs fine.
Perhaps a better check would be for MAJOR and MINOR to be non-empty.

README.LINUX.DEVELOPERS.md
```diff
@@ -40,9 +40,10 @@ If it encounters errors, check your locale settings. It is known to work with:
 ```
     $ cpack -G DEB ..
 ```
-- To build an RPM package:
+- To build RPM packages:
 ```
-    $ cpack -G RPM ..
+    $ cd .. # to the top level
+    $ cd install/rpm; make
```
ronys: cpack still works just fine, and I'd rather have it remain the default build option.
I've no problem with make being described in addition.

README.LINUX.DEVELOPERS.md
```diff
@@ -53,7 +54,7 @@ If it encounters errors, check your locale settings. It is known to work with:
 
 - To install the rpm package you've just built:
 
-    `$ dnf install passwordsafe-fedora-\<version\>.\<arch\>.rpm`
+    `$ dnf install passwordsafe-\<version\>.\<arch\>.rpm`
```
ronys: The distro name is part of the naming convention for all packages.
Please either preserve it in the make process or explain the different package names here.

ronys (changes requested):

install/rpm/Makefile
carlbyington2: bug in the test, should be "|| /bin/true". It is protecting the subsequent "rm -rf build" to avoid clobbering a build directory if make was invoked from some other directory.

README.LINUX.DEVELOPERS.md
carlbyington2: reverted to use cpack. I was under the impression that cpack -G RPM was using install/rpm/* files, but it does not. Those files don't seem to be referenced anywhere. The main problem (from a fedora packaging perspective) is that cpack -G RPM does not build an rpm that passes the fedora packaging guidelines.

README.LINUX.DEVELOPERS.md
carlbyington2: reverted that to properly document the rpm produced by cpack.

install/rpm/Makefile
```diff
@@ -1,58 +1,23 @@
...
-HELPDIR := ../../help
-MODIR   := ../../src/ui/wxWidgets/I18N/mos
-RELEASEDIR := ../../Releases
-
-# I'd prefer the rpm build to be done under this directory,
-# but rpmbuild's default is ~/rpmbuild, and life's too short to find out
-# where to change this.
+MAJOR := $(shell grep MAJOR.= ../../version.wx | awk '{print $$3}')
+MINOR := $(shell grep MINOR.= ../../version.wx | awk '{print $$3}')
+PWD   := $(shell pwd)
```
dskrvk: I see the previous revision defined aliases for shell commands on lines 7-15. I assume it was to prevent command hijacking and/or avoid issues with PATH. Should we keep that approach?

install/rpm/Makefile
```diff
@@ -1,58 +1,23 @@
...
+build: tarball
+	sed 's/@VERSION@/$(VERSION)/g' <pwsafe.spec.tmpl >passwordsafe.spec
+	rpmbuild -ba -v passwordsafe.spec
+	rm passwordsafe.spec
+	find $(RPMBUILDDIR) -name passwordsafe*rpm | grep $(VERSION)
 
-clean:
-	@$(RM)  $(RPMBUILDDIR)
+tarball:
+	mkdir  $(RPMBUILDDIR)/SOURCES || /bin/true
```
dskrvk: Nit:

```suggestion
	mkdir  -p $(RPMBUILDDIR)/SOURCES
```
(plus see the previous comment about explicit paths to shell commands)

install/rpm/Makefile
```diff
@@ -1,58 +1,23 @@
...
+	[ ! -f ../../version.wx ] && echo "current directory $(PWD) should end with install/rpm" || /bin/true
```
dskrvk:
```suggestion
	[ ! -f ../../version.wx ] && echo "Couldn't find version.wx. Make sure current directory $(PWD) ends with install/rpm" || /bin/true
```
(maybe the directory exists, but the file isn't there - this gives the user a bit more info)

install/rpm/Makefile
```diff
@@ -1,58 +1,23 @@
...
+	[ ! -d ../../CMakeModules ] && echo "current directory $(PWD) should end with install/rpm" || /bin/true
```
dskrvk: Same as above

install/rpm/Makefile
```diff
@@ -1,58 +1,23 @@
...
+	(cd ../..; rm -rf build; tar cfz $(RPMBUILDDIR)/SOURCES/passwordsafe-$(VERSION).tar.gz *)
```
dskrvk: Any reason not to put each command on its own line (instead of using a subshell)?

install/rpm/pwsafe.spec.tmpl
```diff
@@ -1,92 +1,115 @@
...
+ninja
+#
+cd ../docs
+grep -E -v 'gitpod|badge.svg' ../README.md >README.md
+# gitpod image corrupted when converted to pdf by ooffice
+# missing image for badge
+ooffice --headless --convert-to pdf pwsafe-state-machine.rtf
+ooffice --headless --convert-to pdf ReleaseNotesWX.md
+ooffice --headless --convert-to pdf ReleaseNotes.md
+ooffice --headless --convert-to pdf README.md
```
dskrvk, line 44: This is nice, but adds a dependency on openoffice. Is it a standard expectation to provide docs in pdf format?

install/rpm/Makefile
carlbyington2: I don't know why that was done. This Makefile is only used by developers, and presumably they can control their environment.

install/rpm/Makefile
carlbyington2: yes, that is cleaner, thanks!

install/rpm/Makefile
carlbyington2: ok

install/rpm/Makefile
carlbyington2: the cd ../.. would not be seen by the rm -rf if they were inside the same subshell.

install/rpm/pwsafe.spec.tmpl
carlbyington2: The dependency is only on the BuildRequires so it is only needed to build the binary rpm. In any case, Fedora workstation by default installs that.

install/rpm/Makefile
```diff
@@ -1,58 +1,23 @@
...
-XMLDIR  := ../../xml
-HELPDIR := ../../help
-MODIR   := ../../src/ui/wxWidgets/I18N/mos
-RELEASEDIR := ../../Releases
-
-# I'd prefer the rpm build to be done under this directory,
-# but rpmbuild's default is ~/rpmbuild, and life's too short to find out
-# where to change this.
+MAJOR := $(shell grep MAJOR.= ../../version.wx | awk '{print $$3}')
+MINOR := $(shell grep MINOR.= ../../version.wx | awk '{print $$3}')
```
nobugshere, line 4: VER_REV is the third part of the version number.  It's not always 0.

carlbyington2: OK. I was trying to make it consistent with the version numbering at https://sourceforge.net/projects/passwordsafe/files/Linux/1.23.0/ but this is more consistent with normal Fedora versioning.

README.LINUX.DEVELOPERS.md
ronys: Can you elaborate on the problems with the cpack generated RPM?
Ideally both 'make -C install/rpm' and 'cpack -G RPM' should generate valid rpm packages.

carlbyington2: cpack generates a valid binary rpm, but it does not meet the Fedora packaging guidelines. Among other things, Fedora needs a source rpm containing the source code and a spec file where the build section has the commands needed to build the binaries. The changelog section needs to include the actual changes in that version, so that "rpm -q --changelog passwordsafe" will show those entries. The license and documentation files need to be tagged, so that "rpm -q --licensefiles passwordsafe" (or --docfiles) lists those files.

carlbyington2: All those Fedora requirements are the responsibility of the packager, not the upstream developers. There is no Fedora requirement that the upstream project make any changes to accommodate Fedora.

install/rpm/Makefile
```diff
@@ -1,58 +1,24 @@
...
+	rm passwordsafe.spec
+	find $(RPMBUILDDIR) -name passwordsafe*rpm | grep $(VERSION)
 
-clean:
-	@$(RM)  $(RPMBUILDDIR)
+tarball:
+	mkdir -p $(RPMBUILDDIR)/SOURCES
+	[ ! -f ../../version.wx ] && echo "Couldn't find version.wx. Make sure current directory $(PWD) ends with install/rpm" || /bin/true
+	[ ! -d ../../CMakeModules ] && echo "Missing CMakeModules. Make sure current directory $(PWD) ends with install/rpm" || /bin/true
+	(cd ../..; rm -rf build; tar cfz $(RPMBUILDDIR)/SOURCES/passwordsafe-$(VERSION).tar.gz *)
```
ronys: 'rm -rf build' is problematic, since that's not necessarily where cmake artifacts are built.
- In the default cmake configurations for this project, they're under 'out'
- Personally, I use 'build' for release, and 'build-dbg' for debug builds.
- Nuking the build directory is an unexpected side effect, to say the least...

I suggest using 'git archive' to create the tarball, as is done in the toplevel Makefile.linux

ronys (changes requested):

install/rpm/Makefile
carlbyington2: done

PR #1771, opened 2026-05-08
[wx] Fix - Help sometimes not available after changing language

froggie2523:
Issue: After changing language, Help for property sheet dialogs does not open. Opening Help from a regular dialog window or using the Help > Get Help command works fine.
This only affects pages in Options and Add/Edit dialog.
Root cause: This is because page labels get translated on changing language but Help definitions for the pages are initialized when the program starts (with a different language).
So, after the program is restarted, a new Help map is loaded and it works OK.

Fix: Decouple localization from lookup logic for Help.
Fix: AddEditPropSheetDlg::CreateAttachmentPanel() is using ID_PANEL_ADDITIONAL for the ID (probably a copy & paste error).

pwsafe_1 23-wxWidgets-bug-Help-Language-01

### DIFF
--- a/src/ui/wxWidgets/AddEditPropSheetDlg.cpp
+++ b/src/ui/wxWidgets/AddEditPropSheetDlg.cpp
@@ -405,9 +405,9 @@ void AddEditPropSheetDlg::CreateControls()
   m_PasswordPolicyOwnSymbolsTextCtrl->SetValue(m_Symbols);
 }
 
-wxScrolledWindow *AddEditPropSheetDlg::CreateScrollablePage(wxWindowID id)
+wxScrolledWindow *AddEditPropSheetDlg::CreateScrollablePage(wxWindowID id, const wxString &pageName)
 {
-  auto *page = new wxScrolledWindow(GetBookCtrl(), id, wxDefaultPosition, wxDefaultSize, wxTAB_TRAVERSAL | wxVSCROLL);
+  auto *page = new wxScrolledWindow(GetBookCtrl(), id, wxDefaultPosition, wxDefaultSize, wxTAB_TRAVERSAL | wxVSCROLL, pageName);
 #if wxCHECK_VERSION(3,1,0)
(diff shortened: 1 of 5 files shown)


src/ui/wxWidgets/AddEditPropSheetDlg.cpp
```diff
@@ -1024,7 +1024,7 @@ wxScrolledWindow* AddEditPropSheetDlg::CreateAttachmentPanel()
   ID_BUTTON_EXPORT = wxWindow::NewControlId();
   ID_BUTTON_REMOVE = wxWindow::NewControlId();
 
-  auto *panel = CreateScrollablePage(ID_PANEL_ADDITIONAL);
+  auto *panel = CreateScrollablePage(ID_PANEL_ATTACHMENT, "Attachment");
```
ronys, line 1027: Nice catch!

src/ui/wxWidgets/AddEditPropSheetDlg.cpp
```diff
@@ -454,7 +454,7 @@ void AddEditPropSheetDlg::RelaxScrollablePageSizes()
 
 wxScrolledWindow* AddEditPropSheetDlg::CreateBasicPanel()
 {
-  auto *panel = CreateScrollablePage(ID_PANEL_BASIC);
+  auto *panel = CreateScrollablePage(ID_PANEL_BASIC, "Basic");
```
ronys, line 457: Shouldn't all the string be in _() for gettext?

froggie2523: The page name must match the string in HelpMap.h for the lookup to work properly.
The page title (GetPageText()) is now only used in the error message when Help cannot be displayed, I kept that code.

dskrvk: I feel like these values should have at least been defined as constants and shared between HelpMap and other usages to ensure consistency.

PR #1773, opened 2026-05-11
Fix runaway DPI font scaling

martona:
## Summary

This fixes runaway font growth in the Windows/MFC UI when running on high-DPI displays, especially at 200% scaling.

The visible symptom is that the tree/list font and password-entry font can become enormous, to the point where normal 10-12 point UI text is rendered at something closer to poster size. The issue is easiest to trigger on high-DPI systems because the existing scaling bug compounds much faster there, but the underlying problem is in the font setup code rather than in the user's display configuration.

## Root Cause

The Windows UI font setup flows through the `Fonts` singleton in `src/ui/Windows/Fonts.cpp`. The default font `LOGFONT` values are stored as static globals, for example:

- `dfltPasswordLogfont`
- `dfltTreeListFont`
- `dfltAddEditLogfont`
- `dfltNotesLogfont`
- `DragFixLogfont`

Those defaults are adjusted for DPI by `tweakFontSizes()`. The bug is that `tweakFontSizes()` was being called from `Fonts::GetInstance()` every time `GetInstance()` was invoked, and it mutates the static `LOGFONT::lfHeight` values in place.

At 200% DPI, that means a default height such as `-13` can become `-26`, then `-52`, then `-104`, and so on as the singleton is requested repeatedly during normal UI initialization and redraw paths. The singleton object itself is only created once, but the static defaults were still being repeatedly modified before returning the existing singleton.

That repeated mutation is enough to explain the enormous tree/list text and also why password-entry fields are affected: both paths use the same `Fonts` singleton and default font state.

There was a second related issue in the same area. The font setters were pre-scaling point sizes before passing them to `CFont::CreatePointFontIndirect()`. MFC's `CreatePointFontIndirect()` expects `lfHeight` to contain the point size in tenths of a point and then performs the point-to-device conversion internally. Pre-scaling those values by DPI before calling `CreatePointFontIndirect()` makes point-size based fonts too large, especially on high-DPI displays.

## What Changed

This PR keeps the fix deliberately narrow and contained to `src/ui/Windows/Fonts.cpp`.

1. `tweakFontSizes()` is now guarded so it can only adjust the static default `LOGFONT` values once.

   This preserves the existing intent of making the built-in defaults DPI-aware, while preventing repeated calls to `Fonts::GetInstance()` from compounding the default heights.

2. `Fonts::GetInstance()` now calls `tweakFontSizes()` only when the singleton is actually being initialized.

   This restores the earlier singleton behavior and avoids mutating shared default font state during routine access.

3. The `CreatePointFontIndirect()` paths no longer pre-scale `iPtSz` by DPI.

   The setters now place the stored point size directly into `lf.lfHeight`, matching the API contract expected by MFC.

4. The `CreateFontIndirect()` paths no longer make a local copy solely to apply another DPI multiplier.

   The relevant `LOGFONT` data is passed through directly. This avoids applying additional scaling on top of defaults or stored font preferences that are already represented in `LOGFONT` units.

## Validation

Validated locally on Windows 11 with the generated CMake Visual Studio project:

```text
MSBuild _build\pwsafe.vcxproj /p:Configuration=Debug /p:Platform=x64
```

The Debug x64 build completed successfully and produced:

```text
_build\Debug\pwsafe.exe
```

Finally, ran the resulting executable and got the expected font size, matching the rest of the application and system.

## Pics or it did not happen

Before:

pwsafe-before

After:

pwsafe-after

### DIFF
--- a/src/ui/Windows/Fonts.cpp
+++ b/src/ui/Windows/Fonts.cpp
@@ -54,6 +54,11 @@ static LOGFONT DragFixLogfont = {
 
 static void tweakFontSizes()
 {
+  static bool tweaked = false;
+  if (tweaked)
+    return;
+  tweaked = true;
+
   const UINT defDPI = WinUtil::defDPI;
   const UINT curDPI = WinUtil::GetDPI();
 
@@ -71,8 +76,8 @@ static void tweakFontSizes()
 
 Fonts *Fonts::GetInstance()
 {
-  tweakFontSizes();
   if (self == nullptr) {
+    tweakFontSizes();
     self = new Fonts();
   }
   return self;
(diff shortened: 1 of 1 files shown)


dskrvk (approved):

ronys:
Hi @martona 

Thanks for the very clear description and elegant fix.

I wonder if this also addresses #1743, which I had trouble reproducing.

martona:
You're very welcome! And no, I don't think so - at least I don't see an obvious path to crashing in the IDC_RESETFONT handler, font sizes being crazy or not.

PR #1774, opened 2026-05-11
Fix main window position restore scaling

### LINKED ISSUE #1092: Size & position not maintained on scaled displays
On systems with high resolution displays and Scale > 100%, the last size & position of the main window is incorrectly restored.

martona:
## Summary

Fixes #1092 - reports complaining about the window drifting down and right.

This removes the restore-time DPI scaling from the Windows main-window placement path.

The main window already saves its position and size from `GetWindowRect()`, which returns the actual window rectangle in screen coordinates. For a per-monitor-DPI-aware window, those coordinates are already in physical pixels for the current display. Feeding that saved rectangle back into `SetWindowPlacement()` should therefore use the same coordinates directly.

## Root Cause

`DboxMain::PlaceWindow()` was scaling the saved rectangle by the current DPI before assigning it to `WINDOWPLACEMENT::rcNormalPosition`.

At 200% display scaling, that means a rectangle saved as physical pixels is multiplied by 2 when restored. Once the restored larger rectangle is saved again, the next launch multiplies that already-scaled rectangle again. The visible result is that the main window drifts down and right and grows larger on every restart, with the error matching the display scale factor.

The save side uses `GetWindowRect()` in the normal move/resize paths, so the preference values are not 96-DPI logical coordinates. The restore side should not reinterpret them as if they were.

## What Changed

`PlaceWindow()` now assigns the saved rectangle directly:

```cpp
wp.rcNormalPosition = *pRect;
```

The previous `WinUtil::GetDPI()` / `MulDiv()` adjustment block was removed.

This keeps the save/restore contract simple and symmetric: save the actual window rectangle, restore that same rectangle.

## Validation

Built and tested on Windows 11 with the Debug build at 200% DPI. Also tested at 100% to ensure consistent behavior.

## Pics or it did not happen

(For the avoidance of doubt, these screenshots represent the pre-patch state of the code, post-patch the positions are identical.)

Before exiting the app:

pwsafe-saveposition

After reopening the app:

pwsafe-restoreposition

### DIFF
--- a/src/ui/Windows/DboxMain.cpp
+++ b/src/ui/Windows/DboxMain.cpp
@@ -3811,19 +3811,10 @@ void DboxMain::PlaceWindow(CWnd *pWnd, CRect *pRect, UINT uiShowCmd)
   WINDOWPLACEMENT wp = {sizeof(WINDOWPLACEMENT)};
   HRGN hrgnWork = WinUtil::GetWorkAreaRegion();
 
-  auto curDPI = WinUtil::GetDPI(nullptr); // need system value
-
-  // adjust rect for current scale factor
-  CRect adjRect;
-  adjRect.bottom = MulDiv(pRect->bottom, curDPI, WinUtil::defDPI);
-  adjRect.top =    MulDiv(pRect->top, curDPI, WinUtil::defDPI);
-  adjRect.left =   MulDiv(pRect->left, curDPI, WinUtil::defDPI);
-  adjRect.right =  MulDiv(pRect->right, curDPI, WinUtil::defDPI);
-
   pWnd->GetWindowPlacement(&wp);  // Get min/max positions - then add what we know
   wp.flags = 0;
   wp.showCmd = uiShowCmd;
-  wp.rcNormalPosition = adjRect;
+  wp.rcNormalPosition = *pRect;
 
   if (!RectInRegion(hrgnWork, &wp.rcNormalPosition)) {
     if (GetSystemMetrics(SM_CMONITORS) > 1)


ronys:
Thanks!

ColeBantam:
This is still not working in 3.72.01 :(

Window position and size is still not saved/restored correctly.

Windows 11 Pro, 175% Display Scale, 3840x2160 Resolution

martona:
@ronys the position bug is back in 3.72.1 because commit e6491bef3 ("Fix Windows scaling issue", 27 Jun) re-added the MulDiv scaling in PlaceWindow() that PR #1774 had removed. It's stacked on top of the old fix: wp.rcNormalPosition = *pRect; is still there (DboxMain.cpp:3824), immediately overwritten by the four new MulDiv lines below it, so that line is now dead code.

The re-added scaling double-applies the scale factor. Save (OnMove/OnSize -> GetWindowRect -> SetPrefRect) and restore (GetPrefRect -> PlaceWindow -> SetWindowPlacement) work in the same coordinate space, so the stored rect round-trips 1:1 with no conversion: that was the point of #1774. GetMonitorDPI() returns the real monitor DPI regardless of process awareness, so MulDiv(coord, 192, 96) at 200% pushes the window down-right and grows it 2x. That's @ColeBantam 's report at 175%.

I suspect the reason this keeps flip-flopping is because the process is manifest-DPI-unaware but flips individual threads to PMv2 (SetThreadDpiAwarenessContext in DboxMain::DoModal, PWDialog::DoModal, Fonts.cpp). Under a PMv2 thread GetWindowRect/SetWindowPlacement are physical; under an unaware thread they're virtualized. So depending on which thread saved vs. restored, a MulDiv looks either necessary or wrong.

Suggest reverting the PlaceWindow scaling back to #1774 and, to stop this recurring, making save and restore run under the same awareness context (or converting symmetrically). Happy to put up a PR if useful.

ronys:
Hi @martona 

Thanks for your analysis and offer. I'd much appreciate a PR as you suggested.

PR #1778, opened 2026-05-13
[win] Move Notes and Custom Fields

ronys:
to a common propsheet under the Basic prop page. This results in a less cluttered dialog, hiding the custom fields from the majority of folks who will never use it.

TODO - update help files.


src/ui/Windows/AddEdit_Basic_Tabs.cpp
```diff
@@ -0,0 +1,93 @@
...
+
+  if (CPropertySheet::Create(pParentWnd, WS_CHILD | WS_VISIBLE | WS_TABSTOP, 0) == FALSE)
+    return false;
+
+  MoveWindow(rect);
+  LayoutPages();
+  return true;
+}
+
+BOOL CAddEdit_Basic_Tabs::OnInitDialog()
```
dskrvk, line 44: Should all methods be consistent in returning either a BOOL or a bool, or are these signatures dictated by MFC?

ronys: Well, those that inherit from an MFC base class should probably be BOOL, for consistency.
(This is really a bit of archeology, since BOOL is a  typedef of int - this predates the addition of 'bool' as a builtin type...)

PR #1782, opened 2026-05-17
issues found by clang static analyzer

carlbyington2:
src/test/TOTPTest.cpp referenced objects after std::move().

src/ui/wxWidgets/PWFiltersStringDlg.cpp had unnecessary tests for null m_TextCtrlValueString which is initialized in the constructor.

src/ui/wxWidgets/TreeCtrl.cpp did not protect GUI_REFRESH_ENTRY from a null pointer dereference.

### DIFF
--- a/src/test/TOTPTest.cpp
+++ b/src/test/TOTPTest.cpp
@@ -79,12 +79,12 @@ TEST(TOTPTest, key_decoder_copy_move_cleanup)
 
   const unsigned char* dec_ptr = dec.get_ptr();
   RFC4648_Base32Decoder dec3(std::move(dec));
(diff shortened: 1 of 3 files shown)


src/test/TOTPTest.cpp
ronys, line 1: Please remove rather than comment out the incorrect checks.

ronys (changes requested):
Thanks, one small request before merging this in.

src/test/TOTPTest.cpp
carlbyington2: done.

ronys (approved):

PR #1789, opened 2026-05-23
possible change to replace pugi::convertBuffer with UTF8Conv, assuming the report files are always utf-8.

carlbyington2:
The code is much simpler if we assume the report files are always utf-8.


src/core/Report.cpp
```diff
@@ -169,155 +169,6 @@ bool CReport::SaveToDisk(const stringT &out_dir)
     return false;
   }
 
-  // **** MOST LIKELY ACTION ****
-  // If file is new/empty, write BOM, as some text editors insist!
```
nobugshere, line 173: Let's keep the BOM  comment somewhere

carlbyington2: done

carlbyington2: What commonly used editor does not handle bom-less utf-8 input? Google only found original notepad on Windows, but does anyone really use that to look at the various report file?

ronys:
Thanks! I love it when less code does (almost) the same work.

Issue #1196, opened 2024-03-22
Support dark mode?


ronys:
A user (DL) wrote:

Hi guys - I love the app, it's terrific. Just wondering whether there is a dark mode for PC on the roadmap? Selfishly, I suffer from floaters in my vision so dark backgrounds are hugely beneficial for me but I also know that dark modes are incredibly popular anyway. It would be great to know if this is something you intend to introduce.

igor2x:
@rony, do you know what instance of Password Safe "DL" is referring to? Is this Password Safe on Windows, maybe?

I did some test on Ubuntu 23.10 on Password Safe 1.19-pre and it looks surprisingly good. Probably wxWidgets is by default dark mode friendly.

I have gone through the program and maybe there are some ways improvement may be done.

1. At master password entry dialog the title is image instead of ordinary text. Can title image be replaced by ordinary text? Password Safe icon looks too dark on dark background.
![image](https://github.com/pwsafe/pwsafe/assets/10895030/e69d00a9-add6-4ca0-8733-76143700892d)

2. Icons in toolbar are too bright and look little bit blurry. In general they look to me like from some older era and can probably be modernized on light mode also.
![image](https://github.com/pwsafe/pwsafe/assets/10895030/4ca9bf1b-20e6-4fe1-b1fa-f5ff03c0d2b7)

3. Too dark red color.
![image](https://github.com/pwsafe/pwsafe/assets/10895030/4c93a6c6-9815-477d-a635-854c5b2a5330)

4. Icons are too bright and too large. Some of the issues related to those icons are already reported in https://github.com/pwsafe/pwsafe/issues/1167
![image](https://github.com/pwsafe/pwsafe/assets/10895030/1f96f6ef-c5f2-42cd-b6e1-a7c3de1447a7)

martona:
If there's interest from the maintainers, I could work on a PR. I do have a version of the app that's in dark mode, but it's very much "just the beaten path", and not much attention was paid to maintaining light mode compatibility - even though the code changes are minimal so it should be effortless to bolt back on.

<img width="1528" height="980" alt="Image" src="https://github.com/user-attachments/assets/5be7e234-1d9c-4f0c-8e89-2e7f14f4904d" />

ronys:
I'd love to see dark mode properly supported, both in Windows (as the original submitter requested) and in wxWidgets.

Let's keep this ticket for Windows dark mode support. I'll create a duplicate for wxWidgets dark mode tuning.

[...]

PR #1790, opened 2026-05-23
Dark Mode Support on Windows

martona:
## Summary

Windows dark mode support through a local copy of `darkmode32plus` and wiring it into Password Safe dialogs and various MFC controls where applicable.

## Technical Notes

- Dark mode support is behind `PWSAFE_USE_DARKMODE32`.
- The new `win32-darkmode32` static library is included only for MSVC Visual Studio CMake builds via `include_external_msproject`.
- `ThisMfcApp::InitInstance()` initializes dark mode.
- Core dialog base classes now install dark-mode erase/background handling, so most derived dialogs pick it up automatically.
- Some dialogs and custom controls need explicit handling because they paint themselves or do not fully inherit the common dialog path.
- `MessageBoxW` support is handled through a scoped, thread-local CBT hook that subclasses the native message box while it is active.
- `LICENSE.PWSAFE.md` documents provenance for the copied dark mode library.

## Gotchas

- The dark mode implementation relies on Windows dark mode APIs and behavior that may vary across Windows 10/11 builds. 
- Owner-drawn controls are the highest-risk area
- High contrast mode was not tested but this patch shouldn't make things worse (or better) with it enabled.
- Native/common dialogs and message boxes may still have edge cases because Windows owns parts of their painting.

## Possible Regressions

- Light-mode painting regressions in controls touched by conditional dark mode logic.
- Dialog flicker or stale background painting from added erase/background subclasses.
- Incorrect text/background contrast in disabled, focused, selected, or hover states.

## Recommendation

I kept the dark mode integration behind explicit `#ifdef PWSAFE_USE_DARKMODE32` guards so the PR has a clear integration boundary. If the PR is adopted, a follow-up cleanup could add a small compatibility/shim header with no-op `DarkMode::` stubs for builds without `darkmode32plus`, which would make the call sites easier to read. Also, introducing intermediary base classes for Password Safe dialogs/property sheets could help localize the required boilerplate calls (e.g. `CPropertySheet` -> `CDarkModePropertySheet` -> `CAddEdit_Basic_Tabs`)


ronys:
Looks promising!

Some comments/suggestions:
1. ui/win32-darkmode32 should be moved to ui/windows/darkmode32, since (a) the ui subdirectories (wx, windows and cli) each map to a separate executable, and (b) it's really windows-specific.
2. For consistency, it would be nice for this to have its own CMakeLists.txt file, instead of vcxproj and vcxproj.filter
3. I'd remove #ifdef PWSAFE_USE_DARKMODE32 entirely, or at least keep it in the compatibility shim/header as you described in your Recommendation section. If you keep it, then it should be configurable via the main CMakeLists.txt file.
4. Finally, I'd add a dropdown in the Manage->Options->Display preferences to allow the user to set this: Display Mode: System(default)|Light|Dark. This will also make testing much easer.

martona:
I believe that covers it, provided the CI checks pass.

ronys:
The normal theme looks 'flat' compared to the current release. In the following, the left side is a screenshot from release 3.71, and the right is from this branch. Note the buttons and tabs, as well as the different shades of green in the password strength indicator.
AddEdit

I think this feature should leave the existing normal theme untouched.

ronys:
In addition, trying to switch the debug build to the dark theme, I hit an assertion in line 55 of StateBitmapManager.cpp

Here's the full stack trace:
```
>	pwsafe.exe!CStateBitmapManager::GetStateBitmap(unsigned int nIdBitmap) Line 55	C++
 	pwsafe.exe!CStateBitmapManager::GetBitmapInfo(unsigned int nIdBitmap, tagBITMAP * pBmpInfo, long * pBmWidthDpi, long * pBmHeightDpi) Line 74	C++
 	pwsafe.exe!CPWStatusBar::DrawItem(tagDRAWITEMSTRUCT * lpDrawItemStruct) Line 145	C++
 	pwsafe.exe!CStatusBar::OnChildNotify(unsigned int message, unsigned __int64 wParam, __int64 lParam, __int64 * pResult) Line 553	C++
 	pwsafe.exe!CPWStatusBar::OnChildNotify(unsigned int message, unsigned __int64 wParam, __int64 lParam, __int64 * pResult) Line 181	C++
 	pwsafe.exe!CWnd::SendChildNotifyLastMsg(__int64 * pResult) Line 3396	C++
 	pwsafe.exe!CWnd::ReflectLastMsg(HWND__ * hWndChild, __int64 * pResult) Line 3433	C++
 	pwsafe.exe!CWnd::OnDrawItem(int __formal, tagDRAWITEMSTRUCT * lpDrawItemStruct) Line 1246	C++
 	pwsafe.exe!DboxMain::OnDrawItem(int nIDCtl, tagDRAWITEMSTRUCT * lpdis) Line 4504	C++
 	pwsafe.exe!CWnd::OnWndMsg(unsigned int message, unsigned __int64 wParam, __int64 lParam, __int64 * pResult) Line 2579	C++
 	pwsafe.exe!CWnd::WindowProc(unsigned int message, unsigned __int64 wParam, __int64 lParam) Line 2100	C++
 	pwsafe.exe!DboxMain::WindowProc(unsigned int message, unsigned __int64 wParam, __int64 lParam) Line 2837	C++
 	pwsafe.exe!AfxCallWndProc(CWnd * pWnd, HWND__ * hWnd, unsigned int nMsg, unsigned __int64 wParam, __int64 lParam) Line 265	C++
 	pwsafe.exe!AfxWndProc(HWND__ * hWnd, unsigned int nMsg, unsigned __int64 wParam, __int64 lParam) Line 418	C++
 	user32.dll!00007ffee191ef5c()	Unknown
 	user32.dll!00007ffee191e9de()	Unknown
 	comctl32.dll!00007ffed3c69eca()	Unknown
 	comctl32.dll!00007ffed3c69e08()	Unknown
 	pwsafe.exe!DarkMode::WindowEraseBgSubclass(HWND__ * hWnd, unsigned int uMsg, unsigned __int64 wParam, __int64 lParam, unsigned __int64 uIdSubclass, unsigned __int64 dwRefData) Line 5785	C++
 	comctl32.dll!00007ffed3c69eca()	Unknown
 	comctl32.dll!00007ffed3c69e08()	Unknown
 	pwsafe.exe!DarkMode::WindowCtlColorSubclass(HWND__ * hWnd, unsigned int uMsg, unsigned __int64 wParam, __int64 lParam, unsigned __int64 uIdSubclass, unsigned __int64 dwRefData) Line 5939	C++
 	comctl32.dll!00007ffed3c69eca()	Unknown
 	comctl32.dll!00007ffed3c69e08()	Unknown
 	pwsafe.exe!DarkMode::WindowNotifySubclass(HWND__ * hWnd, unsigned int uMsg, unsigned __int64 wParam, __int64 lParam, unsigned __int64 uIdSubclass, unsigned __int64 dwRefData) Line 6719	C++
 	comctl32.dll!00007ffed3c69eca()	Unknown
 	comctl32.dll!00007ffed3c69e08()	Unknown
 	pwsafe.exe!DarkMode::WindowMenuBarSubclass(HWND__ * hWnd, unsigned int uMsg, unsigned __int64 wParam, __int64 lParam, unsigned __int64 uIdSubclass, unsigned __int64 dwRefData) Line 7031	C++
 	comctl32.dll!00007ffed3c69eca()	Unknown
 	comctl32.dll!00007ffed3c69cc7()	Unknown
 	user32.dll!00007ffee191ef5c()	Unknown
 	user32.dll!00007ffee191e9de()	Unknown
 	pwsafe.exe!CSubclassWnd::WindowProc(unsigned int msg, unsigned __int64 wp, __int64 lp) Line 98	C++
 	pwsafe.exe!CCoolMenuManager::WindowProc(unsigned int msg, unsigned __int64 wp, __int64 lp) Line 150	C++
 	pwsafe.exe!CSubclassWnd::WindowProc(unsigned int msg, unsigned __int64 wp, __int64 lp) Line 98	C++
 	pwsafe.exe!CMenuTipManager::WindowProc(unsigned int msg, unsigned __int64 wp, __int64 lp) Line 42	C++
 	pwsafe.exe!HookWndProc(HWND__ * hwnd, unsigned int msg, unsigned __int64 wp, __int64 lp) Line 173	C++
 	user32.dll!00007ffee191ef5c()	Unknown
 	user32.dll!00007ffee191dfbb()	Unknown
 	user32.dll!00007ffee191d814()	Unknown
 	pwsafe.exe!DarkMode::paintStatusBar(HWND__ * hWnd, HDC__ * hdc, DarkMode::StatusBarData & statusBarData) Line 5021	C++
 	pwsafe.exe!DarkMode::StatusBarSubclass(HWND__ * hWnd, unsigned int uMsg, unsigned __int64 wParam, __int64 lParam, unsigned __int64 uIdSubclass, unsigned __int64 dwRefData) Line 5137	C++
 	comctl32.dll!00007ffed3c69eca()	Unknown
 	comctl32.dll!00007ffed3c69cc7()	Unknown
 	user32.dll!00007ffee191ef5c()	Unknown
 	user32.dll!00007ffee191e8cc()	Unknown
 	user32.dll!00007ffee19310c3()	Unknown
 	ntdll.dll!00007ffee36f14d4()	Unknown
 	win32u.dll!00007ffee0d91464()	Unknown
 	comctl32.dll!00007ffed3c19b2d()	Unknown
 	comctl32.dll!00007ffed3c199b0()	Unknown
 	comctl32.dll!00007ffed3c19308()	Unknown
 	user32.dll!00007ffee191ef5c()	Unknown
 	user32.dll!00007ffee191e9de()	Unknown
 	pwsafe.exe!CWnd::DefWindowProcW(unsigned int nMsg, unsigned __int64 wParam, __int64 lParam) Line 1101	C++
 	pwsafe.exe!CStatusBar::SetPaneText(int nIndex, const wchar_t * lpszNewText, int bUpdate) Line 441	C++
 	pwsafe.exe!CStatusBar::UpdateAllPanes(int bUpdateRects, int bUpdateText) Line 278	C++
 	pwsafe.exe!CStatusBar::SetIndicators(const unsigned int * lpIDArray, int nIDCount) Line 156	C++
 	pwsafe.exe!DboxMain::OnOptions() Line 329	C++
 	pwsafe.exe!_AfxDispatchCmdMsg(CCmdTarget * pTarget, unsigned int nID, int nCode, void(CCmdTarget::*)() pfn, void * pExtra, unsigned __int64 nSig, AFX_CMDHANDLERINFO * pHandlerInfo) Line 77	C++
 	pwsafe.exe!CCmdTarget::OnCmdMsg(unsigned int nID, int nCode, void * pExtra, AFX_CMDHANDLERINFO * pHandlerInfo) Line 372	C++
 	pwsafe.exe!CDialog::OnCmdMsg(unsigned int nID, int nCode, void * pExtra, AFX_CMDHANDLERINFO * pHandlerInfo) Line 85	C++
 	pwsafe.exe!CWnd::OnCommand(unsigned __int64 wParam, __int64 lParam) Line 2802	C++
 	pwsafe.exe!CWnd::OnWndMsg(unsigned int message, unsigned __int64 wParam, __int64 lParam, __int64 * pResult) Line 2114	C++
 	pwsafe.exe!CWnd::WindowProc(unsigned int message, unsigned __int64 wParam, __int64 lParam) Line 2100	C++
 	pwsafe.exe!DboxMain::WindowProc(unsigned int message, unsigned __int64 wParam, __int64 lParam) Line 2837	C++
 	pwsafe.exe!AfxCallWndProc(CWnd * pWnd, HWND__ * hWnd, unsigned int nMsg, unsigned __int64 wParam, __int64 lParam) Line 265	C++
 	pwsafe.exe!AfxWndProc(HWND__ * hWnd, unsigned int nMsg, unsigned __int64 wParam, __int64 lParam) Line 418	C++
 	user32.dll!00007ffee191ef5c()	Unknown
 	user32.dll!00007ffee191e9de()	Unknown
 	comctl32.dll!00007ffed3c69eca()	Unknown
 	comctl32.dll!00007ffed3c69e08()	Unknown
 	pwsafe.exe!DarkMode::WindowEraseBgSubclass(HWND__ * hWnd, unsigned int uMsg, unsigned __int64 wParam, __int64 lParam, unsigned __int64 uIdSubclass, unsigned __int64 dwRefData) Line 5785	C++
 	comctl32.dll!00007ffed3c69eca()	Unknown
 	comctl32.dll!00007ffed3c69e08()	Unknown
 	pwsafe.exe!DarkMode::WindowCtlColorSubclass(HWND__ * hWnd, unsigned int uMsg, unsigned __int64 wParam, __int64 lParam, unsigned __int64 uIdSubclass, unsigned __int64 dwRefData) Line 5939	C++
 	comctl32.dll!00007ffed3c69eca()	Unknown
 	comctl32.dll!00007ffed3c69e08()	Unknown
 	pwsafe.exe!DarkMode::WindowNotifySubclass(HWND__ * hWnd, unsigned int uMsg, unsigned __int64 wParam, __int64 lParam, unsigned __int64 uIdSubclass, unsigned __int64 dwRefData) Line 6719	C++
 	comctl32.dll!00007ffed3c69eca()	Unknown
 	comctl32.dll!00007ffed3c69e08()	Unknown
 	pwsafe.exe!DarkMode::WindowMenuBarSubclass(HWND__ * hWnd, unsigned int uMsg, unsigned __int64 wParam, __int64 lParam, unsigned __int64 uIdSubclass, unsigned __int64 dwRefData) Line 7031	C++
 	comctl32.dll!00007ffed3c69eca()	Unknown
 	comctl32.dll!00007ffed3c69cc7()	Unknown
 	user32.dll!00007ffee191ef5c()	Unknown
 	user32.dll!00007ffee191e9de()	Unknown
 	pwsafe.exe!CSubclassWnd::WindowProc(unsigned int msg, unsigned __int64 wp, __int64 lp) Line 98	C++
 	pwsafe.exe!CCoolMenuManager::WindowProc(unsigned int msg, unsigned __int64 wp, __int64 lp) Line 150	C++
 	pwsafe.exe!CSubclassWnd::WindowProc(unsigned int msg, unsigned __int64 wp, __int64 lp) Line 98	C++
 	pwsafe.exe!CMenuTipManager::WindowProc(unsigned int msg, unsigned __int64 wp, __int64 lp) Line 42	C++
 	pwsafe.exe!HookWndProc(HWND__ * hwnd, unsigned int msg, unsigned __int64 wp, __int64 lp) Line 173	C++
 	user32.dll!00007ffee191ef5c()	Unknown
 	user32.dll!00007ffee191e684()	Unknown
 	pwsafe.exe!AfxInternalPumpMessage() Line 181	C++
 	pwsafe.exe!CWinThread::PumpMessage() Line 900	C++
 	pwsafe.exe!CWinThread::Run() Line 629	C++
 	pwsafe.exe!CWinApp::Run() Line 787	C++
 	pwsafe.exe!AfxWinMain(HINSTANCE__ * hInstance, HINSTANCE__ * hPrevInstance, wchar_t * lpCmdLine, int nCmdShow) Line 47	C++
 	pwsafe.exe!wWinMain(HINSTANCE__ * hInstance, HINSTANCE__ * hPrevInstance, wchar_t * lpCmdLine, int nCmdShow) Line 26	C++
 	pwsafe.exe!invoke_main() Line 123	C++
 	pwsafe.exe!__scrt_common_main_seh() Line 288	C++
 	pwsafe.exe!__scrt_common_main() Line 331	C++
 	pwsafe.exe!wWinMainCRTStartup(int __formal, const wchar_t *) Line 17	C++
 	kernel32.dll!00007ffee1667374()	Unknown
 	ntdll.dll!00007ffee369cc91()	Unknown

```

martona:
The flatter rendering appears to come from the Common Controls v6 manifest dependency that came in with the dark-mode work. That opt-in mechanism dates back to Windows XP visual styles, but the actual rendering follows the current Windows common-controls/theme implementation, so it changes the app’s baseline native look on modern Windows even outside dark mode.

Without Common Controls v6, Windows does not expose the themed common-control behavior/messages we need consistently, and the dark-mode subclassing/custom draw paths don’t get the expected surface to work against. So the visual-style change is unfortunately coupled to making dark mode viable.

I have, however, changed the initialization path to obey “Classic” mode when Light mode is selected, and not DarkMode32's "Light" theme. Visually they seem identical, but this should be futureproof.

The debug assert was an existing bug: the status bar’s screen-capture pane could be owner-drawn before statustext[SB_SCR_CAP] had been initialized. The dark-mode subclassing/repaint made that easier to hit, but the fix is to initialize that pane bitmap id before resetting/redrawing the status bar.

ronys:
The debug assert is indeed fixed, however, a couple of problems remain.
1. The "flat" issue is unchanged. The background has also changed from grey to white. The green in the password strength indicator matches, though.
2. When switching from dark back to light, any menu that's been opened in dark mode appears all dark when first displayed, and still looks odd after the mouse has been dragged across all items. See below.

AddEdit2
menu1
menu2

martona:
Fixed the remaining dark-mode rendering issues:

1. Menus opened in dark mode no longer stay dark after switching to Light. The popup background brush (MIM_BACKGROUND) is a property of the HMENU and persists across shows; it was only ever set for dark mode and never cleared, so a menu first shown dark kept its dark background in Light. It's now reset to the default themed background when not in dark mode.
2. The Options dialog's left navigation pane could come up blank in dark mode - icons and labels missing until a repaint. The dark-mode WM_ERASEBKGND handler fills the whole sheet client and returns TRUE, and since child controls share the top-level surface, that erase was painting over the nav pane after it had been drawn, with nothing repainting it. WS_CLIPCHILDREN on the sheet keeps the erase out of the child rectangles.
3. The same nav pane was drawing in light colors in dark mode (it's an MFC feature-pack control that reads its colors through MFC's own GetSysColor, which the comctl32-scoped hook doesn't cover) - now colored explicitly.

All three are gated on dark mode; Light/Classic rendering is untouched.

That leaves the "flat" appearance and the grey -> white background as the one thing still open, and I want to be straight about it: it isn't a bug I can fix, it's a direct consequence of the Common Controls v6 manifest. Dark mode only works if the app opts into visual styles; the theming APIs and change notifications the library depends on don't exist otherwise. And the moment an app opts in, Windows draws the standard controls in their themed style and uses the themed dialog background.

Visual styles are how Windows has drawn controls by default since XP introduced them in 2001, so the themed (flatter, lighter) appearance is just the standard rendering apps have used for the last twenty-odd years. The raised/3D grey look is the older, un-themed path that predates it - pwsafe has stayed on that pre-XP rendering all along by never carrying the v6 manifest, which is why it's looked a bit different from the rest of the desktop. By carrying the manifest, apps could follow along as Windows itself updated its look over the years. I realize the shift can read as a downgrade if the classic look is what you're used to, but functionally it's pwsafe coming into line with how Windows has rendered apps for two decades - and it's the same opt-in that makes dark mode possible at all. The manifest applies app-wide at link time, so it affects Light mode too; there's no switch that yields themed dark mode and the classic 3D look at once.

So this one's a design call, and it's yours to make:

- If the themed look is acceptable, this is finished.
- If particular parts grate I can probably soften those individually (e.g. bring the light-mode background back closer to the old grey) while keeping the manifest, but honestly we're likely to end up with something that's neither fish nor fowl.
- The only thing that genuinely can't coexist is the manifest and the exact classic 3D rendering: it's themed-look-with-dark-mode, or classic-without.

Whichever way you'd like to go, I'm happy to do the work.

ronys:
Thanks for spelling out the themed appearance issue.
I'm inclined to move PasswordSafe's UI forward into the 2020's, even if it means redoing many images in the help file...

Note that there's a problem with some disabled static text controls in dark mode:
image
The "(at least" is blurry, at least on my display.
The blurriness extends to the other texts when high contrast is enabled.

There are also some graphics that should have "dark mode" versions, such as the Authentication Code pie chart and text display:

image

Finally, I think the checkmarks and right triangles in the menus need to be a light color in dark mode. Currently they're hard to see:

image

martona:
Good progress this round: got through most of what was left, hit a few new snags:

- Menus - the dark popups and the checkmark glyphs are both sorted now.
- Auth-code pie chart - dark (or light, depending).
- The "blurry" disabled text - I believe that to be a contrast issue; please confirm if fixed.

The new items are more stubborn: the keyboard-shortcut hotkey field, and the expiry date picker (all on Add/Edit Additional or Date tabs) are all Common Controls widgets that seem very resistant to theming. As far as I know nobody dark-styles them, not even Microsoft (outside of HC mode but more on that later). The only thing that fully fixes them is either dropping these three controls and reimplementing them as plain Win32, or owner-drawing them fully, including light/dark/HC compatibility. That's real work and a meaningfully bigger change than this PR should carry, but I'm willing to do it.

On High Contrast: I've mitigated it, sort of. Dark mode lib now detects HC and steps aside completely, deferring to the OS palette, which clears up the worst of it. HC light themes look decent, HC dark themes (e.g. Night Sky) don't. Getting it perfect runs into an unkown amount of legacy owner-draw (which does exclusively light-compatible painting), so I'd lean toward treating full HC polish as its own follow-up rather than a blocker here.

Think of HC as a shortcut the OS tells apps to take: "let me handle the colors". It stops all theming, and issues old-style CTLCOLOR messages only. All Windows components obey it of course, and pwsafe mostly does too - but not everywhere.

Let me know how you'd like to handle it, but I think dark mode itself is in good shape now, and HC is genuinely separable.

ronys:
Much better!

About the keyboard shortcut and date picker (you mentioned 3, but described only these 2?): I think they're rarely used in PasswordSafe, so not worth the effort of reimplementing (!) or owner drawing them. I'd list this as a known limitation with darkmode support.

I'm fine with high contrast being treated as a separate PR. (I know of at least one visually impaired user who's dependent on this, and is awaiting improvements for this mode.)

There's one remaining problem I see with this PR: If you activate a filter while in dark mode, (View->Subviews or View->Filters->...) then the tree view texts are red, which is fine, but when deactivating the filter, they're redrawn in black:

image

martona:
Filters fixed. As to the 3-controls-vs-2 discrepancy: there was a ListView on the same property sheet as the hotkey-picker: I ended up fixing it as it turned out to be easier than I thought.

Somewhat related: there's an assert being hit if a filter collapses the database view to zero. (No valid entries to display after applying a filter.) The one-line fix is to guard the list branch on GetItemCount() like SelectFindEntry already does; but I felt it was out of scope for this PR - I did not want to bundle core fixes into something unrelated. And it's debug-only: in release, MFC's GetItemData memsets its LVITEM to zero, so the failed fetch returns NULL, which SelectFirstEntry handles fine (same as the empty-tree case).

ronys:
Nice.
I'll fix the assert separately.

One more issue: The "Customize Toolbar" dialog appears to have been overlooked:
image

martona:
CToolBarCtrl::Customize() sends TB_CUSTOMIZE to the toolbar, and comctl32 itself creates and runs the "Customize Toolbar" dialog internally - modal, with its own message loop, its own dialog template baked into comctl32.dll, its own two owner-drawn listboxes. pwsafe never sees an HWND, has no CDialog, no OnInitDialog, no template. Our entire dark-theming model attaches to windows we create; here there's literally no pwsafe code path to hook into. So "overlooked" isn't quite right; there was nothing to not-overlook.

But I get it... it's annoying. I *could* do a WH_CBT hook -> into a WH_CALLWNDPROC thread-local hook -> catch WM_INITDIALOG -> attempt theming. It might work (~same approach works for ::MessageBox) but then again the owner-drawn listboxes are going to put up a fight. It's likely to end with a half-result which is arguably worse than a no-result.

I'd file it with the hotkey field and date picker as a known darkmode limitation: same root cause (a Common-Controls UI we'd have to replace to theme).

ronys:
Adding reference to Issue:
Resolves #1196

PR #1792, opened 2026-05-24
cleanup pwsafe-cli library reference; switch to bom-less report files

carlbyington2:
pwsafe-cli had an unnecessary reference to libmagic.

I think any modern editor will handle bom-less utf-8, and it simplifies the code a bit.

### DIFF
--- a/src/ui/cli/Makefile
+++ b/src/ui/cli/Makefile
@@ -40,7 +40,7 @@ else
 	GTEST_HEADERS=/usr/include/gtest/*.h /usr/include/gtest/internal/*.h
 	GTEST_SRCS=$(GTEST_DIR)/src/*.cc $(GTEST_DIR)/src/*.h $(GTEST_HEADERS)
 	GTEST_OBJ = $(OBJPATH)/gtest-all.o
-	EXTLIBS=-luuid -lxerces-c -lpthread -lmagic
+	EXTLIBS=-luuid -lxerces-c -lpthread 
 	TEST_EXTLIBS=$(EXTLIBS)
 	WX_CONFIG?=/usr/bin/wx-config
 endif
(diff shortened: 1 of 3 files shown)


src/ui/cli/Makefile
```diff
@@ -40,7 +40,7 @@ else
 	GTEST_HEADERS=/usr/include/gtest/*.h /usr/include/gtest/internal/*.h
 	GTEST_SRCS=$(GTEST_DIR)/src/*.cc $(GTEST_DIR)/src/*.h $(GTEST_HEADERS)
 	GTEST_OBJ = $(OBJPATH)/gtest-all.o
-	EXTLIBS=-luuid -lxerces-c -lpthread -lmagic
+	EXTLIBS=-luuid -lxerces-c -lpthread
```
dskrvk, line 43: Is magic also mentioned in the setup script?

carlbyington2: I am not familiar with all the build systems used here, but I don't see any setup script.

ronys:
I think that @dskrvk is referring to the Misc/setup-linux-dev-env.sh, which is useful for installing all the packages required to build pwsafe on various distros.

In any case, libmagic is used in os/unix/Media.cpp to implement pws_os::GetMediaType(). This is called when adding an attachment to an entry, to populate the MediaType attachment field.

So it seems to me that the right thing to do is to add support to pwsafe-cli for handling attachments, rather than removing this library.

src/ui/cli/Makefile
carlbyington2: yes, libmagic is referenced in Misc/setup-linux-dev-env.sh which is ok since it is needed by the main pwsafe.

PR #1794, opened 2026-05-25
[wx] Change layout of custom fields dialog

rafaelx:
This PR proposes changes to the layout of the custom fields dialog, in accordance with my review comment in #1746.

The dialog for adding/editing custom fields, as it currently appears:

CustomFieldDlg_master

The changes introduced by this PR to align the layout with other dialogs of this type:

CustomFieldDlg_add

CustomFieldDlg_edit


ronys:
Thanks - it's small consistencies like this that make a difference, even if the user never consciously notices them.

PR #1795, opened 2026-05-26
[wx] Tweaks for the Flatpak version

froggie2523:
This is the first attempt to tackle various aspects of the Flatpak version, and fix issues such as:

Issues:
1. On Linux with Wayland, the virtual keyboard is not detected in the Flatpak version when using the Xwayland backend.
In the native version of the application, virtual keyboard support is detected correctly.
2. Help > About > Check for the latest version fails due to missing permission (network access).
Error: Could not download version data. Could not resolve hostname
This may appear to be an issue with the update infrastructure, but it is actually caused by missing network permissions in the Flatpak sandbox.

Changes:
New: Added detection for whether the Flatpak version is running.
New: Flatpak version is running - check if the XWayland backend is enabled.
New: Flatpak version is running - display network access requirements as a tooltip for the "Check for the latest version" command in Help > About.
New: Flatpak version is running - display a different message about updating the application to the latest version.

Tested on Debian/Kubuntu/Fedora/Mint.

Attached you can see some screenshots.

pwsafe_1 24-wxWidgets-flatpak-Version-01
pwsafe_1 24-wxWidgets-flatpak-Version-02
pwsafe_1 24-wxWidgets-flatpak-Version-03
New:<br>
pwsafe_1 24-wxWidgets-flatpak-Version-04
New:<br>
pwsafe_1 24-wxWidgets-flatpak-Version-05
New:<br>
pwsafe_1 24-wxWidgets-flatpak-xvkbd-01

### DIFF
--- a/src/ui/wxWidgets/wxUtilities.cpp
+++ b/src/ui/wxWidgets/wxUtilities.cpp
@@ -396,7 +396,7 @@ bool IsRunningInFlatpak()
   wxGetEnv("FLATPAK_ID", &appID);
   wxGetEnv("HOME", &homeFolder);
   if (!appID.IsEmpty() && !homeFolder.IsEmpty()) {
-    return (wxDirExists(homeFolder + "/.var/app/" + appID)) ? true : false;
+    return wxDirExists(homeFolder + "/.var/app/" + appID);
   }
   else {
     return false;
--- a/src/ui/wxWidgets/wxUtilities.cpp
+++ b/src/ui/wxWidgets/wxUtilities.cpp
@@ -342,13 +342,16 @@ bool IsXWaylandEnabled()
   wxOperatingSystemId osid = wxGetOsVersion();
 
   if (osid & wxOS_UNIX) { // Includes Linux
-    wxString GDK_BACKEND_VAR = wxEmptyString;
+    wxString GDK_BACKEND_VAR = wxEmptyString; // set by the native version on Linux with Wayland to fall back to using x11 backend
     if (wxGetEnv(wxT("GDK_BACKEND"), &GDK_BACKEND_VAR)) { // provides 'x11' or 'wayland'
       if (!GDK_BACKEND_VAR.IsEmpty()) {
         if (GDK_BACKEND_VAR == wxT("x11")) {
           XWayland = true;
         }
       }
+    }
+    else if (!wxGetEnv(wxT("WAYLAND_DISPLAY"), nullptr)) { // unset by the Flatpak version on Wayland when started with the "--nosocket=wayland --socket=x11" options
+      XWayland = true;
     }
   }
   return XWayland;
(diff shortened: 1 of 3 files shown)


src/ui/wxWidgets/wxUtilities.cpp
```diff
@@ -385,6 +388,24 @@ void wxUtilities::NotifyIfUnsupported(enum Feature feature, wxWindow* window)
...
 
+bool IsRunningInFlatpak()
+{
+#ifdef __WXGTK__
+  wxString appID;
+  wxString homeFolder;
+  wxGetEnv("FLATPAK_ID", &appID);
+  wxGetEnv("HOME", &homeFolder);
+  if (!appID.IsEmpty() && !homeFolder.IsEmpty()) {
+    return (wxDirExists(homeFolder + "/.var/app/" + appID)) ? true : false;
```
ronys: wxDirExists() already return bool, so` ? true : false `is redundant.

nobugshere:
Would it be possible to actually check for a newer Flatpak version, rather than rely on the Linux version number?

ronys:
> Would it be possible to actually check for a newer Flatpak version, rather than rely on the Linux version number?

The 'check for new version' functionality reads https://pwsafe.org/latest.xml and compares the values there to the app's (Windows/wx) version.
latest.xml is build from version.mfc and version.wx as part of the release process.

As flatpack builds are based on the wx releases, but are in a different repo (https://github.com/flathub/org.pwsafe.pwsafe), I don't see a clean way to do this, and am not sure about the benefit as opposed to sticking to version.wx. Any suggestions?

src/ui/wxWidgets/wxUtilities.cpp
```diff
@@ -342,14 +342,17 @@ bool IsXWaylandEnabled()
   wxOperatingSystemId osid = wxGetOsVersion();
 
   if (osid & wxOS_UNIX) { // Includes Linux
-    wxString GDK_BACKEND_VAR = wxEmptyString;
+    wxString GDK_BACKEND_VAR = wxEmptyString; // set by the native version on Linux with Wayland to fall back to using x11 backend
     if (wxGetEnv(wxT("GDK_BACKEND"), &GDK_BACKEND_VAR)) { // provides 'x11' or 'wayland'
       if (!GDK_BACKEND_VAR.IsEmpty()) {
         if (GDK_BACKEND_VAR == wxT("x11")) {
```
dskrvk: Nit: could we combine the two if conditions here?

src/ui/wxWidgets/wxUtilities.cpp
```diff
@@ -385,6 +388,24 @@ void wxUtilities::NotifyIfUnsupported(enum Feature feature, wxWindow* window)
...
 }
 
+bool IsRunningInFlatpak()
+{
+#ifdef __WXGTK__
+  wxString appID;
+  wxString homeFolder;
+  wxGetEnv("FLATPAK_ID", &appID);
+  wxGetEnv("HOME", &homeFolder);
+  if (!appID.IsEmpty() && !homeFolder.IsEmpty()) {
```
dskrvk, line 394: Why need to check the home folder in addition to FLATPAK_ID?

froggie2523: The home-folder check acts as a sanity check to reduce false positives when, for example, FLATPAK_ID environment variable is manually set by a user for testing purposes. The dir is actually created by the process on the file-system when run for the first time.

src/ui/wxWidgets/wxUtilities.cpp
froggie2523: So stupid... Fixed by an additional commit.

src/ui/wxWidgets/wxUtilities.cpp
froggie2523: In the last commit I simplified the nested checks, hope it's what you meant by combining.

PR #1796, opened 2026-05-26
[wx] Add context menu to copy the custom field value(s)

rafaelx:
This PR adds a submenu containing the names of custom fields to the context menu. These submenus allow the user to copy the field values ​​to the clipboard.

Custom Fields of Item
image

Context Menu of Item
image

Refers to my review comment in #1746.

### DIFF
--- a/src/ui/wxWidgets/PasswordSafeFrame.cpp
+++ b/src/ui/wxWidgets/PasswordSafeFrame.cpp
@@ -1892,7 +1892,7 @@ void PasswordSafeFrame::OnContextMenu(const CItemData* item)
     itemEditMenu.Append(ID_COPYURL,        _("Copy UR&L to Clipboard"));
     itemEditMenu.Append(ID_COPYEMAIL,      _("Copy email to Clipboard"));
     itemEditMenu.Append(ID_COPYRUNCOMMAND, _("Copy Run Command to Clipboard"));
-    const auto *item_data = item->IsShortcut() || item->IsAlias() ? m_core.GetBaseEntry(item) : item;
+    const auto *item_data = item->IsShortcut() ? m_core.GetBaseEntry(item) : item;
     if (item_data->IsCustomFieldsSet()) {
       auto customFieldsMenu = new wxMenu;
       auto customFields = item_data->GetCustomFields();
--- a/src/ui/wxWidgets/PasswordSafeFrame.cpp
+++ b/src/ui/wxWidgets/PasswordSafeFrame.cpp
@@ -1896,19 +1896,14 @@ void PasswordSafeFrame::OnContextMenu(const CItemData* item)
       auto customFieldsMenu = new wxMenu;
       auto customFields = item->GetCustomFields();
       for (auto &customField : customFields) {
-        if (customField.HasProperty(CustomField::PropertyId::PROP_NAME)) {
-          auto customFieldName = customField.GetName();
-          auto customFieldValue = customField.GetValue();
-          auto menuID = wxNewId();
-          customFieldsMenu->Append(
-            menuID,
-            wxT("<") + towxstring(customFieldName) + wxT(">")
-          );
+        auto customFieldName = customField.GetName();
+        auto customFieldValue = customField.GetValue();
+        auto menuID = wxNewId();
+        customFieldsMenu->Append(menuID, towxstring(customFieldName));
 
-          Bind(wxEVT_MENU, [&, customFieldValue](wxCommandEvent& WXUNUSED(event)) {
-            DoCopyCustomFieldValue(customFieldValue);
-          }, menuID);
-        }
+        Bind(wxEVT_MENU, [&, customFieldValue](wxCommandEvent& WXUNUSED(event)) {
+          DoCopyCustomFieldValue(customFieldValue);
+        }, menuID);
       }
       itemEditMenu.AppendSubMenu(customFieldsMenu, _("Copy Custom &Field Value..."));
     }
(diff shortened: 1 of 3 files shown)


src/ui/wxWidgets/PasswordSafeFrame.cpp
```diff
@@ -1892,6 +1892,26 @@ void PasswordSafeFrame::OnContextMenu(const CItemData* item)
...
+      auto customFieldsMenu = new wxMenu;
+      auto customFields = item->GetCustomFields();
+      for (auto &customField : customFields) {
+        if (customField.HasProperty(CustomField::PropertyId::PROP_NAME)) {
+          auto customFieldName = customField.GetName();
+          auto customFieldValue = customField.GetValue();
+          auto menuID = wxNewId();
+          customFieldsMenu->Append(
+            menuID,
+            wxT("<") + towxstring(customFieldName) + wxT(">")
```
dskrvk: What’s the rationale for putting <> around the name?

Also, could the name contain any forbidden or special characters?

src/ui/wxWidgets/PasswordSafeFrame.cpp
```diff
@@ -1892,6 +1892,26 @@ void PasswordSafeFrame::OnContextMenu(const CItemData* item)
...
+          );
+
+          Bind(wxEVT_MENU, [&, customFieldValue](wxCommandEvent& WXUNUSED(event)) {
+            DoCopyCustomFieldValue(customFieldValue);
+          }, menuID);
+        }
+      }
+      itemEditMenu.AppendSubMenu(customFieldsMenu, _("Copy Custom &Field Value..."));
```
dskrvk, line 1909: Maybe we should check that the list is not empty before adding the submenu?

src/ui/wxWidgets/PasswordSafeFrame.cpp
rafaelx: > What’s the rationale for putting <> around the name?

Shortcuts are displayed in the context menu as `<GroupName> / <Title> / <Username>`. I have adopted this concept for custom field names, as this format is already in use. However, if the optional group name is missing, a shortcut appears in the context menu as: `<> / <Title> / <Username>`.
Hmm, in this case, the pattern `<GroupName> / <Title> / <Username>` might actually make more sense.
I will remove the characters from the custom field names, since no indicator for missing or non-existent information is required here.

> Also, could the name contain any forbidden or special characters?

I am not aware of any restrictions regarding the characters that may be used, and it appears that any character (UTF-8) is accepted. The field names are displayed in the context menu exactly as they are read from the database. At the UI level, checks are already in place to ensure that a field name cannot consist of an empty string and must be unique. I would therefore like to forward this question to Rony.

@ronys CustomFieldList provides the `getErr()` method, which includes the comment: "Callers should check getErr() before using parsed data." However, throughout the entire source code, I cannot find a single call to `getErr()`.

src/ui/wxWidgets/PasswordSafeFrame.cpp
rafaelx: The submenu is created only if the item actually contains data relevant to custom fields.
The condition is checked on line 1895 via `if (item->IsCustomFieldsSet())`.
As soon as all custom fields for an item have been deleted, this condition is no longer met, since the database field is empty.

@ronys Is my assumption correct?

dskrvk: I mean if the check on 1899 ends up returning false for every entry we’ll get an empty list. Not sure why that check is necessary though.

dskrvk: I suppose the check on 1899 is necessary for correctness whereas the proposed check is a nice to have, and it would be very uncommon that none of the custom fields have a name. So I don’t think it’s worth complicating the code further.

rafaelx: Within `CItemData::AddCustomField`, checks are already performed using `getErr` to detect errors and to verify the presence of the mandatory properties `CustomField::PROP_NAME` and `CustomField::PROP_VALUE` for any custom field. This ensures that only valid and consistent data is stored in the database. Consequently, an additional validation of the data retrieved from the database should no longer be necessary at the UI level; any such validation would serve solely to enhance robustness.
So, if `IsCustomFieldsSet` returns true, then one or more valid custom fields from the database, possessing the properties `PROP_NAME` and `PROP_VALUE` are present, which is guaranteed by the core.
I therefore tend to remove this HasProperty-check and rely on the core and thus on Rony's virtuosity.

dskrvk: Agree, presence of required properties should be an invariant guaranteed by core methods, not the UI layer.

dskrvk (approved):

src/ui/wxWidgets/PasswordSafeFrame.cpp
ronys: `CustomFieldList::getErr()` is called several times within `CItemData`, in the member functions that handle custom fields.

src/ui/wxWidgets/PasswordSafeFrame.cpp
ronys: +1 for "Rony's virtuosity" :-)

Seriously: Custom property names must be present and cannot be empty, custom property values must only be present.
I think that ASSERT() statements are a good way of (1) documenting these assumptions and (2) helping catch them in development time, with no runtime overhead.

ronys:
I've just tested it - very nice!

For consistency, could you add this to shortcut and alias context menus?

rafaelx:
> I've just tested it - very nice!
> 
> For consistency, could you add this to shortcut and alias context menus?

I noticed that custom fields can also be added to an alias. If this is intended, which custom field names should then be displayed in the context menu?

ronys:
For aliases, the custom fields of the alias should be displayed.

An alias only shares the "credentials" with its base entry: Password, auth code / TOTP  and the passkey stuff.

PR #1798, opened 2026-05-27
code cleanups

carlbyington2:
ReadFromDisk does not have to guess unicode flavor, just ignore any utf-8 bom. Much simpler code.

The language files should have a consistent Project-Id. Also, windows has more language files than wxwidgets , but it looks like we cannot just copy those.

Compiled in debug mode, the linux version shows:

Couldn't load default language catalog for Chinese
Couldn't load default language catalog for Korean

Is there something wrong with those two language files?

### DIFF
--- a/src/core/Report.cpp
+++ b/src/core/Report.cpp
@@ -143,7 +143,7 @@ bool CReport::SaveToDisk(const stringT &out_dir)
   }
   else {
     fclose(fd);
-    pws_os::IssueError(_T("SaveToDisk: Conversion error"));
+    pws_os::IssueError(_T("Encoding error while writing report file"));
     return false;
   }
   fclose(fd);
@@ -230,7 +230,7 @@ bool CReport::ReadFromDisk()
   CUTF8Conv conv;
   std::streampos size = ss.tellp();
   if (!conv.FromUTF8(reinterpret_cast<const unsigned char*>(ss.str().c_str()), size, sx)) {
-    pws_os::IssueError(_T("ReadFromDisk: Conversion error"));
+    pws_os::IssueError(_T("Encoding error while reading report file"));
     return false;
   }
 
--- a/src/core/Report.cpp
+++ b/src/core/Report.cpp
@@ -92,51 +92,6 @@ void CReport::StartReport(int iAction, const stringT &csDataBase, bool writeHead
   }
 }
 
-static pugi::xml_encoding guessBufferEncoding(uint8_t d0, uint8_t d1, uint8_t d2, uint8_t d3)
-{
-  if (d0 == 0 && d1 == 0 && d2 == 0xfe && d3 == 0xff) return pugi::encoding_utf32_be;
-  if (d0 == 0xff && d1 == 0xfe && d2 == 0 && d3 == 0) return pugi::encoding_utf32_le;
-  if (d0 == 0xfe && d1 == 0xff) return pugi::encoding_utf16_be;
-  if (d0 == 0xff && d1 == 0xfe) return pugi::encoding_utf16_le;
-  if (d0 == 0xef && d1 == 0xbb && d2 == 0xbf) return pugi::encoding_utf8;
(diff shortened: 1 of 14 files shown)


carlbyington2:
Fedora has wxwidgets 3.2, which has

/usr/share/locale/zh_CN/LC_MESSAGES/wxstd-3.2.mo
/usr/share/locale/ko_KR/LC_MESSAGES/wxstd-3.2.mo

I have a patch for src/ui/wxWidgets/PasswordSafeFrame.cpp to use 
wxLANGUAGE_CHINESE_SIMPLIFIED and wxLANGUAGE_KOREAN_KOREA, combined with moving the pwsafe.mo locale files into zh_CN and ko_KR. That works on Fedora. Do you want those changes in a pull request?

I presume the files in src/ui/wxWidgets/I18N/pos would also need to be renamed,and a fixup in CMakeLists.txt.

src/core/Report.cpp
```diff
@@ -279,7 +229,10 @@ bool CReport::ReadFromDisk()
   StringX sx;
   CUTF8Conv conv;
   std::streampos size = ss.tellp();
-  conv.FromUTF8(reinterpret_cast<const unsigned char*>(ss.str().c_str()), size, sx);
+  if (!conv.FromUTF8(reinterpret_cast<const unsigned char*>(ss.str().c_str()), size, sx)) {
+    pws_os::IssueError(_T("ReadFromDisk: Conversion error"));
```
ronys: 1. I think that "Encoding error while reading report file" would be a bit clearer.
2. Since 'core' is also used by Windows, which doesn't have gettext support for i18n, the text should be added to core.rc2 (with a suitable #define in core.h), and set by stringT::LoadAString()

ronys:
> Fedora has wxwidgets 3.2, which has
> 
> /usr/share/locale/zh_CN/LC_MESSAGES/wxstd-3.2.mo /usr/share/locale/ko_KR/LC_MESSAGES/wxstd-3.2.mo
> 
> I have a patch for src/ui/wxWidgets/PasswordSafeFrame.cpp to use wxLANGUAGE_CHINESE_SIMPLIFIED and wxLANGUAGE_KOREAN_KOREA, combined with moving the pwsafe.mo locale files into zh_CN and ko_KR. That works on Fedora. Do you want those changes in a pull request?
> 
> I presume the files in src/ui/wxWidgets/I18N/pos would also need to be renamed,and a fixup in CMakeLists.txt.

Nice catch!

This can be part of the current PR, or a separate one, as you prefer.

src/core/Report.cpp
carlbyington2: Done. Also changed the error message while writing a report file. But Report.cpp is the only file in core that uses IssueError, and it has these messages:

- SaveToDisk: Finding path to database
- SaveToDisk: Action not filled
- SaveToDisk: Opening log file
- Encoding error while writing report file
- ReadFromDisk: Finding path to database
- ReadFromDisk: Action not filled
- ReadFromDisk: Opening log file
- Encoding error while reading report file
- PurgeFromDisk: Finding path to database
- PurgeFromDisk: Action not filled
- ReportExistsOnDisk: Finding path to database
- ReportExistsOnDisk: Action not filled
- Uexpected report i-d
- GetDatabasePath: Finding path to database

None of those strings are in core.rc2

carlbyington2: None of the implementations of pws_os::IssueError attempt to translate the error string.

ronys: Ugh. We just stepped into a pile of technical debt.
I guess this hasn't come up in the field because these are error conditions in a rarely used feature.
Let's kick this down the road for now.

ronys:
Merging this + a typo fix.

src/core/Report.cpp
carlbyington2: Those messages are seen if compiled in debug mode. In release mode, IssueError is a nop.

PR #1799, opened 2026-05-30
docs: correct Yubico header field type

huven:
This updates the V3 and V4 format documentation for the Yubico `0x12` database header field.

The docs previously listed the field type as `Text`, but the core implementation stores and reads it as a fixed 20-byte binary YubiKey secret key via `HDR_YUBI_SK`. The field is now documented as `20 bytes` and marked as implemented in both format tables.

No code changes.


ronys:
Thanks!

PR #1803, opened 2026-05-31
Treat alias passkey fields as base-owned

huven:
## Summary

This updates alias effective-field resolution so passkey fields are taken from the base entry, matching the existing behavior for passwords, password history, and TOTP fields.

The affected passkey fields are:

- Passkey Credential ID
- Passkey Relying Party ID
- Passkey User Handle
- Passkey Algorithm ID
- Passkey Private Key
- Passkey Sign Count

## Testing

- Added coverage to `AliasShortcutTest` verifying effective passkey field values resolve from the base entry for aliases and shortcuts.
- Ran unit tests on Linux:
  - `cli-test`
  - `Coretests`

### DIFF
--- a/src/test/AliasShortcutTest.cpp
+++ b/src/test/AliasShortcutTest.cpp
@@ -15,12 +15,22 @@
 #include "core/PWSAuxParse.h"
 #include "gtest/gtest.h"
 
+#include <vector>
+
 class AliasShortcutTest : public ::testing::Test
 {
 protected:
   AliasShortcutTest() {}
   PWScore core;
   CItemData base;
+  const std::vector<CItemData::FieldType> passkey_fields = {
+    CItemData::PASSKEY_CRED_ID,
+    CItemData::PASSKEY_RP_ID,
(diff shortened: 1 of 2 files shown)


src/test/AliasShortcutTest.cpp
```diff
@@ -97,6 +109,19 @@ TEST_F(AliasShortcutTest, Alias)
   EXPECT_EQ(effci.GetRunCommand(), al.GetRunCommand());
   EXPECT_EQ(effci.GetCustomFieldsRaw(), al.GetCustomFieldsRaw());
   EXPECT_TRUE(sx_totpauthcode.empty());
+
+  EXPECT_EQ(base.GetEffectiveFieldValue(CItemData::PASSKEY_CRED_ID, nullptr),
```
dskrvk: For maintainability, could we use a loop over a vector of fields?

huven: Yes, changed that.

dskrvk (approved):

PR #1807, opened 2026-06-03
update to pugixml 1.15

carlbyington2:
Update to the latest pugixml version to match Fedora.


carlbyington2:
Hm, the windows build considers the deprecated warning on pugi::xml_document::load to be a fatal error.

We cannot update to pugixml 1.15 until os/mac/file.cpp changes. It looks like it is trying to handle various utf bom flavors in a lock file that will only contain user@machine:nnnnnnnn. Do we need any bom on that file?

nobugshere:
> We cannot update to pugixml 1.15 until os/mac/file.cpp changes. It looks like it is trying to handle various utf bom flavors in a lock file that will only contain user@machine:nnnnnnnn. Do we need any bom on that file?

I don't know the history, so I can't answer the question.  However, after removing that, there are about 60 more errors waiting because you commented-out the PUGIXML_WCHAR_MODE and moved it to the CMakeLists.txt file.  If I add it to the Xcode file (and comment-out the file.cpp code) it will compile.  ~~What was the advantage to treating the PUGIXML_* parameters that way?~~. Nevermind, I see now, you didn't comment them out, you just didn't un-comment them from the original file.  Okay.  Pending guidance on file.cpp and the missing pugi::convertBuffer(), I'll add them to the Mac build.

carlbyington2:
I tried to keep the embedded pugixml identical to the upstream pugixml 1.15 source, and configure it with compiler switches. That should make it easier to update to newer version. I missed doing compiler switches for macOS. 

And I don't know how to make the Windows build accept the deprecated warning.

huven:
> And I don't know how to make the Windows build accept the deprecated warning.

Try replacing `doc.load` with `doc.load_string`.

nobugshere:
Just an observation: Looks like the lock file is written as raw wchar_t (TCHAR) types with no BOM.  On Mac and Linux/Unix they are 32 bit (UTF-32 LE) and Windows it's UTF-16 LE.  The Mac version must have been trying to be compatible with other variations, but the Win and Unix versions don't seem to be so concerned :-)

huven:
> Just an observation: Looks like the lock file is written as raw wchar_t (TCHAR) types with no BOM. On Mac and Linux/Unix they are 32 bit (UTF-32 LE) and Windows it's UTF-16 LE. The Mac version must have been trying to be compatible with other variations, but the Win and Unix versions don't seem to be so concerned :-)

It might be better to always use UTF-8 there.

There is a more serious incompatibility: Windows uses CREATE_ALWAYS and relies on Windows sharing violations, not file existence [src/os/windows/file.cpp (line 276)]. Unix/macOS create the .plk and close it, so Windows can overwrite an existing Unix/macOS .plk instead of treating it as a live lock.

carlbyington2:
Hm, I wanted the "kill executable bits" to be a separate pull request, but it ended up here. Not sure how to fix that.

rathann:
> Hm, I wanted the "kill executable bits" to be a separate pull request, but it ended up here. Not sure how to fix that.

You can `git reset` to a previous commit and then `git push --force` to your fork/branch.

nobugshere:
> It might be better to always use UTF-8 there.

Agreed.

> There is a more serious incompatibility: Windows uses CREATE_ALWAYS and relies on Windows sharing violations, not file existence [src/os/windows/file.cpp (line 276)]. Unix/macOS create the .plk and close it, so Windows can overwrite an existing Unix/macOS .plk instead of treating it as a live lock.

The most likely method of sharing the files is, arguably, via SMB.  My understanding is that O_EXCL|O_CREAT is not guaranteed to work over SMB in any case.  Is there a preferred approach that works via local files, SMB, and NFS?

At this point, I'm not sure what to do.  The Mac version seems unnecessary on current systems; perhaps it was needed during past transitions (e.g. PowerPC to IA-32 to x86_64)?  The current systems are effectively UTF-32LE, in which case convertBuffer() just returns the same buffer, no conversion is necessary.  I expect that would also be the case on 64-bit Linux systems (Intel and ARM, at least).  As @huven pointed out, it may not be working between Windows and any other system anyway.  

My gut reaction is to strip it out of the Mac version for now and open an issue for cross-platform locking: 
- Use UTF-8
- Use a common locking mechanism that works for local files, SMB, and NFS.

Suggestions welcome :-)

carlbyington2:
Google on "cross platform file locking for windows macos linux on file systems that are local or smb or nfs" comes up with the "Atomic Lock" Alternative.

Generate a temporary lock file that includes unique identifiers (e.g., hostname and Process ID).
Attempt to atomically create/move the lock file using an atomic operation (such as os.rename() in Python or link() in Linux).
Read and write to your actual data file.
Delete the lock file to release the lock.

So the existence of x.psafe3.plk means that x.psafe3 is locked. Create host.pid.plk and try to renameat2(,,,RENAME_NOREPLACE). But does that work actually work on smb or nfs? Google says it works on smb but not nfs. But it might work if the .plk is a directory rather than a file.

But that just punts the problem to deleting stale lock files caused by a crash or a forced disconnect from the shared file system.

dskrvk:
FWIW, this is the change that added pugi for Mac lockfile parsing: https://github.com/pwsafe/pwsafe/pull/723/changes/414d69023d5bc399f7724c7a803280e9c85e93cb#diff-28a82f1c88b662f3505bc2800534d17cef0d7c1631e3b42e2af00da16d2849bd

huven:
To complete this PR, I'd align the Mac implementation with the Linux one as @nobugshere suggests and accept the platform incompatibility which the encoding is not going to solve.

And deal with improved locking in a separate PR. Using either compatible locking primitives (like `LockFileEx` on Windows and `fcntl` on Linux/Mac) or the "atomic lock" approach.

carlbyington2:
Since I don't have a Mac to build or test this, perhaps someone else can cherry-pick the two patches for the pugixml update and open a new PR. We can then close this PR.

nobugshere:
> Since I don't have a Mac to build or test this, perhaps someone else can cherry-pick the two patches for the pugixml update and open a new PR. We can then close this PR.

I just submitted a PR to your repo to align the Mac and unix versions.

carlbyington2:
Thank you. Very nice!

ronys:
Thanks to everyone for pitching in on this.

As to .plk:
The .plk file mechanism was designed without cross-platform functionality in mind. 
Writing the raw TCHAR bytes was just a bug that was left as-is in order not to break compatibility.
In any case, improvements to .plk are welcome, but as a separate PR.

PR #1810, opened 2026-06-06
Kill executable src files

carlbyington2:
I hope I did this correctly. This commit removes the executable bit on all the source files. Fedora packaging does not like those bits.


carlbyington2:
Forgot to switch to a commit before branching. Closing this and starting over

PR #1811, opened 2026-06-06
Gorilla pwsafe import

carlbyington2:
Importing files created by gorilla goes thru PWSfileV1V2, even though they claim to be compatible with V3 files. In any case, the URL is in there as CItemData::URL, so we might as well pick it up.


src/core/PWSfileV1V2.cpp
```diff
@@ -491,6 +491,9 @@ int PWSfileV1V2::ReadRecord(CItemData &item)
             case CItemData::RMTIME:
             case CItemData::POLICY:
             case CItemData::XTIME_INT:
+              break;
+            case CItemData::URL:
```
ronys, line 495: Actually, the URL is *not* part of the V2 spec, so I suggest adding a comment to that effect to avoid confusion. Something like "URL added to support import from Gorilla pwsafe"

ronys (changes requested):
Since the pugixml-related changes aren't relevant to this fix, I suggest closing this PR and creating a new one just with the PWSfileV1v2.cpp change (but see comment there first).

carlbyington2:
Yes, I forgot to switch to an appropriate commit before branching. Closing this and starting over

PR #1813, opened 2026-06-06
[SF938] Add "Copy Custom Field Value..." item

ronys:
to entry's popup, allowing copying custom values by name. Shown if/when an entry has custom values.
Corresponds to work done on wx.


src/ui/Windows/resource2.h
```diff
@@ -292,6 +292,9 @@ Do NOT change the value of any resource used in a menu from V3.17 onwards.
 #define ID_MENUITEM_TRAYBROWSEALT1      33501
 #define ID_MENUITEM_TRAYBROWSEALTMAX    33525
 
+#define ID_MENUITEM_COPYCUSTOMFIELD1    33526
+#define ID_MENUITEM_COPYCUSTOMFIELDMAX  33550
```
dskrvk, line 296: Is this defined based on the max supported number of custom fields?

src/ui/Windows/MainMenu.cpp
```diff
@@ -1574,6 +1574,32 @@ void DboxMain::OnContextMenu(CWnd * /* pWnd */, CPoint screen)
       pPopup->RemoveMenu(pci->IsProtected() ? ID_MENUITEM_PROTECT : ID_MENUITEM_UNPROTECT, MF_BYCOMMAND);
     }
 
+    // Build "Copy Custom Field Value..." submenu. Must run before field-based
+    // removals so ID_MENUITEM_COPYRUNCOMMAND is still present as a position anchor.
+    // For shortcuts, inherit custom fields from their base; For aliases, use theirs.
+    m_vCustomFieldValues.clear();
+    {
+      const CItemData *pcf = pci->IsShortcut() ? pbci : pci;
+      const int icustom_pos = app.FindMenuItem(pPopup, ID_MENUITEM_COPYRUNCOMMAND) + 1;
```
dskrvk, line 1582: Can we assert and/or check that this is non-negative?

src/ui/Windows/MainMenu.cpp
```diff
@@ -1574,6 +1574,32 @@ void DboxMain::OnContextMenu(CWnd * /* pWnd */, CPoint screen)
...
+      CMenu *pCustomFieldsPopup = pPopup->GetSubMenu(icustom_pos);
+      const CustomFieldList customFields = pcf->IsCustomFieldsSet()
+                                           ? pcf->GetCustomFields() : CustomFieldList();
+      if (!customFields.empty() && pCustomFieldsPopup != nullptr) {
+        pCustomFieldsPopup->RemoveMenu(0, MF_BYPOSITION);
+        UINT nCFID = ID_MENUITEM_COPYCUSTOMFIELD1;
+        for (const auto &cf : customFields) {
+          if (nCFID > ID_MENUITEM_COPYCUSTOMFIELDMAX)
+            break;
+          m_vCustomFieldValues.push_back(cf.GetValue());
```
dskrvk: I'm not convinced we have to maintain a copy of the custom field data in `m_vCustomFieldValues`. Why not instead save a pointer to the `CustomFieldList` of the currently-selected `CItemData`? That list can be accessed using the same offset.
The benefits are that we avoid additional memory allocation and not run into a situation where the two lists get out of sync.

src/ui/Windows/resource2.h
huven: The spec allows unlimited number of custom fields

src/ui/Windows/MainMenu.cpp
ronys: Good idea - added ASSERT()

src/ui/Windows/MainMenu.cpp
ronys: Since this is used in the context of a right-click on a selected item, we don't even need to save a pointer.
Thanks for pointing (ahem) this out.

dskrvk (approved):

PR #1814, opened 2026-06-06
kill executable source files

carlbyington2:
Kill the executable bit on all source files. Fedora automated review tools don't like those bits.


dskrvk (approved):

CodeLite/exec-wx3-config
dskrvk, line 1: This is also a script

install/deb/postinst
dskrvk, line 1: This and the next file too

dskrvk (approved):

PR #1815, opened 2026-06-06
Gorilla pwsafe import2

carlbyington2:
gorilla passwordsafe import goes thru PWSfileV1V2.cpp, but they use CItemData::URL which is V3.

### DIFF
--- a/src/core/PWSfileV1V2.cpp
+++ b/src/core/PWSfileV1V2.cpp
@@ -493,6 +493,7 @@ int PWSfileV1V2::ReadRecord(CItemData &item)
             case CItemData::XTIME_INT:
               break;
             case CItemData::URL:
+              // not part of V2, but used by gorilla passwordsafe
               item.SetURL(tempdata); break;
             default:
               break;
--- a/src/core/PWSfileV1V2.cpp
+++ b/src/core/PWSfileV1V2.cpp
@@ -491,6 +491,9 @@ int PWSfileV1V2::ReadRecord(CItemData &item)
             case CItemData::RMTIME:
             case CItemData::POLICY:
             case CItemData::XTIME_INT:
+              break;
+            case CItemData::URL:
+              item.SetURL(tempdata); break;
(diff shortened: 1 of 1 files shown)


nobugshere:
Is it worth investigating why the gorilla file is not being recognized as v3?

carlbyington2:
It does not have the 'PWS3' marker as the first four bytes in the file.

dskrvk:
Not sure I understand the fix. Which file format is Gorilla using - v2, v3, or some mix of the two? E.g. if it's using v3 while not setting the header appropriately then the fix should be on the Gorilla's side. If it's using v2 with an extra field (URL) then we shouldn't break the spec and again the fix should be on their side. There is danger in accumulating these app-specific "tweaks" in this codebase.

carlbyington2:
Gorilla can use either v2 or v3. When using v3, it writes the proper PWS3 marker and it imports here nicely. When using v2, it writes a v2 file with an extra URL field. Is there any official spec for storing a URL in a v2 file as is done here, by cramming it into the notes field?

carlbyington2:
Yes, they should have forced V3 output if any entry contained a URL.

dskrvk:
I don't see anything mentioned in the v2 spec related to URL. If Gorilla wants to store it in the Notes field it would have to implement its own ser/de since this behavior is not defined by the official spec (and then no changes would be needed on the pwsafe side).

That said, maybe @ronys knows how similar format extensions have been handled in the past?

ronys:
@dskrvk is correct in that the URL field isn't supported in V2.

However, supporting this field while reading for an app that mistakenly wrote it doesn't break anything, and documenting this in code accordingly, as @carlbyington2 has done, avoids confusing future developers. So I don't see any harm in this PR.

Note that if the user chooses to export to V2 format, the URL will be put in the NOTES field, per ReMergeNotes(), so a round trip isn't idempotent, but I really doubt that this is a problem.

Issue #1818, opened 2026-06-09
Dark mode tweak


ronys:
Background of text control in Message Box needs some tweaking:

<img width="330" height="133" alt="Image" src="https://github.com/user-attachments/assets/fd3761c1-bdd6-41c7-af51-b4b7fb0ef84d" />

Dark mode is fine:

<img width="331" height="134" alt="Image" src="https://github.com/user-attachments/assets/cc6e9a32-9cf3-486e-9a6b-488ddf7e0c5f" />

@martona - can you help here?

martona:
I can try but you'll have to tell me what's wrong with the light-mode messagebox. :)

ronys:
The static text ("Do you want to save...") has a white background around it that's distinct from the light gray background of the dialog box itself.
In dark mode, the text background and the dialog box are the same color.

martona:
You're right - it was hard to see on my screen. 

Regression from the dark-theming child pass resetting the rich edit's custom background in non-dark mode; re-asserted after the pass - fixed. PR submitted.

ronys:
Resolved by #1820

PR #1820, opened 2026-06-10
fix for rich edit control having the incorrect background color in messageboxes

martona:
see: https://github.com/pwsafe/pwsafe/issues/1818


ronys:
That was quick - thanks!

PR #1821, opened 2026-06-10
Darkmode help update

ronys:
- Update all option tab images
- Update Display text with new Display mode and minor corrections
- Update Misc text with minor updaters and corrections.


ronys:
Something wrong here - closing.

PR #1828, opened 2026-06-16
3.72 i18n updates

ronys:


src/ui/Windows/I18N/pos/pwsafe_sv.po
dskrvk, line 1: Could you remove the executable flag?

ronys: Done - thanks!

dskrvk (approved):

PR #1829, opened 2026-06-17
[wx] Reset Statusbar info

froggie2523:
This PR clears stale status bar text when a new database is opened or when the application language is changed.

The main fields that are reset are:
-Double-click Action value
-Last Clipboard Action
-Number of Entries
-Filter indicator

Tested on Debian/macOS/FreeBSD.

Attached you can see some screenshots.

pwsafe_1 24-wxWidgets-bug-StatusBar-01


dskrvk (approved):

ronys:
Thanks!

PR #1830, opened 2026-06-18
Expiry update fix

ronys:
Addresses https://sourceforge.net/p/passwordsafe/bugs/1630/ (Windows), as well as a related bug in wx:

1. Changing an entry's password and setting an expiration date cause the change to the expiration to be lost - Windows only
2. Changing an entry's password reset the expiration to 'never' - Windows and wx.

### DIFF
--- a/docs/ReleaseNotes.md
+++ b/docs/ReleaseNotes.md
@@ -2,7 +2,7 @@ This note describes the new features, fixed bugs and known problems with the lat
 Password Safe, please see the accompanying [**README.md**](../README.md) file. For more information on the product and the project, please visit
 https://pwsafe.org/. Details about changes to older releases may be found in the file ChangeLog.txt.
 
-In the following, SFxxxx refers to Bug Reports, Feature Requests and Service Requests in PasswordSafe SourceForge Project tickets, and GHxxxx refers to issues in the PasswordSafe GitHub project.
+In the following, SFxxxx refers to Bug Reports, Feature Requests and Support Requests in PasswordSafe SourceForge Project tickets, and GHxxxx refers to issues in the PasswordSafe GitHub project.
 
 PasswordSafe 3.72.0pre Release June 12 2026
 ===========================================
@@ -10,7 +10,8 @@ Bugs Fixed in 3.72.0pre
 -----------------------
 * Improved font scale handling -should resolve font size issues on high resolution displays.
 * [GH1749](https://github.com/pwsafe/pwsafe/issues/1749) In the Master Password Setup window, "Show Master Password" is no longer truncated on some displays.
-* [GH1092](https://github.com/pwsafe/pwsafe/issues/1092)[SF1595](https://sourceforge.net/p/passwordsafe/bugs/1595/) Size and position of main window is now correctly restored on scaled displays.
+* [GH1092](https://github.com/pwsafe/pwsafe/issues/1092), [SF1595](https://sourceforge.net/p/passwordsafe/bugs/1595/) Size and position of main window is now correctly restored on scaled displays.
+* [SF1630](https://sourceforge.net/p/passwordsafe/bugs/1630/) Keep password expiry date when both password and password expiry are changed.
 * [SF1628](https://sourceforge.net/p/passwordsafe/bugs/1628/) Custom values can now be copied to the clipboard in read-only mode via Ctrl-C and right-click->Copy Value.
 
 New Features in 3.72.0pre
--- a/docs/ReleaseNotes.md
+++ b/docs/ReleaseNotes.md
@@ -11,7 +11,7 @@ Bugs Fixed in 3.72.0pre
 * Improved font scale handling -should resolve font size issues on high resolution displays.
 * [GH1749](https://github.com/pwsafe/pwsafe/issues/1749) In the Master Password Setup window, "Show Master Password" is no longer truncated on some displays.
 * [GH1092](https://github.com/pwsafe/pwsafe/issues/1092), [SF1595](https://sourceforge.net/p/passwordsafe/bugs/1595/) Size and position of main window is now correctly restored on scaled displays.
-* [SF1630](https://sourceforge.net/p/passwordsafe/bugs/1630/) Keep password expiry date when both password and password expiry are changed.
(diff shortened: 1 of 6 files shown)


src/ui/wxWidgets/AddEditPropSheetDlg.cpp
```diff
@@ -3175,6 +3175,14 @@ Command* AddEditPropSheetDlg::NewEditEntryCommand()
     m_Item.SetXTimeInt(0);
   }
 
+  // If only the password changed, UpdatePassword() may have cleared a
+  // non-recurring expiry date - restore what the user still sees in the UI.
+  if ((changes & Changes::Password) &&
+      !(changes & (Changes::XTime | Changes::XTimeInt | Changes::XTimeNever)) &&
+    m_DatesTimesExpireOnCtrl->GetValue()) {
+    m_Item.SetXTime(NormalizeExpDate(m_DatesTimesExpiryDateCtrl->GetValue()).GetTicks());
+  }
```
nobugshere: Could we just remove the call to SetXTime(0) from UpdatePassword(), instead of fixing it up here?

docs/ReleaseNotes.md
```diff
@@ -2,15 +2,16 @@ This note describes the new features, fixed bugs and known problems with the lat
 Password Safe, please see the accompanying [**README.md**](../README.md) file. For more information on the product and the project, please visit
 https://pwsafe.org/. Details about changes to older releases may be found in the file ChangeLog.txt.
 
-In the following, SFxxxx refers to Bug Reports, Feature Requests and Service Requests in PasswordSafe SourceForge Project tickets, and GHxxxx refers to issues in the PasswordSafe GitHub project.
+In the following, SFxxxx refers to Bug Reports, Feature Requests and Support Requests in PasswordSafe SourceForge Project tickets, and GHxxxx refers to issues in the PasswordSafe GitHub project.
```
nobugshere, line 5: Need the same change in ReleaseNotesWX.md

docs/ReleaseNotes.md
```diff
@@ -2,15 +2,16 @@ This note describes the new features, fixed bugs and known problems with the lat
...
 
 PasswordSafe 3.72.0pre Release June 12 2026
 ===========================================
 Bugs Fixed in 3.72.0pre
 -----------------------
 * Improved font scale handling -should resolve font size issues on high resolution displays.
 * [GH1749](https://github.com/pwsafe/pwsafe/issues/1749) In the Master Password Setup window, "Show Master Password" is no longer truncated on some displays.
-* [GH1092](https://github.com/pwsafe/pwsafe/issues/1092)[SF1595](https://sourceforge.net/p/passwordsafe/bugs/1595/) Size and position of main window is now correctly restored on scaled displays.
+* [GH1092](https://github.com/pwsafe/pwsafe/issues/1092), [SF1595](https://sourceforge.net/p/passwordsafe/bugs/1595/) Size and position of main window is now correctly restored on scaled displays.
+* [SF1630](https://sourceforge.net/p/passwordsafe/bugs/1630/) Keep password expiry date when both password and password expiry are changed.
```
nobugshere: Should add 1630 to ReleaseNotesWX.md.

src/ui/wxWidgets/AddEditPropSheetDlg.cpp
```diff
@@ -3175,6 +3175,14 @@ Command* AddEditPropSheetDlg::NewEditEntryCommand()
     m_Item.SetXTimeInt(0);
   }
 
+  // If only the password changed, UpdatePassword() may have cleared a
+  // non-recurring expiry date - restore what the user still sees in the UI.
```
dskrvk: If I understand the issue correctly, `UpdatePassword` sets the exp time based on the item's own `XTIME_INT` property, but in this case the property isn't set (yet?), so `UpdatePassword` uses the default of 0. Why is it not set in this case? Can the make it so the item is aware of the desired expiry time by the time `UpdatePassword` is called?

src/ui/wxWidgets/AddEditPropSheetDlg.cpp
```diff
@@ -3175,6 +3175,14 @@ Command* AddEditPropSheetDlg::NewEditEntryCommand()
...
+  if ((changes & Changes::Password) &&
```
dskrvk: May not apply depending on the other comment's outcome, but it would be more logical to put this if block after the `XTimeInt` and `XTimeNever` if blocks below.

src/ui/wxWidgets/AddEditPropSheetDlg.cpp
```diff
@@ -3175,6 +3175,14 @@ Command* AddEditPropSheetDlg::NewEditEntryCommand()
...
+      !(changes & (Changes::XTime | Changes::XTimeInt | Changes::XTimeNever)) &&
+    m_DatesTimesExpireOnCtrl->GetValue()) {
```
dskrvk: Nit: would it make sense to create a local var with `m_DatesTimesExpireOnCtrl->GetValue()`, since this is accessed in multiple places in this method? The control's value shouldn't be modified here, so it should be safe.

src/ui/wxWidgets/AddEditPropSheetDlg.cpp
nobugshere: It looks to me like the logic in UpdatePassword is:
```
if entry has an expiration interval
    update the expiration date
else if it has a fixed expiration date
    invalidate it (Setting the time to 0 effectively means never)
```

If that is not the desired behavior (i.e. the fixed date should be preserved), then I think setting it to 0 in UpdatePassword is incorrect.

If the user changes the PW and sets a new interval at the same time, the code should work as expected because the changes to the expiration controls will be detected.

src/test/ItemDataTest.cpp
```diff
@@ -282,6 +282,73 @@ TEST_F(ItemDataTest, PasswordHistory)
...
+    di.SetCTime();
+    di.SetPassword(L"first");
+    di.SetXTime(time_t(0));
+    di.SetXTimeInt(0);
+
+    di.UpdatePassword(L"second");
+
+    time_t xt;
+    di.GetXTime(xt);
+    EXPECT_EQ(time_t(0), xt);
```
dskrvk, line 324: Could we also verify the interval for good measure?

dskrvk (approved):

PR #1831, opened 2026-06-19
[wx] Show entry count in group tooltips

froggie2523:
This PR makes the wx version behave the same way as the Windows version with regard to entry/shortcut and group tooltips:
Shows Notes for shortcuts (if enabled in Options).
Shows the number of entries in a group tootip.

Note:
Translations (PO files) for the group tooltip were taken from the Windows PO files.

This PR also includes additional updates to the Slovak PO file:
-updated translations to align with recent changes to the Export to TXT/XML feature, custom fields and Flatpak tweaks

Tested on Debian(GNOME)/Mint(Cinnamon)/FreeBSD(KDE)/macOS.

pwsafe_1 24-wxWidgets-feat-group-tooltip-01
pwsafe_1 24-wxWidgets-feat-group-tooltip-02
pwsafe_1 24-wxWidgets-feat-group-tooltip-03

### DIFF
--- a/src/ui/wxWidgets/TreeCtrl.cpp
+++ b/src/ui/wxWidgets/TreeCtrl.cpp
@@ -990,7 +990,7 @@ void TreeCtrlBase::SelectItem(const CUUID & uuid)
   }
 }
 
-size_t TreeCtrl::GetEntriesCount(const wxTreeItemId& item)
+size_t TreeCtrl::GetEntriesCount(const wxTreeItemId& item) const
 {
   size_t count = 0;
   auto *itemData = dynamic_cast<PWTreeItemData *>(GetItemData(item));
@@ -1017,9 +1017,9 @@ void TreeCtrl::OnGetToolTip( wxTreeEvent& evt )
     }
   }
   else if (ItemIsGroup(id)) { // we're on a group other than root, show number of entries including subgroups
-    int count = GetEntriesCount(id);
+    const size_t count = GetEntriesCount(id);
     if (count > 0) {
-      const wxString entries = (count > 1) ? wxString::Format(_("%d entries under group"), count) : wxString::Format(_("%d entry under group"), count);
+      const wxString entries = wxString::Format(_("Number of entries: %d"), count);
       evt.SetToolTip(entries);
     }
   }
--- a/src/ui/wxWidgets/TreeCtrl.cpp
+++ b/src/ui/wxWidgets/TreeCtrl.cpp
@@ -990,14 +990,37 @@ void TreeCtrlBase::SelectItem(const CUUID & uuid)
   }
 }
 
+size_t TreeCtrl::GetEntriesCount(const wxTreeItemId& item)
+{
+  size_t count = 0;
+  auto *itemData = dynamic_cast<PWTreeItemData *>(GetItemData(item));
+  if (itemData && GetChildrenCount(item, false) == 0)
+    ++count;
+  wxTreeItemIdValue cookie;
+  wxTreeItemId child = GetFirstChild(item, cookie);
+  while (child.IsOk()) {
+    count += GetEntriesCount(child);
+    child = GetNextChild(item, cookie);
+  }
+  return count;
+}
+
 void TreeCtrl::OnGetToolTip( wxTreeEvent& evt )
 { // Added manually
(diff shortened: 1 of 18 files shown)


src/ui/wxWidgets/TreeCtrl.cpp
```diff
@@ -990,16 +990,39 @@ void TreeCtrlBase::SelectItem(const CUUID & uuid)
...
+    if (PWSprefs::GetInstance()->GetPref(PWSprefs::ShowNotesAsTooltipsInViews)) {
+      const CItemData *pci = ci->IsShortcut() ? m_core.GetBaseEntry(ci) : ci;
+      const wxString note = pci->GetNotes().c_str();
       evt.SetToolTip(note);
     }
   }
+  else if (ItemIsGroup(id)) { // we're on a group other than root, show number of entries including subgroups
+    int count = GetEntriesCount(id);
+    if (count > 0) {
+      const wxString entries = (count > 1) ? wxString::Format(_("%d entries under group"), count) : wxString::Format(_("%d entry under group"), count);
```
dskrvk: This logic doesn't  work for counts ending in 1 (21, 31, ...) in English or for languages with more than 2 forms of plurals (e.g. Arabic). IIRC Android's i18n libraries support this natively, but there's no simple way to do it here. An existing pattern seems to be to use a generic string (e.g. "entry(s)").

src/ui/wxWidgets/TreeCtrl.cpp
```diff
@@ -990,16 +990,39 @@ void TreeCtrlBase::SelectItem(const CUUID & uuid)
...
+  const CItemData *ci = GetItem(id);
+  if (ci != nullptr) {
+    if (PWSprefs::GetInstance()->GetPref(PWSprefs::ShowNotesAsTooltipsInViews)) {
+      const CItemData *pci = ci->IsShortcut() ? m_core.GetBaseEntry(ci) : ci;
+      const wxString note = pci->GetNotes().c_str();
       evt.SetToolTip(note);
     }
   }
+  else if (ItemIsGroup(id)) { // we're on a group other than root, show number of entries including subgroups
+    int count = GetEntriesCount(id);
```
dskrvk: Why declare with a `size_t` return type only to cast to int here?

src/ui/wxWidgets/TreeCtrl.cpp
```diff
@@ -990,16 +990,39 @@ void TreeCtrlBase::SelectItem(const CUUID & uuid)
   }
 }
 
+size_t TreeCtrl::GetEntriesCount(const wxTreeItemId& item)
```
dskrvk: Can we have some tests for this?

dskrvk (changes requested):

src/ui/wxWidgets/TreeCtrl.h
```diff
@@ -223,6 +223,8 @@ class TreeCtrl : public TreeCtrlBase, public Observer
   /// wxEVT_MOTION event handler for mouse events
   void OnMouseMove(wxMouseEvent& event);
 
+  size_t GetEntriesCount(const wxTreeItemId& item); // Get count of entries in a group, including subgroups
```
ronys: Conceptually, at least, this should be a const member function:
`size_t GetEntriesCount(const wxTreeItemId& item) const;`
`--------------------------------------------------^^^^`

src/ui/wxWidgets/TreeCtrl.cpp
froggie2523: Yes, my mother tongue also has more than one plural forms.
I've changed the format to a single string: "Number of entries: %d". A similar translation already exists in the PO files.

src/ui/wxWidgets/TreeCtrl.cpp
froggie2523: Do you mean the unit tests in the source tree under the "test" directory? As I understand it, they are intended to test core functionality and are therefore frontend-agnostic. This PR focuses on enhancing the user experience in the wxWidgets version.

I have performed the following manual tests though:
-Creating entries and shortcuts, and changing their groups via the Add/Edit dialog
-Creating empty groups
-Moving entries and groups to other groups or to the root using the mouse
-Performing Undo and Redo operations
-Importing entries from TXT and XML files
-Importing entries and groups via drag-and-drop from another Password Safe instance

PR #1837, opened 2026-06-27
[wx] Clear stale Tree tooltips when closing a database

froggie2523:
This PR clears stale Tree tooltips when closing a database.

Steps to reproduce:
-Hover the mouse pointer over a group or an entry with Notes (When "Show Notes as Tooltips" is checked in Options).
-Close the database using the main menu, the toolbar, or by pressing Ctrl+W.
-Hover the mouse pointer over the empty area of the Tree. The previously displayed entry or group tooltip is still shown.

Note: With the introduction of group tooltips in PR #1831, this issue may occur more frequently than before.

Tested on Debian/Mint/FreeBSD/macOS.

pwsafe_1 24-wxWidgets-bug-tree-tooltip-01
pwsafe_1 24-wxWidgets-bug-tree-tooltip-02

### DIFF
--- a/src/ui/wxWidgets/PasswordSafeFrame.cpp
+++ b/src/ui/wxWidgets/PasswordSafeFrame.cpp
@@ -3397,6 +3397,8 @@ void PasswordSafeFrame::CloseDB(std::function<void(bool)> callback)
           m_core.SetReadOnly(false);
           UpdateStatusBar();
           UpdateMenuBar();
+          if (IsTreeView())
+            m_tree->SetToolTip(nullptr); // Clear stale entry (Notes) or group (number of entries) tooltip
           PWSprefs::GetInstance()->SetPref(PWSprefs::FindToolBarActive, showSearchBar);
         }
         else {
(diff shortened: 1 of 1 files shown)


froggie2523:
Build pwsafe with CMake on macOS check failed.
step:
Run sudo xcode-select -s /Applications/Xcode_${XCODE_VER}.app
error:
xcode-select: error: invalid developer directory '/Applications/Xcode_16.4.app'

Most probably due to https://github.com/actions/runner-images/issues/14167:
macos-26 is ready to become the default version for the macos-latest label in GitHub Actions

Software difference between macOS-15 and macOS-26
Default Xcode switches from 16.4 to 26.4.1.

nobugshere:
> Software difference between macOS-15 and macOS-26

I've been meaning to submit a PR for that.  It shouldn't block this PR.

src/ui/wxWidgets/PasswordSafeFrame.cpp
```diff
@@ -3397,6 +3397,8 @@ void PasswordSafeFrame::CloseDB(std::function<void(bool)> callback)
           m_core.SetReadOnly(false);
           UpdateStatusBar();
           UpdateMenuBar();
+          if (IsTreeView())
+            m_tree->SetToolTip(nullptr); // Clear stale entry (Notes) or group (number of entries) tooltip
           PWSprefs::GetInstance()->SetPref(PWSprefs::FindToolBarActive, showSearchBar);
```
dskrvk: Nit: it seems this line should go right after 3396 (since the subsequent lines don't deal with the search bar). This PR adds more lines in between, increasing the likelihood of future mistakes (due to loss of context between getting and setting the preference).

dskrvk (approved):

dskrvk (approved):

PR #1838, opened 2026-06-27
Bump Xcode version in macos-cmake-latest.yml

nobugshere:
GitHub has upgraded macos-latest to 26.x, which does not have Xcode 16.  The current version Xcode is 26.5.

Note that the macos-latest.yml workflow is explicitly using macOS 15 and Xcode 16.  This PR does not change that.


nobugshere:
Well that's lame.  The workflow specifies macos-latest, but it still got 15.7.7 instead of 26.  @ronys , maybe the roll-out isn't complete yet.  Maybe try re-running the failed action in a couple of days.  Roll the dice and see what OS you get :-)

ronys:
The rollout is planned to be completed by July 15.

I've also had sporadic failures of this job. I attribute it to GitHub's general flakiness lately.

PR #1847, opened 2026-07-03
[wx] System Tray MRU not working properly

froggie2523:
This PR fixes several issues with the Entries MRU List in System Tray in the wx version.

Issues:
1. The "Remember last X used entries" setting in Options > System is not honored due to an incorrect reference. It always returns 10, and the value cannot be changed.
2. The feature cannot be disabled by setting the value to 0.
3. The List is stored in the database file but is not loaded when the database is opened.

Tested on Debian/Mint/macOS.

pwsafe_1 24-wxWidgets-bug-SysTrayMRU-01

### DIFF
--- a/src/ui/wxWidgets/PasswordSafeFrame.cpp
+++ b/src/ui/wxWidgets/PasswordSafeFrame.cpp
@@ -461,7 +461,7 @@ void PasswordSafeFrame::Init()
     SetTreeSortType(TreeSortType::GROUP);
   }
 
-  m_RUEList.SetMax(PWSprefs::GetInstance()->PWSprefs::MaxREItems);
+  m_RUEList.SetMax(PWSprefs::GetInstance()->GetPref(PWSprefs::MaxREItems));
 ////@begin PasswordSafeFrame member initialisation
   m_Toolbar = nullptr;
   m_Dragbar = nullptr;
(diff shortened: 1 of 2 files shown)


nobugshere:
Testing on macOS Sequoia, I noticed a couple of problems:
1. An updated list is not actually saved to the DB unless some other change is made.  That is, changes only to the RUE list are not detected as a change to the DB file, so it's not saved.  But, do we really want to update the DB every time an entry is used?  (Assuming access times are not recorded.) I don't think I would like that.
2. When the list is re-loaded, it seems to be in reverse order.  When at the limit, a new entry replaces the most recent instead of the oldest.

froggie2523:
I've noticed the same behavior during testing. Yes, the actual List manipulation deserves further review; I've observed similar quirks on Windows as well.
However, my primary goal is to restore the basic functionality on Linux/macOS so that it is on par with the Windows version:
-Allow the feature to be disabled
-Allow the list size to be configured
-Load the list when the database is opened

src/ui/wxWidgets/PasswordSafeFrame.cpp
```diff
@@ -1277,6 +1277,7 @@ int PasswordSafeFrame::Load(const StringX &passwd)
     SetTitle(m_core.GetCurFile().c_str());
     m_sysTray->SetTrayStatus(SystemTray::TrayStatus::UNLOCKED);
     m_core.ResumeOnDBNotification();
+    m_RUEList.SetRUEList(m_core.GetRUEList()); // It is populated in PWScore::ReadFile() from the file header
```
dskrvk: Not: this kind of comment doesn’t provide any information that’s not already contained in the code and can get outdated when the code changes. It’s easy enough to trace the usages of the property.

ronys:
Nice catch. Thanks!

Issue #1530, opened 2025-07-01
[wx33x] Rework related to the use of _()


nobugshere:
Related to #1516, #1520, #1529, and [this](https://wxwidgets.org/help/msgid-literals/).
Just for reference, this is the list of files I worked from when I attempted to remove _():
```
pwsafe/src % grep '_([^"L].*)' ui/wxWidgets/*.{cpp,h} core/*.{cpp,h} os/{mac,unix}/*.{cpp,h} os/*.h
ui/wxWidgets/DragBarCtrl.cpp:      wxString::Format(_("Drag this image onto another window to paste the '%s' field."), _(name));
ui/wxWidgets/MenuViewHandlers.cpp:    wxMessageBox(_(tcAction) + _(L" file \'") + rpt.GetFileName() + _(L"' not readable"), _("View Report"), wxOK|wxICON_ERROR);
ui/wxWidgets/OptionsPropertySheetDlg.cpp:    Backup_SuffixCBStrings.Add(_(BACKUP_SUFFIX[i]));
ui/wxWidgets/OptionsPropertySheetDlg.cpp:    wxString tmp = _(DCAStrings[I]);
ui/wxWidgets/OptionsPropertySheetDlg.cpp:  m_Misc_DoubleClickActionCB->SetValue(_(DCAStrings[m_DoubleClickAction]));
ui/wxWidgets/OptionsPropertySheetDlg.cpp:  m_Misc_ShiftDoubleClickActionCB->SetValue(_(DCAStrings[m_ShiftDoubleClickAction]));
ui/wxWidgets/OptionsPropertySheetDlg.cpp:    if (str == _(DCAStrings[i])) {
ui/wxWidgets/PasswordSafeFrame.cpp:    wxString lang_name = _(get<1>(item.second));
ui/wxWidgets/PasswordSafeFrame.cpp:    m_statusBar->SetStatusText(_(PWSprefs::GetDCAdescription(-1)), StatusBar::Field::DOUBLECLICK);
ui/wxWidgets/PasswordSafeFrame.cpp:  m_statusBar->SetStatusText(_(PWSprefs::GetDCAdescription(dca)), StatusBar::Field::DOUBLECLICK);
ui/wxWidgets/PWFiltersEditor.cpp:  return _(iter->second);
ui/wxWidgets/PWFiltersEditor.cpp:      m_bestSize.IncTo(DoGetBestSize(attr, dc, _(value)));
[...]

PR #1850, opened 2026-07-04
Fix build/run issues on macOS with Xcode 26.6

sauravg:


froggie2523:
Why all the changes that disable translations?

Xcode/pwsafe-xcode6.xcodeproj/project.pbxproj
```diff
@@ -2536,6 +2536,7 @@
 					"$(inherited)",
 					"/usr/local/include/ykpers-1",
 				);
+				LIBRARY_SEARCH_PATHS = /opt/homebrew/Cellar/ykpers/1.20.0_2/lib/;
```
dskrvk: Should we be hard-coding the version like this?

Xcode/pwsafe-cli.entitlements
```diff
@@ -1,5 +1,8 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
-<dict/>
+<dict>
+<key>com.apple.security.cs.disable-library-validation</key>
```
dskrvk: Why?

nobugshere:
> Why all the changes that disable translations?

It looks like they are trying to build for wxWidgets 3.3.x.  See #1516, #1520, #1530 and [this](https://wxwidgets.org/help/msgid-literals/).  This is a project that we need to undertake sometime soon, but it's going to need more testing on Mac and Linux.

The other part of this PR changes to the Xcode project file to use a version of the ykpers library installed via Homebrew.  That's all good for their own use, but I don't think it's appropriate for our release builds.
1. A user that downloads pwsafe would need to know how to install the dependency.
2. Homebrew only installed the Arm version, so they disabled universal binary.
3. ykpers is technically EOL, so Homebrew has marked it for removal by the end of the year.

We get around this by building from source and static linking all but Apple provided libs.

The title is misleading since this has nothing to do with Xcode 26.

sauravg:
> Why all the changes that disable translations?

The `_(...)` macro seems to be meant for string literals only. There's static assert in wx3.3 against other usages. Those `_(x)` where I removed the translation were passing in something like StringX or wxString.

Xcode/pwsafe-xcode6.xcodeproj/project.pbxproj
sauravg: > Should we be hard-coding the version like this?

Probably not.  But the only other way I can think of is to put it in the xcconfig files, which other developers would still need to hand-edit on their machines.

And it doesn't solve the bigger issue anyway: just what are the stadard location & version that we should be using?

Xcode/pwsafe-cli.entitlements
sauravg: Built with Xcode 26.6, macOS refuses to load the wx dylibs into pwsafe's process space since they are not signed by the same Team-ID.

I tried to disable library validation in runtime hardening checks for pwsafe. Didn't work for me.

sauravg:
> The other part of this PR changes to the Xcode project file to use a version of the ykpers library installed via Homebrew. That's all good for their own use, but I don't think it's appropriate for our release builds.
> 
That's correct. But the same issue applies to the use of wxWidgets also, which is now hardcoded in pwsafe project file.

I have not tried building yk* libraries on macOS, but I have built wxWidgets many times since 2.8, and it's no fun. It kind of makes sense that most people, for their personal use, would just use homebrew to install the dependencies, including wxWidgets. In which case there's no control over the versions of those libraries, and the path remains the same for all users anyway (except intel vs apple silicon version of homebrew). Which is why I didn't put in any effort to not "hardcode" the version or location of dependencies.

For wxWidgets, there's at least a script in the `Xcode` folder to generate the xcconfig files given the location of wx-config script that comes with wxWidgets. I don't know how to do that for custom-built yk* libraries.

> 1. A user that downloads pwsafe would need to know how to install the dependency.
> 2. Homebrew only installed the Arm version, so they disabled universal binary.
> 3. ykpers is technically EOL, so Homebrew has marked it for removal by the end of the year.
> 
> We get around this by building from source and static linking all but Apple provided libs.
> 
How would you specify the location of the libraries you build?

> The title is misleading since this has nothing to do with Xcode 26.

Except the Team-ID mismatch issue. It has been a while since I last built pwsafe with Xcode 16, but I don't remember that one ever.

nobugshere:
> How would you specify the location of the libraries you build?

The GitHub workflow builds the yk* products then does a "make install" to put them in /usr/local/lib, the Xcode project is configured to find them there.  The Homebrew install should have symlinks for all the libs and headers in /opt/homebrew/{lib,include}, so you could search for them there without having to know the version numbers.

Rather than changing the Release configuration, may I suggest adding a more generalized "use Homebrew" scheme and configuration to the Xcode project?  Call it "Homebrew" or "Homebrew-debug".  You can set-up that configuration to search /opt/homebrew for libs and headers. You can also add the entitlement to the specific configuration. 

I also respectfully suggest that this PR should be split in two since the wx3.3.x part does not depend on the Xcode project related changes.

Xcode/pwsafe-xcode6.xcodeproj/project.pbxproj
dskrvk: I like the suggestions by @nobugshere below. If we have to specify a version here, maybe inject it via an env variable?

sauravg:
> Rather than changing the Release configuration, may I suggest adding a more generalized "use Homebrew" scheme and configuration to the Xcode project? Call it "Homebrew" or "Homebrew-debug". You can set-up that configuration to search /opt/homebrew for libs and headers. You can also add the entitlement to the specific configuration.
> 
Done. This PR now only contains the changes required to build & run with Xcode 26, using homebrew installed dependencies. I introduced a new scheme called "homebrew". All my changes are contained in that scheme. There are separate entitlement & xcconfig files for that scheme.

I also noticed that homebrew creates symlinks in /opt/homebrew/opt, so I referenced Cellar links with those. That eliminates explicit version references.


> I also respectfully suggest that this PR should be split in two since the wx3.3.x part does not depend on the Xcode project related changes.

I moved the wx3.3 part (translation related fixes) to https://github.com/pwsafe/pwsafe/pull/1852

nobugshere:
@ronys 
Dude! I was in the middle of reviewing it :-)
Oh well, so far I've only found a minor issue...I'll submit a PR later.

ronys:
@nobugshere 
Sorry, I thought there were no more open issues around this PR.

nobugshere:
@sauravg ,
I like what you did here.  Thanks for working on it.

sauravg:
@nobugshere thanks for your inputs. It turned out much better than my initial hasty PR

Did you file a PR? Curious what the minor issue was with this one.

nobugshere:
It looks like it removed some of the "Build active architecture only" settings.  (ONLY_ACTIVE_ARCH items in the file.)  I just put them back and made a few more edits to the markdown doc.  Plus the editor I used striped off a bunch of trailing whitespace....

PR #1851, opened 2026-07-04
Fix Linux compiler warnings & treat any future ones as errors

dskrvk:
There have been cases before when a missed compiler warning caused a bug, so this enforcement should help reduce such instances and also improve the code quality a bit. We should do this for other platforms too, but I can't verify those builds work locally. Will try and see if the Github builds succeed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -218,7 +218,13 @@ endif (WX_WINDOWS)
 include(set-options-msvc)
 else ()
 set(CMAKE_CXX_FLAGS
-  "${CMAKE_CXX_FLAGS} -Wall -DUNICODE -DWCHAR_INCOMPATIBLE_XMLCH ${CMAKE_WXWINDOWS_CXX_FLAGS}")
+    "${CMAKE_CXX_FLAGS} -Wall -DUNICODE -DWCHAR_INCOMPATIBLE_XMLCH ${CMAKE_WXWINDOWS_CXX_FLAGS}")
+  if (NOT APPLE)
+    # TODO: with the below flag the Mac cmake-based build fails with an error:
+    #   clang: error: argument unused during compilation: '-fstack-clash-protection'
+    # it's possible check_cxx_compiler_flag is using the wrong compiler for the check?
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Werror")
+  endif (NOT APPLE)
 set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -D_DEBUG -DDEBUG")
 if (USE_ASAN)
   set(CMAKE_CXX_FLAGS_DEBUG
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -41,7 +41,7 @@ if (UNIX)
 endif ()
 
 if (NOT WIN32)
-include(FindUnixCommands)
+  include(FindUnixCommands)
 endif (NOT WIN32)
 
 list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/CMakeModules")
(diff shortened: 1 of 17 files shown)


dskrvk:
Looks like wx 3.2 already defines enum combination operators. Let me look into this some more.

nobugshere:
Hi @dskrvk ,
I've been following along here and building this branch on my Mac.  (But not much run-time testing yet.)  My process is close to that used by the macos-latest.yml workflow, which is the one that counts :-)

The cmake version isn't the same.  It looks like it's warning about an option that Apple clang doesn't support and is not offered in the Xcode IDE, it's not even mentioned in the clang manual, so it fails because of -Werror.  I don't know why cmake is adding it, but it's nice of clang to say it's not using it.
`clang: warning: argument unused during compilation: '-fstack-clash-protection' [-Wunused-command-line-argument]`

Is your goal to be able to build with a specific set of options e.g. -Wall -Werror, but not necessarily -Wextra and -pedantic ?

With -Wall I get a number of "hides overloaded virtual function" warnings that I don't see when I build this PR on Fedora 44 (gcc 16).  I'll be happy to fix these and add -Wall -Werror after this is merged.

dskrvk:
Hi @nobugshere,
Is the plan to eventually use only the macos-latest (xcodebuild-based) workflow, which works fine? If so, that would be a relief, since I also can't figure out why cmake (or something else) is adding the unrecognized flag (see below). My goal is just `-Wall -Werror` for now.

Here's where I am right now:
1. The flag is added on https://github.com/pwsafe/pwsafe/blob/1e101957824d08e82d2e453a909bd431c703db84/CMakeModules/harden/harden-gcc.cmake#L46, which goes through a check on https://github.com/pwsafe/pwsafe/blob/1e101957824d08e82d2e453a909bd431c703db84/CMakeModules/create-harden-interface.cmake#L29
2. The check seems to work as expected, since the build log prints `Performing Test HARDEN_COMPILE_STACK_CLASH_PROTECTION - Success`, whereas for features that are reported as Failed there are no flags added. Perhaps the check is performed against gcc instead of clang?
3. What's also puzzling is that the log reports compilation for only 4 of the files, 3 of which (strutils.cpp, split-test.cpp, searchaction.cpp) fail because of the unrecognized flag. Either the log is incomplete or there's some caching going on, otherwise why isn't the 4th file reporting the same error, and where are the others?
4. I'm thinking of not enabling `-Werror` on Mac for now and letting you (or someone else with access to a Mac) continue this in a separate PR, does that sound good?

nobugshere:
> Hi @nobugshere, Is the plan to eventually use only the macos-latest (xcodebuild-based) workflow, which works fine? 

macos-latest.yml _is_ the workflow for release builds.  It was long before I came along and it's the one I maintain.  I think @ronys would like to migrate to cmake for the Mac build, but it's still a work-in-progress that I haven't really contributed to.  Honestly, I like using the Xcode GUI to manage multiple build configurations, but I can see where cmake would make it easier to maintain without a Mac.

> My goal is just `-Wall -Werror` for now.

I think we're close.

>  Perhaps the check is performed against gcc instead of clang?

That would make sense.  I'm sure there's a way to tell cmake to use clang for its tests instead of gcc.

> 4. I'm thinking of not enabling `-Werror` on Mac for now and letting you (or someone else with access to a Mac) continue this in a separate PR, does that sound good?

I agree with omitting -Werror from the cmake workflow since there is much more work to do there.

I'll take care of the macos-latest.yml workflow.  Other than the "override" warnings I mentioned above, it's ready for -Wall and -Werror.  I might also add -pedantic, it only added a few cases of lines ending with ";;".  (I haven't tried it on Linux.)

nobugshere (approved):

dskrvk:
This is ready now. I'll look into enabling `-Wextra` and `-Wpedantic`, at least for Linux (in another PR).

ronys:
@nobugshere is right regarding cmake and the Mac build, but without access to a Mac (for now), I'm pretty limited in what I can do with this.

PR #1852, opened 2026-07-06
Fix wx3.3 translation issues

sauravg:
See https://wxwidgets.org/help/msgid-literals/

### DIFF
--- a/Misc/rc2cpp.pl
+++ b/Misc/rc2cpp.pl
@@ -138,7 +138,7 @@ sub WriteHFile {
 #if !defined(_WIN32) || defined(__WX__)
 #include <map>
 #include "../os/typedefs.h" // for definition of TCHAR
-extern std::map<int, const TCHAR *> ${b}_st;
+extern std::map<int, const char *> ${b}_st;
 #endif
 
 #endif /* __${B}_ST_H */
@@ -168,23 +168,24 @@ sub WriteCPPFile {
 #define _(x) L ## x
 #include "./${BASE}_st.h"
 #include <utility>
+#include <wx/intl.h>
 ${include}
 
 using namespace std;
 
 namespace {
-  pair<int, const TCHAR *> Pairs[] = {
+  pair<int, const char *> Pairs[] = {
 PREAMBLE
 # print %MAP, sorted by constant name, just to be consistent
     my $key;
     foreach $key (sort keys %MAP) {
-        print CPP "    make_pair($key, _($MAP{$key})),\n";
+        print CPP "    make_pair($key, wxTRANSLATE($MAP{$key})),\n";
 }
     print CPP <<"POSTAMBLE";
   }; // Pairs array
 } // anonymous namespace
 
-map<int, const TCHAR *> ${b}_st(Pairs, Pairs + sizeof(Pairs)/sizeof(Pairs[0]));
+map<int, const char*> ${b}_st(Pairs, Pairs + sizeof(Pairs)/sizeof(Pairs[0]));
 
 #endif
 POSTAMBLE
--- a/src/core/StringX.cpp
+++ b/src/core/StringX.cpp
@@ -156,7 +156,7 @@ template<class T> void LoadAString(T &s, int id)
   cs.LoadString(id);
   s = cs;
 #else
-  s = _(core_st[id]).c_str();
+  s = core_st[id];
 #endif
 }
 
--- a/src/core/StringX.cpp
+++ b/src/core/StringX.cpp
@@ -16,6 +16,7 @@
 #include "os/pws_tchar.h"
 
 #if !defined(_WIN32) || defined(__WX__)
+#include <wx/intl.h>
 #include "core_st.h"
 #endif
 
@@ -155,7 +156,7 @@ template<class T> void LoadAString(T &s, int id)
   cs.LoadString(id);
   s = cs;
 #else
-  s = core_st[id];
+  s = wxGetTranslation(core_st[id]);
 #endif
 }
 
--- a/src/core/StringX.cpp
+++ b/src/core/StringX.cpp
@@ -16,7 +16,6 @@
 #include "os/pws_tchar.h"
 
 #if !defined(_WIN32) || defined(__WX__)
-#include <wx/intl.h>
 #include "core_st.h"
 #endif
 
--- a/src/ui/wxWidgets/GridCtrl.cpp
+++ b/src/ui/wxWidgets/GridCtrl.cpp
@@ -752,11 +752,11 @@ void GridCtrl::RearrangeItemsDateTimeBased(ItemsCollection& collection, int colu
       else if (dt.ParseDate(datetime, &end)) {
         collection.insert(std::pair<wxDateTime, const CItemData*>(dt, GetItem(row)));
         pws_os::Trace(L"Warning - Sorting may be incorrect for date string: %ls ; day: %d ; month: %d ; year: %d",
-          datetime.wc_str().data(), dt.GetDay(), dt.GetMonth(), dt.GetYear());
+          static_cast<const wchar_t*>(datetime.wc_str()), dt.GetDay(), dt.GetMonth(), dt.GetYear());
       }
     }
     else {
-      pws_os::Trace(L"Failed to parse item's date string: %ls ; using: %ls", datetime.wc_str().data(), wxDateTime::Today().FormatISOCombined().wc_str().data());
+      pws_os::Trace(L"Failed to parse item's date string: %ls ; using: %ls", static_cast<const wchar_t*>(datetime.wc_str()), static_cast<const wchar_t*>(wxDateTime::Today().FormatISOCombined().wc_str()));
       collection.insert(std::pair<wxDateTime, const CItemData*>(wxDateTime::Today(), GetItem(row)));
     }
   }
--- a/src/ui/wxWidgets/GridCtrl.cpp
+++ b/src/ui/wxWidgets/GridCtrl.cpp
@@ -752,11 +752,11 @@ void GridCtrl::RearrangeItemsDateTimeBased(ItemsCollection& collection, int colu
       else if (dt.ParseDate(datetime, &end)) {
         collection.insert(std::pair<wxDateTime, const CItemData*>(dt, GetItem(row)));
         pws_os::Trace(L"Warning - Sorting may be incorrect for date string: %ls ; day: %d ; month: %d ; year: %d",
-          datetime.wc_str(), dt.GetDay(), dt.GetMonth(), dt.GetYear());
+          datetime.wc_str().data(), dt.GetDay(), dt.GetMonth(), dt.GetYear());
       }
     }
     else {
-      pws_os::Trace(L"Failed to parse item's date string: %ls ; using: %ls", datetime.wc_str(), wxDateTime::Today().FormatISOCombined().wc_str());
+      pws_os::Trace(L"Failed to parse item's date string: %ls ; using: %ls", datetime.wc_str().data(), wxDateTime::Today().FormatISOCombined().wc_str().data());
       collection.insert(std::pair<wxDateTime, const CItemData*>(wxDateTime::Today(), GetItem(row)));
     }
   }
--- a/src/ui/wxWidgets/PWFiltersEditor.cpp
+++ b/src/ui/wxWidgets/PWFiltersEditor.cpp
@@ -141,7 +141,7 @@ wxString pwFiltersFTChoiceRenderer::getFieldTypeString(int ft)
 {
   std::map<int, wxString>::const_iterator iter = FieldTypeString.find(ft);
   ASSERT(iter != FieldTypeString.end());
-  return _(iter->second);
+  return iter->second;
 }
 
 /*!
@@ -369,11 +369,11 @@ wxSize pwFiltersFTChoiceRenderer::GetBestSize(wxGrid& WXUNUSED(grid), wxGridCell
   
     /* C++17 like
     for(const auto& [key, value] : FieldTypeString) {
-      m_bestSize.IncTo(DoGetBestSize(attr, dc, _(value)));
+      m_bestSize.IncTo(DoGetBestSize(attr, dc, value));
     }
      */
     for(const auto& kv : FieldTypeString) {
-      m_bestSize.IncTo(DoGetBestSize(attr, dc, _(kv.second)));
+      m_bestSize.IncTo(DoGetBestSize(attr, dc, kv.second));
     }
   }
   return m_bestSize;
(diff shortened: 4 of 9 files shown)


dskrvk (approved):
Nice, looks like in most of these cases _() was unnecessary in the first place.

dskrvk:
TIL: wxWidgets has [`wxPLURAL`](https://docs.wxwidgets.org/3.2.7/group__group__funcmacro__string.html#gadc7c2f1bab3914af93feb47945003409) which is the proper way to define multiple forms of plurals (this came up in a PR discussion recently).

Also https://www.gnu.org/software/gettext/manual/gettext.html#Plural-forms

src/core/StringX.cpp
```diff
@@ -156,7 +156,7 @@ template<class T> void LoadAString(T &s, int id)
   cs.LoadString(id);
   s = cs;
 #else
-  s = _(core_st[id]).c_str();
+  s = core_st[id];
```
huven: If this is committed, then maybe also remove

#include <wx/intl.h>

from line 19? Looks like that include was only for this line. Saves me from making two edits on every update 😊

sauravg: Makes sense. Done in [eb86350](https://github.com/pwsafe/pwsafe/pull/1852/commits/eb8635000ffcd2b980644a769f7250f61ee9b828)

nobugshere:
Back when I tried this in #1520, I received this comment:

> Looks like, that most of these changes should be reverted or implemented in other way. After these changes items in combo boxes in settings, messages in status bar, field names in dragbar tooltips, ... become untranslated.

I reverted the change and never got back to it.  It appears the same issues are present in this PR, at least when I build with wx3.2.10 on macOS.  Aparently some of these supposedly "unnecessary" uses of _() actually have an effect.

froggie2523:
This is a breaking change - most, if not all the _("<variable_name>") strings are translated into the current application language at runtime using PO files.

dskrvk:
The one change I'm not sure about is `FieldTypeString`: looking at the code, if we remove the translations around it, the strings are just passed as is to the UI. That said, I'm not sure how this could've worked before seeing as that map contains `wxString`s which aren't supposed to be translated.

nobugshere:
I think what's needed is to, judiciously, replace _() with explicit calls to wxGetTranslation().  I did some quick experimentation, but had mixed results.
- Fixing the drag bar tool-tips by inserting wxGetTranslation() in DragBarCtrl.cpp where _() use to be seems to work
- Trying to fix the double-click/shift-double-click combo boxes in OptionsPropertySheetDlg.cpp the same way did not.

Clearly, I have much to learn about this process.

ronys:
Non-Windows builds are now failing with:
```
2026-07-09T12:17:01.9824079Z /home/runner/work/pwsafe/pwsafe/src/ui/wxWidgets/GridCtrl.cpp: In member function ‘void GridCtrl::RearrangeItemsDateTimeBased(ItemsCollection&, int, bool)’:
2026-07-09T12:17:01.9827108Z /home/runner/work/pwsafe/pwsafe/src/ui/wxWidgets/GridCtrl.cpp:755:29: error: request for member ‘data’ in ‘datetime.wxString::wc_str()’, which is of non-class type ‘const wchar_t*’
2026-07-09T12:17:01.9829166Z   755 |           datetime.wc_str().data(), dt.GetDay(), dt.GetMonth(), dt.GetYear());
2026-07-09T12:17:01.9830033Z       |                             ^~~~
2026-07-09T12:17:01.9832289Z /home/runner/work/pwsafe/pwsafe/src/ui/wxWidgets/GridCtrl.cpp:759:96: error: request for member ‘data’ in ‘datetime.wxString::wc_str()’, which is of non-class type ‘const wchar_t*’
2026-07-09T12:17:01.9834561Z   759 |       pws_os::Trace(L"Failed to parse item's date string: %ls ; using: %ls", datetime.wc_str().data(), wxDateTime::Today().FormatISOCombined().wc_str().data());
2026-07-09T12:17:01.9836063Z       |                                                                                                ^~~~
2026-07-09T12:17:01.9838444Z /home/runner/work/pwsafe/pwsafe/src/ui/wxWidgets/GridCtrl.cpp:759:153: error: request for member ‘data’ in ‘wxDateTime::Today().wxDateTime::FormatISOCombined().wxString::wc_str()’, which is of non-class type ‘const wchar_t*’
2026-07-09T12:17:01.9841832Z   759 |       pws_os::Trace(L"Failed to parse item's date string: %ls ; using: %ls", datetime.wc_str().data(), wxDateTime::Today().FormatISOCombined().wc_str().data());
2026-07-09T12:17:01.9843407Z       |
```

src/ui/wxWidgets/wxUtilities.cpp
```diff
@@ -227,7 +227,7 @@ SafeCombinationCtrl* wxUtilities::CreateLabeledSafeCombinationCtrl(wxWindow* par
   auto *sizer = new wxBoxSizer(wxVERTICAL);
   parent->GetSizer()->Add(sizer, 0, wxBOTTOM|wxLEFT|wxRIGHT|wxEXPAND, 12);
 
-  auto *labelCtrl = new wxStaticText(parent, wxID_STATIC, _(label), wxDefaultPosition, wxDefaultSize, 0);
```
sauravg, line 230: Translations for label are done at the call sites

src/ui/wxWidgets/PasswordSafeFrame.cpp
```diff
@@ -3145,7 +3145,7 @@ void PasswordSafeFrame::UpdateSelChanged(const CItemData *pci)
     if (dca == -1)
       dca = PWSprefs::GetInstance()->GetPref(PWSprefs::DoubleClickAction);
   }
-  m_statusBar->SetStatusText(_(PWSprefs::GetDCAdescription(dca)), StatusBar::Field::DOUBLECLICK);
```
sauravg, line 3148: GetDCADescription() returns translated strings

Misc/rc2cpp.pl
```diff
@@ -138,7 +138,7 @@ sub WriteHFile {
 #if !defined(_WIN32) || defined(__WX__)
 #include <map>
 #include "../os/typedefs.h" // for definition of TCHAR
-extern std::map<int, const TCHAR *> ${b}_st;
```
sauravg, line 141: wxTRANSLATE returns `char *`

Misc/rc2cpp.pl
```diff
@@ -168,23 +168,24 @@ sub WriteCPPFile {
...
 using namespace std;
 
 namespace {
-  pair<int, const TCHAR *> Pairs[] = {
+  pair<int, const char *> Pairs[] = {
 PREAMBLE
 # print %MAP, sorted by constant name, just to be consistent
     my $key;
     foreach $key (sort keys %MAP) {
-        print CPP "    make_pair($key, _($MAP{$key})),\n";
```
sauravg, line 181: Strings in core_st.cpp are part of static variables, so they may not get translated with `_(...)`

See https://docs.wxwidgets.org/3.2/group__group__funcmacro__string.html#ga437ea6ba615b75dac8603e96ec864160

src/ui/wxWidgets/GridCtrl.cpp
```diff
@@ -752,11 +752,11 @@ void GridCtrl::RearrangeItemsDateTimeBased(ItemsCollection& collection, int colu
       else if (dt.ParseDate(datetime, &end)) {
         collection.insert(std::pair<wxDateTime, const CItemData*>(dt, GetItem(row)));
         pws_os::Trace(L"Warning - Sorting may be incorrect for date string: %ls ; day: %d ; month: %d ; year: %d",
-          datetime.wc_str(), dt.GetDay(), dt.GetMonth(), dt.GetYear());
+          static_cast<const wchar_t*>(datetime.wc_str()), dt.GetDay(), dt.GetMonth(), dt.GetYear());
```
sauravg, line 755: `wc_str()` may return a wraper object if `wx_USE_UNICODE_UTF8` is defined. Nothing to do with wx3.3 specifically

src/ui/wxWidgets/PWFiltersEditor.cpp
```diff
@@ -85,40 +85,40 @@ const std::map<int, wxString> FieldTypeString = {
...
-  {FT_ENTRYTYPE, _("Entry type")},
-  {FT_ENTRYSTATUS, _("Entry status")},
-  {FT_UNKNOWNFIELDS, _("Unknown fields")},
-  {FT_PASSWORDLEN, _("Password Length")},
-  {HT_PRESENT, _("Field present")},
-  {HT_ACTIVE, _("Field active")},
-  {HT_NUM, _("Number stored")},
-  {HT_MAX, _("Maximal Number stored")},
-  {HT_CHANGEDATE, _("Change Date")},
-  {HT_PASSWORDS, _("Previous Passwords")},
```
sauravg, line 98: These are static strings, `_(...)` may not work for them.

src/ui/wxWidgets/PasswordSafeFrame.cpp
```diff
@@ -859,7 +859,7 @@ void PasswordSafeFrame::AddLanguageMenu(wxMenu* parent)
   wxLanguage system_language = wxGetApp().GetSystemLanguage();
 
   for (auto &item : m_languages) {
-    wxString lang_name = _(get<1>(item.second));
+    wxString lang_name = wxGetTranslation(get<1>(item.second));
```
sauravg, line 862: This may require the language string literals in ` PasswordSafeFrame::RegisterLanguageMenuItems()` to be wrapped in `wxTRANSLATE(...)`

```c++
- AddLanguage( ID_LANGUAGE_CHINESE,   wxLANGUAGE_CHINESE_SIMPLIFIED, L"Chinese"  );  /* code: 'zh_CN' */
+ AddLanguage( ID_LANGUAGE_CHINESE,   wxLANGUAGE_CHINESE_SIMPLIFIED, wxTRANSLATE("Chinese")  );  /* code: 'zh_CN' */

```

Misc/rc2cpp.pl
```diff
@@ -168,23 +168,24 @@ sub WriteCPPFile {
 #define _(x) L ## x
```
dskrvk, line 168: No longer used.

Misc/rc2cpp.pl
dskrvk: It’s a macro, it doesn’t return anything. I didn’t get why we need this change.

src/ui/wxWidgets/GridCtrl.cpp
dskrvk: Maybe use `tostringx(datetime)` instead?

src/ui/wxWidgets/PasswordSafeFrame.cpp
dskrvk: I think so too, though `_()` should suffice there since it's not a static context.

src/ui/wxWidgets/PWFiltersEditor.cpp
```diff
@@ -369,11 +369,11 @@ wxSize pwFiltersFTChoiceRenderer::GetBestSize(wxGrid& WXUNUSED(grid), wxGridCell
   
     /* C++17 like
```
dskrvk, line 370: Nit: not sure what was the point of adding this comment, but given we're on C++20 now, we should either use the new syntax or delete the comment. As it is we're paying the cost of maintaining this commented-out code.

src/ui/wxWidgets/PasswordSafeFrame.cpp
nobugshere: Wouldn't you need _() here to expand to wxGetTranslation() ?

sauravg: We're doing the translation explicitly above, in PasswordSafeFrame.cpp:862. So `_(...)` is not be required here.

Or we can use `_(...)` in `PasswordSafeFrame::RegisterLanguageMenuItems()` and remove the wxGetTranslation() call here.

But essentially, the names of the languages would display in English, unless we have translations for those, and we use `_(...)` or `wxTRANSLATE(...)` + `wxGetTranslation()`.

src/ui/wxWidgets/GridCtrl.cpp
sauravg: No point in paying the runtime cost of constructing a C++ object when it can be done at compile-time.

Misc/rc2cpp.pl
sauravg: It translates to it's arg type, which is char *.

We can keep TCHAR & change `_(...)` to `_T(...)` in the perl script, but `_T(...)` & `wxT(...)` are now deprecated in wx

Misc/rc2cpp.pl
sauravg: Oh, I missed this! So there was no translation happening for strings in core_st?

ronys:
Hi,

From reading the comments, I can't decide if this is ready for merging or not. I'm converting this to 'draft'. When it's ready, please convert it back and I'll give it a final check and merge.

Thanks.

sauravg:
@ronys It's not ready. Could you please add your thoughts on the implications of the changes in rc2cpp.pl, in particular the change that replaces `_(...)` with `wxTRANSLATE(...)`?

Before this PR, the `_(...)` was redfined in core_st.cpp to `L"..."`, so it wasn't being used for translations anyway, at least not how that macro is usually used for translations. I don't see how translations was working for those strings, unless xgettext works regardless of what macro we surround raw strings with. But the changes in this PR explicitly add translation for those strings. Is that the right thing to do? What am I missing here?

Also weigh in on whether we should retain TCHAR * in core_st map.

src/ui/wxWidgets/PasswordSafeFrame.cpp
nobugshere: Ah, I missed that.  I think "wxTRANSLATE(...) + wxGetTranslation()" is better; if the user changes the language the menu should be updated with the new translation.

Misc/rc2cpp.pl
dskrvk: My understanding is previously this was done by `LoadAString`.

src/ui/wxWidgets/PasswordSafeFrame.cpp
dskrvk: Oh right, we don't want `_()` in `RegisterLanguageMenuItems` since it translates in place, not just marks for later translation.

Misc/rc2cpp.pl
dskrvk: What you have now in the script + StringX.cpp looks fine - as long as the compiler is happy I think we can keep `char *`.

ronys:
@sauravg - thanks for the clear summary.

As I understand it, `_()` is used by xgettext when scanning the source file, regardless of how it's defined, so that explains how it works in core_st.cpp

So in core_st.cpp I'd leave the _() in the code. As to its definition, it can be changed to 
`#define _(x) x`
along with replacing TCHAR with char and removing `#include "../os/typedefs.h" `

Finally, for clarity, I'd replace
`  s = _(core_st[id]).c_str();`
with
`  s = wxGetTranslation(core_st[id]).c_str();`
in line 159 of StringX.cpp

Issue #1856, opened 2026-07-08
Clicking any row in the "conflicting items" grid in comparison dialog hangs pwsafe


sauravg:
Repro:

1. Take any pwsafe db and make a copy
2. Change something in some of the items in the copied db
3. Compare with the original database (File -> Compare)
4. Click on any row in the grid that shows differences between conflicting items

In a second or two, pwsafe will hang.

rafaelx:
Fixed by #1857

PR #1857, opened 2026-07-08
Fix infinite loop in Comparison Dialog

sauravg:
https://github.com/pwsafe/pwsafe/issues/1856

Something changed in recent versions of wxWidgets that broke pwsafe's logic of always selecting 2 rows at a time. Use wxGrid's selection state data to avoid selecting a row that's already selected.

### DIFF
--- a/src/ui/wxWidgets/ComparisonGridTable.cpp
+++ b/src/ui/wxWidgets/ComparisonGridTable.cpp
@@ -552,16 +552,16 @@ wxPen ComparisonGrid::GetRowGridLinePen(int row)
 void ComparisonGrid::OnGridRangeSelect(wxGridRangeSelectEvent& evt)
 {
   //select grids asynchronously, or else we get in an infinite loop of selections & their notifications
-  if (evt.GetTopRow()%2 != 0 && !IsInSelection(evt.GetTopRow()-1, 0)) {
+  if (auto topRow = evt.GetTopRow(); topRow%2 != 0 && !IsInSelection(topRow-1, 0)) {
     wxCommandEvent cmdEvent(EVT_SELECT_GRID_ROW);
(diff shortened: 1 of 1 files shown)


src/ui/wxWidgets/ComparisonGridTable.cpp
```diff
@@ -552,13 +552,13 @@ wxPen ComparisonGrid::GetRowGridLinePen(int row)
...
   //select grids asynchronously, or else we get in an infinite loop of selections & their notifications
-  if (evt.GetTopRow()%2 != 0) {
+  if (evt.GetTopRow()%2 != 0 && !IsInSelection(evt.GetTopRow()-1, 0)) {
     wxCommandEvent cmdEvent(EVT_SELECT_GRID_ROW);
     cmdEvent.SetEventObject(evt.GetEventObject());
     cmdEvent.SetInt(evt.GetTopRow()-1);
     GetEventHandler()->AddPendingEvent(cmdEvent);
   }
-  if (evt.GetBottomRow()%2 == 0) {
+  if (evt.GetBottomRow()%2 == 0 && !IsInSelection(evt.GetTopRow()+1, 0)) {
```
dskrvk: Shouldn’t it be
```suggestion
  if (evt.GetBottomRow()%2 == 0 && !IsInSelection(evt. GetBottomRow()+1, 0)) {
```

sauravg: You're right! Poor copy/paste on my part. Fixed in [846c5a8](https://github.com/pwsafe/pwsafe/pull/1857/commits/846c5a8e4c68d8838a0accbda7e1c660a2a05797)

dskrvk (approved):

PR #1859, opened 2026-07-11
[wx] Fix - Build fails on BSD systems

froggie2523:
The "Fix Linux compiler warnings & treat any future ones as errors" change in PR #1851 causes compilation to fail on BSD systems because Clang is the default compiler.

This PR provides a quick fix to restore successful compilation on FreeBSD and OpenBSD.

Tested on Debian/FreeBSD/OpenBSD.

FreeBSD - before:
pwsafe_1 24-wxWidgets-bug-FreeBSD-CMake-01
FreeBSD - after (it builds successfully - warnings only):
pwsafe_1 24-wxWidgets-bug-FreeBSD-CMake-02
OpenBSD - before:
pwsafe_1 24-wxWidgets-bug-OpenBSD-CMake-01
OpenBSD - after (it builds successfully - warnings only):
pwsafe_1 24-wxWidgets-bug-OpenBSD-CMake-02

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -220,10 +220,12 @@ else ()
   set(CMAKE_CXX_FLAGS
     "${CMAKE_CXX_FLAGS} -Wall -DUNICODE -DWCHAR_INCOMPATIBLE_XMLCH ${CMAKE_WXWINDOWS_CXX_FLAGS}")
   if (NOT APPLE)
-    # TODO: with the below flag the Mac cmake-based build fails with an error:
+    # TODO: with the below flag the Mac and BSD cmake-based builds fail with an error such as:
     #   clang: error: argument unused during compilation: '-fstack-clash-protection'
     # it's possible check_cxx_compiler_flag is using the wrong compiler for the check?
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Werror")
+    if (${CMAKE_SYSTEM_NAME} MATCHES "Linux")
+      set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Werror")
+    endif (${CMAKE_SYSTEM_NAME} MATCHES "Linux")
   endif (NOT APPLE)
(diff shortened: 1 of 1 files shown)


dskrvk:
We should probably configure GitHub CI to build on BSD too so these things could be caught in time.

dskrvk (approved):

nobugshere:
I don't think GitHub has a BSD image.  But, maybe, repeating the Linux build using clang would help?  
I'm working with clang on the Mac, but the particular file in the example is one that is not used in the Mac build.  I get a number of other warnings in a few gui headers.

CMakeLists.txt
```diff
@@ -220,10 +220,12 @@ else ()
   set(CMAKE_CXX_FLAGS
     "${CMAKE_CXX_FLAGS} -Wall -DUNICODE -DWCHAR_INCOMPATIBLE_XMLCH ${CMAKE_WXWINDOWS_CXX_FLAGS}")
   if (NOT APPLE)
-    # TODO: with the below flag the Mac cmake-based build fails with an error:
+    # TODO: with the below flag the Mac and BSD cmake-based builds fail with an error such as:
     #   clang: error: argument unused during compilation: '-fstack-clash-protection'
     # it's possible check_cxx_compiler_flag is using the wrong compiler for the check?
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Werror")
+    if (${CMAKE_SYSTEM_NAME} MATCHES "Linux")
```
nobugshere: If you can match "Linux" does it still need to be enclosed in "NOT APPLE"?

ronys:
Indeed, GitHub doesn't have any BSD "runner images", according to https://docs.github.com/en/actions/reference/runners/github-hosted-runners#standard-github-hosted-runners-for-public-repositories

In any case, I'd much rather fix these warnings at the source, rather than just turning `-Werror` off selectively.
Perhaps we can add clang builds to the workflow in addition to gcc?

nobugshere (approved):

PR #1860, opened 2026-07-12
[macOS] Fix compiler warnings and add -Wall -pedantic -Werror

nobugshere:
Xcode flavored Follow-on to #1851 
Tested with: macOS M1Max 15.7.7 Sequoia, Xcode 16.4, wx 3.2.10 (local build)

1. Adds -Wall and -pedantic to all targets and configurations
2. **Core** and **os** libs also have -Wextra (They've already had that for some time now.)
3. The Release configuration now has -Werror (But, not the other configs.  Release is used by the GitHub workflows.)
4. Fixed a few warning and indentation issues
5. Suppressed some hidden overload warnings, discussion follows:

Clang with -Wall reports a number of warnings that gcc apparently doesn't include in -Wall.
There were several warnings of this type spread across three header files:

```
/Users/rpowell/Documents/Xcode/pwsafe/src/ui/wxWidgets/./RecentDbList.h:48:8: error: 'RecentDbList::Save' hides overloaded virtual function [-Werror,-Woverloaded-virtual]
   48 |   void Save() const {
      |        ^
In file included from /Users/rpowell/Documents/Xcode/pwsafe/src/ui/wxWidgets/SafeCombinationEntryDlg.cpp:37:
In file included from /Users/rpowell/Documents/Xcode/pwsafe/src/ui/wxWidgets/PWSafeApp.h:29:
In file included from /Users/rpowell/Documents/Xcode/pwsafe/src/ui/wxWidgets/./RecentDbList.h:17:
In file included from /Users/rpowell/NotBackedUp/pwsafe_testing/wxWidgets/wxWidgets-3.2.10/include/wx/docview.h:22:
/Users/rpowell/NotBackedUp/pwsafe_testing/wxWidgets/wxWidgets-3.2.10/include/wx/filehistory.h:54:18: note: hidden overloaded virtual function 'wxFileHistoryBase::Save' declared here: different number of parameters (1 vs 0)
   54 |     virtual void Save(wxConfigBase& config);
      |                  ^
1 error generated.
```

Various combinations of "virtual" and "override" did not work in these cases due to there being different numbers of parameters.  Two of them were easily fixed by adding "using" statements, in other cases "using" just lead to ambiguous overloads at the call sites.  I started to rename the offending functions, but I wasn't liking it, so instead I inserted a pragma in the headers.  If there are objections to either approach, I can take another run at renaming.

```
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Woverloaded-virtual"
...
#pragma GCC diagnostic pop
```

### DIFF
--- a/docs/README.MAC.DEVELOPERS.md
+++ b/docs/README.MAC.DEVELOPERS.md
@@ -25,6 +25,8 @@ It is organized in the following sections:
 
 1. Install the dependencies via homebrew:
 
+Note: The default wxWidgets in Homebrew is v3.3.x. Password Safe does not yet build with that version.  wxwidgets@3.2 should work better.
+
 ```
 brew install wxwidgets yubikey-personalization googletest
 ```
--- a/docs/README.MAC.DEVELOPERS.md
+++ b/docs/README.MAC.DEVELOPERS.md
@@ -28,7 +28,7 @@ It is organized in the following sections:
 Note: The default wxWidgets in Homebrew is v3.3.x. Password Safe does not yet build with that version.  wxwidgets@3.2 should work better.
 
 ```
-brew install wxwidgets yubikey-personalization googletest
+brew install wxwidgets@3.2 yubikey-personalization googletest
 ```
 
 2. Configure the project
--- a/src/ui/wxWidgets/PasswordSafeFrame.h
+++ b/src/ui/wxWidgets/PasswordSafeFrame.h
@@ -710,7 +710,12 @@ private:
   void UpdateMainToolbarSeparators(bool insert);
   void DeleteMainToolbarSeparators();
   wxAuiPaneInfo& GetMainToolbarPane();
+
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Woverloaded-virtual"
   wxAuiToolBar* GetToolBar() { return m_Toolbar; }
+  void CreateStatusBar();
+#pragma GCC diagnostic pop
 
   void CreateDragBar();
   void UpdateDragbarTooltips();
@@ -724,8 +729,6 @@ private:
   wxAuiPaneInfo& GetSearchBarPane();
   PasswordSafeSearch* GetSearchBar() { return m_search; };
   void UpdateSearchBarVisibility();
-
-  void CreateStatusBar();
 
   long GetEventRUEIndex(const wxCommandEvent& evt) const;
   bool IsRUEEvent(const wxCommandEvent& evt) const;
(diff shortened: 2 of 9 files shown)


nobugshere:
After this is merged, I'll follow-up with an update to the cmake build to add -Werror and downgrade Xcode to 26.3 so it will quit failing whenever GitHub gives it the macOS 15 image instead of 26. (Xcode 26.3 is in both OS images.)

docs/README.MAC.DEVELOPERS.md
```diff
@@ -25,6 +25,8 @@ It is organized in the following sections:
 
 1. Install the dependencies via homebrew:
 
+Note: The default wxWidgets in Homebrew is v3.3.x. Password Safe does not yet build with that version.  wxwidgets@3.2 should work better.
```
dskrvk, line 28: Can we instead specify the version in the example below? A few people would probably copy-paste it without paying attention to the note.

src/ui/wxWidgets/PasswordSafeFrame.h
```diff
@@ -710,7 +710,12 @@ class PasswordSafeFrame : public wxFrame, public Observer, public wxModalDialogH
   void UpdateMainToolbarSeparators(bool insert);
   void DeleteMainToolbarSeparators();
   wxAuiPaneInfo& GetMainToolbarPane();
+
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Woverloaded-virtual"
   wxAuiToolBar* GetToolBar() { return m_Toolbar; }
```
dskrvk, line 716: I also considered the rename option, but in this case it seems we do want to hide the base member since it returns an unrelated `wxToolBar`, and there's no way to add one to a `PasswordSafeFrame`.

src/ui/wxWidgets/PasswordSafeFrame.h
```diff
@@ -710,7 +710,12 @@ class PasswordSafeFrame : public wxFrame, public Observer, public wxModalDialogH
...
+  void CreateStatusBar();
```
dskrvk, line 717: Here too, if we kept the base `CreateStatusBar` available, it would provide a way to add a `wxStatusBar` to this control, and we don't want that, so hiding does seem like a better approach.

src/ui/wxWidgets/PasswordSafeFrame.h
nobugshere: Yes, the same reasoning applies to the other cases where the base function just does different things.  I guess one could implement all the overloads and have them call assert() if they are called, but that doesn't seem very good either.  

Hiding the cases we have and using -Wall to catch future ones seems like a reasonable approach.

docs/README.MAC.DEVELOPERS.md
nobugshere: Done. Thanks!

src/ui/wxWidgets/PasswordSafeFrame.h
dskrvk: The changes LGTM, though I'm no expert in C++ best practices.

dskrvk (approved):

PR #1861, opened 2026-07-12
[wx] Fix - Flatpak build fails with the latest compiler flags

froggie2523:
This PR provides a fix that allows the Flatpak build to compile successfully with the latest source code.

Issue:
The "Fix Linux compiler warnings & treat any future ones as errors" change in PR #1851 causes the Flatpak build to fail due to conflicting compiler flags.

Reason:
Inside a Flatpak build, the Freedesktop SDK already provides hardening flags via CFLAGS/CXXFLAGS, including _FORTIFY_SOURCE. As a result, the build ends up with something like below, producing the redefinition error (previously, it was a warning only).
error: ‘_FORTIFY_SOURCE’ redefined [-Werror]
note: this is the location of the previous definition
cc1plus: all warnings being treated as errors
ninja: build stopped: subcommand failed.
Error: module pwsafe: Child process exited with code 1

Tested on Ubuntu 26.04/Fedora 44.

pwsafe_1 24-wxWidgets-bug-flatpak-Werror-01
pwsafe_1 24-wxWidgets-bug-flatpak-Werror-02


CMakeModules/harden/harden-gcc.cmake
```diff
@@ -38,7 +38,9 @@ if (HARDEN_USE_VISIBILITY)
   endfunction()
 endif ()
 
-harden_add_compile_definition(_FORTIFY_SOURCE=2 TRUE)
+if (NOT CMAKE_CXX_FLAGS MATCHES "_FORTIFY_SOURCE")
+  harden_add_compile_definition(_FORTIFY_SOURCE=2 TRUE)
```
dskrvk: Maybe move the check inside `harden_add_compile_definition` so all callers can benefit (even though there's only one right now)?

dskrvk (approved):

CMakeModules/harden/harden-gcc.cmake
froggie2523: Done.

CMakeModules/harden/harden-gcc.cmake
```diff
@@ -38,9 +38,7 @@ if (HARDEN_USE_VISIBILITY)
   endfunction()
 endif ()
 
-if (NOT CMAKE_CXX_FLAGS MATCHES "_FORTIFY_SOURCE")
-  harden_add_compile_definition(_FORTIFY_SOURCE=2 TRUE)
-endif ()
+harden_add_compile_definition(_FORTIFY_SOURCE=2 TRUE)
```
dskrvk: Perhaps it's better to pass the name and value as separate arguments, both to avoid the regex parsing and to make this invocation look a bit more natural?
```
harden_add_compile_definition(_FORTIFY_SOURCE 2 TRUE)
```

dskrvk (approved):

PR #1862, opened 2026-07-12
 Fix main window position restore double-scaling at >100% display scale

martona:
This reverts commit e6491bef3 ("Fix Windows scaling issue"), restoring the behavior of PR #1774, and removes the GetMonitorDPI() helper introduced by that commit (this was its only call site). Fixes the window position/size restore bug re-reported by @ColeBantam on 3.72.01 at 175% scale (see #1774 comments).

The saved rect is written verbatim from GetWindowRect() (OnMove/OnSize) and handed back verbatim to SetWindowPlacement(). Both calls run on the main GUI thread, and both run after that thread has been elevated to per-monitor-V2 (Fonts::GetInstance() -> tweakFontSizes() elevates before the master password dialog, which precedes any save or restore). So the physical pixel coordinate space the OS presents is the same on both ends, and the round-trip is correct by construction. 

e6491bef3's rationale is captured in the comment it added to GetDPI(): "Since we're currently DPI unaware, this will always return 96." That holds for the process manifest, but not at the call sites: by the time save/restore run, the thread has been elevated, so GetDpiForSystem() returns the real system DPI. This PR also corrects that GetDPI() comment (the returned value depends on the calling thread's DPI awareness context, not the manifest) and adds a comment at PlaceWindow() documenting the round-trip invariant, so the scaling doesn't get reintroduced a third time.

Verified: at 200% scale, a seeded Position rect round-trips through open -> exit unchanged with this fix; stock 3.72.1 restores it at 2x.

@ronys: since save and restore are symmetric, the verbatim restore in 3.72.0 can't itself have produced the misplacement you saw before e6491bef3: that was likely a different defect. I can perhaps track that down rather than have it patched here again. Also refining my earlier comment in #1774: I suggested the flip-flop came from save and restore running under different thread DPI contexts. On closer analysis they always run under the same context.

### DIFF
--- a/src/core/PWSprefs.h
+++ b/src/core/PWSprefs.h
@@ -227,11 +227,10 @@ public:
   StringX GetAllStringPrefs(const bool bUseCopy = false);
 
   // Special cases
-  void GetPrefRect(long &top, long &bottom, long &left, long &right) const;
-  // dpi: display scale the rect was captured under, for rescaling the window
-  // at restore if it changed since; 0 = unknown (legacy config, non-MFC UI)
+  // dpi (optional out): display scale the rect was captured under, for rescaling
+  // the window at restore if it changed since; 0 = unknown (legacy config, non-MFC UI)
+  void GetPrefRect(long &top, long &bottom, long &left, long &right, unsigned *dpi = nullptr) const;
   void SetPrefRect(long top, long bottom, long left, long right, unsigned dpi = 0);
-  unsigned GetPrefRectDPI() const {return m_rect.dpi;}
   void GetPrefPSSRect(long &top, long &bottom, long &left, long &right) const;
   void SetPrefPSSRect(long top, long bottom, long left, long right);
   unsigned int GetMRUList(std::vector<stringT> &MRUFiles) const;
--- a/src/core/PWSprefs.h
+++ b/src/core/PWSprefs.h
@@ -228,7 +228,10 @@ public:
 
   // Special cases
   void GetPrefRect(long &top, long &bottom, long &left, long &right) const;
-  void SetPrefRect(long top, long bottom, long left, long right);
+  // dpi: display scale the rect was captured under, for rescaling the window
+  // at restore if it changed since; 0 = unknown (legacy config, non-MFC UI)
+  void SetPrefRect(long top, long bottom, long left, long right, unsigned dpi = 0);
+  unsigned GetPrefRectDPI() const {return m_rect.dpi;}
   void GetPrefPSSRect(long &top, long &bottom, long &left, long &right) const;
   void SetPrefPSSRect(long top, long bottom, long left, long right);
   unsigned int GetMRUList(std::vector<stringT> &MRUFiles) const;
@@ -324,7 +327,7 @@ private:
   bool m_boolValues[NumBoolPrefs];
   unsigned int m_intValues[NumIntPrefs];
   StringX m_stringValues[NumStringPrefs];
-  struct {long top, bottom, left, right; bool changed;} m_rect, m_PSSrect;
+  struct {long top, bottom, left, right; unsigned dpi; bool changed;} m_rect, m_PSSrect;
   bool m_boolChanged[NumBoolPrefs];
   bool m_intChanged[NumIntPrefs];
   bool m_stringChanged[NumStringPrefs];
--- a/src/core/PWSprefs.cpp
+++ b/src/core/PWSprefs.cpp
@@ -260,8 +260,10 @@ PWSprefs::PWSprefs() : m_pXML_Config(nullptr)
   for (i = 0; i < NumStringPrefs; i++) m_stringChanged[i] = false;
 
   m_rect.top = m_rect.bottom = m_rect.left = m_rect.right = -1;
+  m_rect.dpi = 0;
   m_rect.changed = false;
   m_PSSrect.top = m_PSSrect.bottom = m_PSSrect.left = m_PSSrect.right = -1;
+  m_PSSrect.dpi = 0;
   m_PSSrect.changed = false;
 
   InitializePreferences();
@@ -749,7 +751,7 @@ bool PWSprefs::DeletePref(const StringX &name)
 }
 
 void PWSprefs::SetPrefRect(long top, long bottom,
-                           long left, long right)
+                           long left, long right, unsigned dpi)
 {
   if (m_rect.top != top) {
     m_rect.top = top; m_rect.changed = true;
@@ -762,6 +764,9 @@ void PWSprefs::SetPrefRect(long top, long bottom,
   }
   if (m_rect.right != right) {
     m_rect.right = right; m_rect.changed = true;
+  }
+  if (m_rect.dpi != dpi) {
+    m_rect.dpi = dpi; m_rect.changed = true;
   }
   if (m_rect.changed)
     m_prefs_changed[APP_PREF] = true;
@@ -1364,6 +1369,7 @@ void PWSprefs::LoadProfileFromRegistry()
   m_rect.bottom = pws_os::RegReadValue(PWS_REG_POSITION, _T("bottom"), -1);
   m_rect.left = pws_os::RegReadValue(PWS_REG_POSITION, _T("left"), -1);
   m_rect.right = pws_os::RegReadValue(PWS_REG_POSITION, _T("right"), -1);
+  m_rect.dpi = pws_os::RegReadValue(PWS_REG_POSITION, _T("dpi"), 0);
 
   // Load last Password subset window size & pos:
   m_PSSrect.top = pws_os::RegReadValue(PWS_REG_POSITION, _T("PSS_top"), -1);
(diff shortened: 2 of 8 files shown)


ColeBantam:
Thanks Martona for fixing this (hopefully) :)

Can you test that patch at "uneven" scale-factors like 125% or 175% too? because sometimes apps work on straight scale factors but not on "uneven" ones ...

martona:
Yes, I have just now, at 125% and 150% - works perfectly. One note regarding that: saving at one DPI setting and then restoring at another _will_ preserve window placement perfectly _in physical pixels_. Which is to say the window will appear at a different size after scaling. I can't help but think that this was @ronys' root cause for  e6491be. 

This could be worked around by saving the DPI (that we don't do right now) as well as the physical pixels (which is the only thing we store), and upon restoration MulDivving with the saved value and the current system DPI. I could potentially add this, but I don't think it's necessary as DPI setting is something that you pick once and then stick with it. Maybe adjust upwards every 5 years as your eyesight deteriorates. :) Thoughts?

ColeBantam:
> Yes, I have just now, at 125% and 150% - works perfectly. 

Great, Thanks :)

> I don't think it's necessary as DPI setting is something that you pick once and then stick with it. Maybe adjust upwards every 5 years as your eyesight deteriorates. :) Thoughts?

Ahm, you are right when it comes to Desktops, but Notebook Users (at least in Businesses) often connect to external Monitors, that might have different scale-factors than the builtin display. I mean position can't really be restored when there is a completely different screen, but it would be cool, if the window-size would be more or less the "used" one.

Also there could be a case when a user has two screens attached with different dpi and moves the app from one screen to the other. In that case, the app also should maintain its visual size. Just a notice, doesn't know if thats currently a problem.

Update:
Hmmm, the relative position on the screen could be more or less reconstructed if the last resolution of the screen is also saved. Or maybe it would be better to just "remember" those settings per Resolution/dpi combination, so that the size/position on the screen is saved for each monitor. Than a notebook user that boots in business workspace always has the same setting and when in homeoffice too. Thats most likely poor english, but I hope you understand me :D

martona:
You are right of course. Laptops and external monitors are a very valid source of frequent DPI flips.

I am going to ignore what you said about moving the app window between different DPI monitors. :) That is a much larger project (handling WM_DPICHANGED and all the baggage that comes with it). 

But I implemented dropping the current (window) DPI in prefs, and reading it back on app startup, then recalculating the window size based on the factor (if any): the saved window rect is tagged with the display scale it was captured at. On restore, if the current scale differs, the window is resized by current/saved so it reopens at the same logical size, with the top-left anchored to the same physical point (and slid back on-screen if the larger size would overflow a monitor edge). When the scale is unchanged the ratio is 1 and nothing happens.

Tested flipping the same display between 125% and 200% back and forth across restarts: window comes back the right size every time. Back-compat: existing configs have no scale tag; they restore exactly as before and pick up the tag on the next save. No migration, no reset.

ColeBantam:
I understand that moving between differently scaled monitors is a different project, and its not needed to fix the regression that was introduced back in 3.65, as that might not have worked before also :D

Sounds great what you achieved now. Would be good to finally have a modern version of PasswordSafe usable, because the last working version, 3.46.1, is really "ancient" now :D

And restoring at least the window-size in different DPI-Environments is a cool "new feature" for Notebook Users :)

src/core/PWSprefs.h
```diff
@@ -228,7 +228,10 @@ class PWSprefs
 
   // Special cases
   void GetPrefRect(long &top, long &bottom, long &left, long &right) const;
-  void SetPrefRect(long top, long bottom, long left, long right);
+  // dpi: display scale the rect was captured under, for rescaling the window
+  // at restore if it changed since; 0 = unknown (legacy config, non-MFC UI)
+  void SetPrefRect(long top, long bottom, long left, long right, unsigned dpi = 0);
+  unsigned GetPrefRectDPI() const {return m_rect.dpi;}
```
dskrvk: If we're adding dpi to `SetPrefRect`'s parameters, why not similarly extend `GetPrefRect` to also return the dpi (instead of adding a new getter)?

martona: Because we're not the only caller, and an out-param can't be a defaulted reference. I did not want to use a pointer out-param because it felt very un-MFC-like. On the other hand the pwsafe code has examples of it, so it wouldn't be unprecedented. Changed.

ColeBantam:
I couldn't sleep last night because this issue kept running through my head. If I understand it correctly, the window position is still being stored using absolute coordinates. That explains why switching between monitors with different resolutions causes the window to appear somewhere other than where the user expects it.

What if, instead of storing absolute coordinates, the window position was stored as relative values? Here's one possible approach:

## Storing the window position as relative values

Let's use a notebook with a 1000-pixel-wide display and an external monitor with a width of 2000 pixels as an example. Assume the PWSafe window is 200 pixels wide.

On the notebook, the window can move horizontally between 0 and `1000 - 200 = 800` pixels. In other words, a horizontal position of 0 pixels corresponds to 0%, while 800 pixels corresponds to 100%.

If the application stores the position as a percentage instead of an absolute pixel value, the window position can be reconstructed correctly on displays with different resolutions.

## Restoring the window on the larger monitor

On the external monitor, the available horizontal range is `2000 - 200 = 1800` pixels, so 1800 pixels now represent 100%.

If the window was positioned at the far right on the notebook, the stored horizontal position would be 100%. When restoring the window on the larger monitor, it would again appear at the far right.

With the current absolute-coordinate approach, however, the stored value of 800 pixels would place the window roughly in the middle of the larger screen—which is almost certainly not where the user expects it. 😁

## The taskbar edge case

So far this is fairly straightforward, but things may become slightly more complicated once the taskbar is taken into account.

The first question is whether it needs to be considered at all. That depends on whether Windows already provides applications with a client area (or "working area") that excludes the taskbar, or whether coordinates are based on the full monitor dimensions.

So here's a question for the developers:

If the taskbar is docked on the left side of the screen and the PWSafe window is positioned flush against the left edge, is its X coordinate 0, or would it be something like 80 pixels (assuming an 80-pixel-wide taskbar)?

## Magnetic window edges

The following goes beyond fixing the regression, but I wanted to mention it because it could be implemented almost as a by-product of switching to relative coordinates.

Once window positions are stored as percentages, it becomes very easy to automatically snap windows to the desktop edges when they are *close enough*.

For example, if the horizontal position is stored as 98%, it's probably safe to assume that the user intended to place the window flush against the right edge but simply couldn't position it perfectly by hand.

In that case, values above a configurable threshold could simply be rounded to the nearest edge. For example:

* 1.45% → 0%
* 98.6% → 100%

The snapping threshold could be exposed as a user-configurable option. Users who don't want this behavior could simply set the threshold to 0, effectively disabling the feature.

dskrvk:
Storing the window position as a % seems like a good idea, but to that I'd also add storing the width/height as a % of window area, so that the window takes up the same % of space on every monitor. IMO this would look more natural - the window elements become bigger/smaller depending on the DPI, but the relative window size itself doesn't change. This also avoids the problem of the window overflowing the available area (e.g. position at 50% of a 1920px screen, but the width was 1000px).

> Magnetic window edges

Goes too far IMO, since it's the window manager's responsibility to support edge snapping. We just need to faithfully store how the window was represented on the screen (in relative terms). Let's move this into the repo's Discussion tab since this goes outside the scope of this PR?

ColeBantam:
> ... I'd also add storing the width/height as a % of window area, so that the window takes up the same % of space on every monitor. 

I personally would not want that. The Window Size should always stay the same, and not grow just because its a bigger screen.

> the window elements become bigger/smaller depending on the DPI, but the relative window size itself doesn't change.

That would be strange. Icons should NOT become bigger on a screen with higher dpi, they just should become more clear because of the higher resolution.
 
> Magnetic window edges
> Goes too far IMO, since it's the window manager's responsibility to support edge snapping. We just need to faithfully store how the window was represented on the screen (in relative terms). Let's move this into the repo's Discussion tab since this goes outside the scope of this PR?

I know that is a bit out of range of the regression, but Windows doesn't have such a feature and changing to relative units for the position would quite easily allow for this feature too I think. Its just a rounding-function in the end.

ronys:
Hi,

I've made a pre-release available at https://sourceforge.net/projects/passwordsafe/files/PreReleases/3.72.2pre/pwsafe64-3.72.2pre.exe/download
I'm not sure there will be a 3.72.2 release, or if this will be rolled into 3.73.0 later.

@martona Thanks once again!

@ColeBantam About the % suggestion: It's a nice idea, but I think the current solution is more straightforward, so I suggest we go forward with this.

ColeBantam:
> I've made a pre-release available at https://sourceforge.net/projects/passwordsafe/files/PreReleases/3.72.2pre/pwsafe64-3.72.2pre.exe/download I'm not sure there will be a 3.72.2 release, or if this will be rolled into 3.73.0 later.

The Window-Size and Position restore works for me in this build :)

But two issues visible at testing:

Icons in the menu don't get enough space and collide with the menu-text:

PWSafe_Spacing-Glitch_1

And the editing-window for Password-Entrys does waste a looot of space:

PWSafe_Spacing-Glitch_2

I can create a new bugreport for this if it helps.

ronys:
Hi @ColeBantam 

Thanks for the confirmation.

As to the new issues: Please open a new ticket for these. I suggest a title like "More display issues with > 100% display scale".

ColeBantam:
Done. See #1877.

PR #1863, opened 2026-07-12
[wx] Remove an unused variable

nobugshere:
Sometimes, under some conditions, I get the unused variable warning, but most of the time not.  Go figure.


dskrvk (approved):
Weird, my IDE also doesn't mark it as unused

ronys:
Strange indeed. Mysterious are the ways of the Compiler, blessed be its output :-)

PR #1867, opened 2026-07-15
Enable and fix extra compiler warnings

dskrvk:
Also enable pedantic, but @nobugshere already fixed those in https://github.com/pwsafe/pwsafe/pull/1860

### DIFF
--- a/src/core/CoreImpExp.cpp
+++ b/src/core/CoreImpExp.cpp
@@ -1151,55 +1151,54 @@ int PWScore::ImportPlaintextFile(const StringX &ImportedPrefix,
       startpos < slinebuf.size();
       /* startpos advanced in body */) {
       size_t nextchar = slinebuf.find_first_of(fieldSeparator, startpos);
-      if (nextchar == StringX::npos)
+      if (nextchar == StringX::npos) {
         nextchar = slinebuf.size();
-      if (nextchar >= 0) {
-        if (itoken != columns[CItem::NOTES]) {
-          const StringX tsx(slinebuf.substr(startpos, nextchar - startpos));
-          tokens.push_back(tsx.c_str());
-        }
-        else {
-          // Notes field which may be double-quoted, and
-          // if they are, they may span more than one line.
-          stringT note(slinebuf.substr(startpos).c_str(), nextchar - startpos);
-          size_t first_quote = note.find_first_of('\"');
-          size_t last_quote = note.find_last_of('\"');
-          if (first_quote == last_quote && first_quote != stringT::npos) {
-            //there was exactly one quote, meaning that we've a multi-line Note
-            bool noteClosed = false;
-            do {
-              if (!getline(iss, linebuf, '\n')) {
-                Format(cs_error, IDSC_IMPMISSINGQUOTE, numlines);
-                rpt.WriteLine(cs_error);
-                return (numImported > 0) ? SUCCESS : INVALID_FORMAT;
-              }
-              numlines++;
-              // remove MS-DOS linebreaks, if needed.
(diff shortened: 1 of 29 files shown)


src/core/crypto/Fish.h
```diff
@@ -47,7 +47,7 @@ template<typename FishT, typename HashT>
 FishT *makeFish(const unsigned char* pass, unsigned int passlen,
   const unsigned char* salt, unsigned int saltlen)
 {
-  unsigned char passkey[HashT::HASHLEN];
+  unsigned char passkey[HashT::HASHLEN] = {};
```
dskrvk, line 50: Fixes

> error: ‘passkey’ may be used uninitialized

Shouldn't be any change in behavior, since as I understand we're just explicitly saying we want an all-zero array. Previously this was the implicit behavior (automatic array variables are default-initialized).

src/core/CoreImpExp.cpp
```diff
@@ -1151,55 +1151,54 @@ int PWScore::ImportPlaintextFile(const StringX &ImportedPrefix,
       startpos < slinebuf.size();
       /* startpos advanced in body */) {
       size_t nextchar = slinebuf.find_first_of(fieldSeparator, startpos);
-      if (nextchar == StringX::npos)
+      if (nextchar == StringX::npos) {
         nextchar = slinebuf.size();
-      if (nextchar >= 0) {
```
dskrvk, line 1156: `size_t` is unsigned, pointless check

src/ui/wxWidgets/ManageFiltersTable.cpp
```diff
@@ -597,7 +597,7 @@ stringT pwSortedManageFilters::GetSelectedFilterName()
 
 void pwSortedManageFilters::ClearFlagAt(size_t idx, UINT mask)
 {
-  if(idx >= 0 && idx < m_SortFilterData.size()) {
+  if (idx < m_SortFilterData.size()) {
```
dskrvk, line 600: Same as above

src/ui/wxWidgets/wxUtilities.cpp
```diff
@@ -156,7 +156,7 @@ bool MultiCheckboxValidator::Validate(wxWindow* parent)
   }
 }
 
-void UpdatePasswordTextCtrl(wxSizer *sizer, wxTextCtrl* &textCtrl, const wxString text, wxControl* before, const int style)
+void UpdatePasswordTextCtrl([[maybe_unused]] wxSizer *sizer, wxTextCtrl* &textCtrl, const wxString &text, [[maybe_unused]] wxControl* before, const int style)
```
dskrvk, line 159: May be used in non-WXGTK environments

nobugshere:
FYI...WIth -Wextra I get a few warnings in WX and Google test headers. Other clang builds may get these as well.  With this PR, there are no additional errors in pwsafe code.
[ClangWextraWarnings.txt](https://github.com/user-attachments/files/30070007/ClangWextraWarnings.txt)

Tested with:
macOS M1Max 15.7.7 Sequoia, Xcode 16.4, wx 3.2.10 (local build)

dskrvk:
But gtest and wxwidgets should have their own makefiles and settings, so this shouldn't block building those locally? FWIW, with gtest it seems the problem is at the place of use, not with the code itself: https://github.com/google/googletest/blob/a25f43576effe4ebe887412a603cddfa8fc3ba64/googletest/include/gtest/gtest.h#L1394-L1404

nobugshere:
> with gtest it seems the problem is at the place of use

Ah, yes.  I see that now.  I'll keep looking into it.  I see no reason to block this PR.

PR #1869, opened 2026-07-17
[wx] Fix a number of signed to unsigned comparisons

nobugshere:
These warnings show-up in a template class used by the EXPECT_EQ macro, but only in Xcode 16.4 (clang 17) with the -Wextra option:

`/opt/homebrew/include/gtest/gtest.h:1394:11: warning: comparison of integers of different signs: 'const unsigned long' and 'const int' [-Wsign-compare]
`

In all cases, the 'int' was an integer constant, so I just added the 'u' suffix.
Thanks to @dskrvk for the pointer.


dskrvk (approved):
Nice!

ronys:
Thanks.

Testing this, I've found a few more warnings compiling on Windows for x86, will create a separate PR.

PR #1875, opened 2026-07-18
Fedora release

carlbyington2:
Password safe is now in the Fedora repositories. However, the 1.25.0 build in S390 arch needs a patch, since the S390 C++ does not allow does not allow variable length arrays.

### DIFF
--- a/README.LINUX.md
+++ b/README.LINUX.md
@@ -23,7 +23,7 @@ To install it just use the following command.
 $ sudo apt install passwordsafe
 ```
 
-or 
+or
 
 1. Download the .deb file that corresponds to your distribution.
 2. Install it using dpkg:
@@ -45,8 +45,17 @@ To install it just use the following command.
    $ sudo dnf install passwordsafe
    ```
 
+or
+
+1. Download the .rpm file that corresponds to your distribution.
+2. Install it using dnf:
+
+   ```
+   $ sudo dnf install passwordsafe-*.rpm
+   ```
+
 ## Installation on Gentoo
-As usual there are USE flags to control the features of the package. 
+As usual there are USE flags to control the features of the package.
 On Gentoo, suport for Yubi keys and QR is disabled by default.
 
 ```
@@ -71,7 +80,7 @@ Finally, run
 ```
 $ sudo pacman -U passwordsafe-*.zst
 ```
-to install the package on your machine. 
+to install the package on your machine.
 
 For more details on building and installing packages on Arch, see https://wiki.archlinux.org/title/Arch_User_Repository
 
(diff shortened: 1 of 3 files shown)


README.LINUX.md
```diff
@@ -36,10 +37,12 @@ or
    ```
 
 ## Installation on Fedora
-1. Download the .rpm file that corresponds to your distribution.
```
rafaelx, line 39: How about expanding the description to include the ability to download and install the packages provided by the project maintainer? That way, both options would be covered, just like it's described for Debian.

README.LINUX.md
```diff
@@ -36,10 +37,12 @@ or
...
 ## Installation on Fedora
-1. Download the .rpm file that corresponds to your distribution.
-2. Install it using dnf:
+
+Password Safe is available as a package in the Fedora repositories.
+To install it just use the following command.
+
    ```
-   $ sudo dnf install passwordsafe-*.rpm
+   $ sudo dnf install passwordsafe
```
rafaelx, line 45: What about package "passwordsafe-doc"?

src/core/ItemAtt.cpp
```diff
@@ -706,20 +706,15 @@ bool CItemAtt::DeSerializePlainText(const std::vector<char> &v)
...
 
-        memcpy(buf, &(*iter), len);
-        byteswap(buf, buf + len - 1);
-
-        if (!SetField(type, buf, len))
-          return false;
-        break;
+        unsigned char *begin = reinterpret_cast<unsigned char *>(&(*iter));
+        unsigned char *end   = begin + len - 1;
+        byteswap(begin, end);
```
dskrvk: Are we now modifying the input vector in place instead of copying it?

carlbyington2: aargh - I missed that the input is const. I will redo that.

carlbyington2: I rewrote the commit with forced push, copying the bytes to be swapped into a new std::vector.

dskrvk (approved):

README.LINUX.md
carlbyington2: The fedora packages at https://github.com/pwsafe/pwsafe/releases don't follow the requirements for fedora packages. In particular, those packages try to own directories that are owned by filesystem (/usr/share/locale/hu, /usr/share/locale/hu/LC_MESSAGES, etc).  Those packages also don't include the license files.

README.LINUX.md
carlbyington2: The main passwordsafe package "recommends" passwordsafe-doc, so by default installing the main one will also install the -doc subpackage. I just added that in the 1.25.0 package which is pending a push to updates-testing. In a few days it should be available for users.

README.LINUX.md
rafaelx: Would it be an option to extend the CMake script to produce a Fedora-compliant package?

carlbyington2: Probably, but I am not at all familiar with CMake. At a minimum the build machine would need to run Fedora with the fedora-packager package installed.

rafaelx: If you are open to it, you could publish your spec file here so that someone else can make the necessary changes to the CMake script.

ronys:
Thanks for the fixes to both DeSerializePlainText() implementations.
These are architecture-specific, not directly related to Fedora.

I'd either remove the proposed changes to README.LINUX.md or modify them to describe both options, including the issues raised in the comments. This can be amended if/when we get cpack to create a compliant RPM package.

README.LINUX.md
carlbyington2: Hm, I cannot get github to attach a file here, so 

https://www.five-ten-sg.com/util/passwordsafe.spec
https://www.five-ten-sg.com/util/changelog

PR #1876, opened 2026-07-18
[macOS] memory leaks in QREncode.mm

nobugshere:
Fix a couple memory leaks detected by the clang analyzer and confirmed with Xcode instruments.

macOS M1Max 15.7.7 Sequoia, Xcode 16.4, wx 3.2.10 (local build)


dskrvk (commented):
Ugh, it's so easy to make these mistakes. Isn't there some kind of RAII pattern for objC?

dskrvk (approved):

PR #1885, opened 2026-07-22
[wx] Fix help build dependencies

### LINKED ISSUE #1881: [wx] Some native packages contain outdated Help files
Issue:
Some native packages (wx version 1.25) contain outdated Help files (text and screenshots).
The Flatpak version is up to date because it is built directly from the source tree.

This primarily affects the English and Slovak Help, which have received numerous updates in recent months.
The packages with the most outdated Help are for Linux Mint, Debian 12 and 13, Fedora 43, and Ubuntu 24 (Slovak Help). The remaining packages are approximately 99% up to date.

I don't think this requires immediate action. Instead, it is something to look into for the next release.

ronys:
Resolves #1881

wx help files (zip) will now be built whenever a source help file is newer than the zip file instead of only when the zip file wasn't present in the build/help directory.

### DIFF
--- a/help/CMakeLists.txt
+++ b/help/CMakeLists.txt
@@ -45,23 +45,20 @@ else (WIN32 AND NOT WX_WINDOWS)
   function(make_help OUTPUT_FILE_VAR DIR LANG)
     set(OUTPUT_FILE "${CMAKE_CURRENT_BINARY_DIR}/help${LANG}.zip")
 
+    set(INPUT_DIR "${DIR}")
+    # If a helpfile directory exists for wxWidgets (ending with _wx), use it for
+    # generating the zip file, otherwise use the same one that's used for Windows.
     if(IS_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/${DIR}_wx")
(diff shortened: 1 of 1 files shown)


help/CMakeLists.txt
```diff
@@ -43,18 +43,20 @@ if (WIN32 AND NOT WX_WINDOWS)
 
 else (WIN32 AND NOT WX_WINDOWS)
   function(make_help OUTPUT_FILE_VAR DIR LANG)
-    set_lang_help_files(INPUT_FILES "${DIR}")
-
     set(OUTPUT_FILE "${CMAKE_CURRENT_BINARY_DIR}/help${LANG}.zip")
 
     if(IS_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/${DIR}_wx")
```
dskrvk, line 51: The only difference between the if branches seems to be the help output dir - maybe we can just use a variable and combine the branches?

ronys: Good point - implemented. Thanks!

dskrvk (approved):

PR #1886, opened 2026-07-22
[wx] Sorting by date in grid view does not work on macOS

nobugshere:
Sorting the grid view by clicking on any date column has no effect.  Running in the debugger, I get a number of "Failed to parse item's date string:..." messages.  The crux of the problem seems to be that the strftime() %c format looks like this (with the year at the end):
`Wed Jul 22 10:29:44 2026`
This appears to be the one format that wxDateTIme::ParseDateTime() cannot parse.  ~This PR proposes to change the format string to look as follows~, which is more consistent with the %c result on Linux:
`Wed 22 Jul 2026 10:32:04 PDT`

~This affects the format in a few other places, such as the "Dates and Times" tab on the add/edit dialog and the password history, but that seems acceptable.~  An alternative might be to change the grid view to use the TMC_LOCALE_SIMPLIFIED ("%F %T"), which would be more consistent with Windows, but would be a change to all WX platforms.

Caveats:
1. I suspect the *BSDs may suffer the same issue, but I don't have an environment set-up to test them.
2. I don't know if the change in core will impact @huven project.  If so, I can use a different define symbol.

Tested on:
macOS M1Max 15.7.7 Sequoia, Xcode 16.4, wx 3.2.10, 3.2.11
Fedora 44 ARM64 VMware,  KDE/Wayland,   gcc 16.1.1, wx 3.2.9, GTK 3.24.52

### DIFF
--- a/src/ui/wxWidgets/GridCtrl.cpp
+++ b/src/ui/wxWidgets/GridCtrl.cpp
@@ -749,10 +749,12 @@ void GridCtrl::RearrangeItemsDateTimeBased(ItemsCollection& collection, int colu
       // because localization is not taken into account and day-month detection algorithm is weak.
       // (e.g. 01.02.2026 -> Day=02 & Month=01; 02.01.2026 -> Day=01, Month=02)
       // See GitHub: https://github.com/wxWidgets/wxWidgets/issues/3049
+      // 2026-07-23 update: Now using YYYY-MM-DD format, so ParseDate() should be good.
       else if (dt.ParseDate(datetime, &end)) {
         collection.insert(std::pair<wxDateTime, const CItemData*>(dt, GetItem(row)));
-        pws_os::Trace(L"Warning - Sorting may be incorrect for date string: %ls ; day: %d ; month: %d ; year: %d",
-          datetime.wc_str(), dt.GetDay(), dt.GetMonth(), dt.GetYear());
+        // Note: GetMonth() return value range is 0..11
+        // pws_os::Trace(L"Warning - Sorting may be incorrect for date string: %ls ; day: %d ; month: %d ; year: %d",
+        //   datetime.wc_str(), dt.GetDay(), dt.GetMonth()+1, dt.GetYear());
       }
     }
     else {
--- a/src/ui/wxWidgets/GridCtrl.cpp
+++ b/src/ui/wxWidgets/GridCtrl.cpp
@@ -752,9 +752,6 @@ void GridCtrl::RearrangeItemsDateTimeBased(ItemsCollection& collection, int colu
       // 2026-07-23 update: Now using YYYY-MM-DD format, so ParseDate() should be good.
       else if (dt.ParseDate(datetime, &end)) {
         collection.insert(std::pair<wxDateTime, const CItemData*>(dt, GetItem(row)));
-        // Note: GetMonth() return value range is 0..11
-        // pws_os::Trace(L"Warning - Sorting may be incorrect for date string: %ls ; day: %d ; month: %d ; year: %d",
-        //   datetime.wc_str(), dt.GetDay(), dt.GetMonth()+1, dt.GetYear());
       }
     }
     else {
--- a/src/ui/wxWidgets/GridCtrl.cpp
+++ b/src/ui/wxWidgets/GridCtrl.cpp
@@ -713,12 +713,15 @@ template<typename ItemsCollection>
 void GridCtrl::RearrangeItemsDateTimeBased(ItemsCollection& collection, int column, bool dateOnly)
 {
   int row = 0;
-  struct tm tm;
   wxDateTime dt = wxDateTime::Today(); // initialize with valid date
   wxString datetime;
-  wxString dateFormat = wxGetApp().GetLocaleShortDateFormat();
   wxString::const_iterator end;
 
+  /*
+   This version of the code expects the date and time in the following format:
+   "YYYY-MM-DD HH:MM:SS" which is equivalent to "%F %T" format in strftime()
+   and is compatible with ISO 8601
+   */
   for (row = 0; row < GetNumberRows(); row++) {
     /*
       The items's date-time or date information.
@@ -740,19 +743,8 @@ void GridCtrl::RearrangeItemsDateTimeBased(ItemsCollection& collection, int colu
     /*
       The case if a date without a time is expected.
     */
-    else if (dateOnly) {
-      memset(&tm, 0, sizeof(tm)); // init / reset
-      if (!dateFormat.IsEmpty() && strptime(datetime.c_str(), dateFormat.c_str(), &tm)) {
-        collection.insert(std::pair<wxDateTime, const CItemData*>(wxDateTime(tm), GetItem(row)));
-      }
-      // Fallback to buggy version where day and month are in some cases swapped,
-      // because localization is not taken into account and day-month detection algorithm is weak.
-      // (e.g. 01.02.2026 -> Day=02 & Month=01; 02.01.2026 -> Day=01, Month=02)
-      // See GitHub: https://github.com/wxWidgets/wxWidgets/issues/3049
-      // 2026-07-23 update: Now using YYYY-MM-DD format, so ParseDate() should be good.
-      else if (dt.ParseDate(datetime, &end)) {
-        collection.insert(std::pair<wxDateTime, const CItemData*>(dt, GetItem(row)));
-      }
+    else if (dateOnly && dt.ParseDate(datetime, &end)) {
+      collection.insert(std::pair<wxDateTime, const CItemData*>(dt, GetItem(row)));
     }
     else {
       pws_os::Trace(L"Failed to parse item's date string: %ls ; using: %ls", datetime.wc_str(), wxDateTime::Today().FormatISOCombined().wc_str());
--- a/src/core/ItemData.h
+++ b/src/core/ItemData.h
@@ -162,6 +162,7 @@ public:
   StringX GetXTimeXML() const {return GetTime(XTIME, PWSUtil::TMC_XML);}  // V30
   StringX GetPMTimeXML() const {return GetTime(PMTIME, PWSUtil::TMC_XML);}  // V30
   StringX GetRMTimeXML() const {return GetTime(RMTIME, PWSUtil::TMC_XML);}  // V30
+  StringX GetTime(int whichtime, PWSUtil::TMC result_format, bool convert_epoch = false, bool utc_time = false) const; // V30
   //  These populate (and return) time_t instead of giving a character string
   time_t GetATime(time_t &t) const {CItem::GetTime(ATIME, t); return t;}  // V30
   time_t GetCTime(time_t &t) const {CItem::GetTime(CTIME, t); return t;}  // V30
@@ -422,9 +423,7 @@ private:
   EntryStatus m_entrystatus;
 
   // move from pre-2.0 name to post-2.0 title+user
-  void SplitName(const StringX &name,
-                 StringX &title, StringX &username);
-  StringX GetTime(int whichtime, PWSUtil::TMC result_format, bool convert_epoch = false, bool utc_time = false) const; // V30
+  void SplitName(const StringX &name, StringX &title, StringX &username);
   void SetTime(const int whichtime); // V30
   bool SetTime(const int whichtime, const stringT &time_str, bool utc_time = false); // V30
 
--- a/src/core/Util.cpp
+++ b/src/core/Util.cpp
@@ -547,11 +547,7 @@ StringX PWSUtil::ConvertToDateTimeString(const time_t &t, TMC result_format, boo
       is_error = !_tcsftime(datetime_str, sizeof(datetime_str) / sizeof(datetime_str[0]), _T("%Y-%m-%dT%H:%M:%S"), st);
       break;
     case TMC_LOCALE:
-#ifdef __APPLE__
-      is_error = !_tcsftime(datetime_str, sizeof(datetime_str) / sizeof(datetime_str[0]), _T("%a %d %b %Y %T %Z"), st);
-#else
       is_error = !_tcsftime(datetime_str, sizeof(datetime_str) / sizeof(datetime_str[0]), _T("%c"), st);
-#endif
       break;
     case TMC_LOCALE_SIMPLIFIED:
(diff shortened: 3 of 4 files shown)


huven:
> I don't know if the change in core will impact @huven project. If so, I can use a different define symbol.

No impact for my project, I use the raw time_t values, convert to NSDate and then use Foundation's date formatter to get locale-aware formatted strings.

dskrvk:
I'm not sure using a fixed (and locale-specific) format is the right approach when the caller is asking for a locale-dependent string. For example, the proposed format won't work well for American users.

The ideal approach would be:
1. For display purposes always format strings according to the user's locale (and consistently in different places).
2. For comparisons use a numeric date representation (or a well-ordered string format one like ISO-8601 which should be supported by most parsers).

A slightly worse approach would be to use a "middle ground" format like `YYYY-MM-DD HH:MM:SS` (in the user's timezone, so no need to show it).

ronys:
I agree with @dskrvk . Since the TMC_LOCALE / "%c" formatting is used wherever a date/time is displayed to the user, replacing it with a non-locale specific format is much more noticeable than for the less frequently used grid view.

So switching to TMC_LOCALE_SIMPLIFIED ("%F %T") for the grid view seems the lesser evil here, I think.

rafaelx:
Could the different date-time format be related to the environments locale? This might potentially affect the use of `setlocal` in `PWSafeApp::OnInit`. See also the [discussion](https://stackoverflow.com/questions/45867378/unexpected-strftime-behaviour-on-python-and-macos) on Stack Overflow.

`PWSafeApp::OnInit` already includes some MacOS specific logic.
```
bool PWSafeApp::OnInit()
{
  // Get the locale environment variable 'LC_CTYPE' specified by the environment
  // For instance, the behavior of function 'wcstombs' depends on the LC_CTYPE 
  // category of the selected C locale.
#if defined(__WXMAC__)
  const char *lcCType = setlocale(LC_CTYPE, NULL);

  if (lcCType && strcmp(lcCType, "UTF-8")) { // MAC OS only have UTF-8 as default, but conversion with this string is not running well
    setlocale(LC_CTYPE, "");
  }
  else {
    const char *env_lc_cType = std::getenv("LC_CTYPE");
    
    if(env_lc_cType) {
      setlocale(LC_CTYPE, env_lc_cType);
    }
    else {
      setlocale(LC_CTYPE, "en_US.UTF-8");
    }
  }
  // This value must be set for mac OS starting with version 11, but is no problem for earlier versions, see
  // https://trac.wxwidgets.org/ticket/19023
  setlocale(LC_NUMERIC, "C");
#else
  setlocale(LC_CTYPE, "");
#endif
```

nobugshere:
@dskrvk 
> 1. For display purposes always format strings according to the user's locale (and consistently in different places).

I think %c is meant to do that. When I try different locals on the Mac, the day and month get translated, but the year-last layout is pretty common.  The %a and %b specifiers are also translated.

> 2. For comparisons use a numeric date representation (or a well-ordered string format one like ISO-8601 which should be supported by most parsers).

Yes.  I intend for any comparisons to remain consistent, even if the exact string format used is different.

@rafaelx 
Locale definitely makes a difference.  However, on macOS, C and en_US both use the "year-last" format; en_GB also uses it, but swaps month-day.  I think that format is an artifact of old Unix/C conventions.  On Linux, C is year-last but en_US and en_GB are both as shown above.  

I actually started looking at the locale code because the behavior changes slightly in wx3.3.  I'm trying to figure-out a "right" way to do it without breaking anything.  That's when I noticed the column sorting wasn't working at all.

@ronys 

> So switching to TMC_LOCALE_SIMPLIFIED ("%F %T") for the grid view seems the lesser evil here, I think.

The more I think about it the more I agree.  The latest commit removes my previous change and changes GridTable.cpp to fetch the simplified format.  In this form, it applies to all WX builds, but can easily use a different format on different builds.  This is similar to, but different than, the Windows format, which looks like "%x %T"


ISO 8601 ...FTW!

src/ui/wxWidgets/GridCtrl.cpp
```diff
@@ -749,10 +749,12 @@ void GridCtrl::RearrangeItemsDateTimeBased(ItemsCollection& collection, int colu
       // because localization is not taken into account and day-month detection algorithm is weak.
       // (e.g. 01.02.2026 -> Day=02 & Month=01; 02.01.2026 -> Day=01, Month=02)
       // See GitHub: https://github.com/wxWidgets/wxWidgets/issues/3049
+      // 2026-07-23 update: Now using YYYY-MM-DD format, so ParseDate() should be good.
       else if (dt.ParseDate(datetime, &end)) {
         collection.insert(std::pair<wxDateTime, const CItemData*>(dt, GetItem(row)));
-        pws_os::Trace(L"Warning - Sorting may be incorrect for date string: %ls ; day: %d ; month: %d ; year: %d",
-          datetime.wc_str(), dt.GetDay(), dt.GetMonth(), dt.GetYear());
+        // Note: GetMonth() return value range is 0..11
```
ronys: Why is this comment relevant?

src/ui/wxWidgets/GridCtrl.cpp
```diff
@@ -749,10 +749,12 @@ void GridCtrl::RearrangeItemsDateTimeBased(ItemsCollection& collection, int colu
...
+        // pws_os::Trace(L"Warning - Sorting may be incorrect for date string: %ls ; day: %d ; month: %d ; year: %d",
```
ronys: Please avoid commenting out code. If it's not relevant, then it should be removed.

src/ui/wxWidgets/GridCtrl.cpp
nobugshere: Removed

src/ui/wxWidgets/GridCtrl.cpp
nobugshere: Removed

src/ui/wxWidgets/GridCtrl.cpp
```diff
@@ -749,10 +749,9 @@ void GridCtrl::RearrangeItemsDateTimeBased(ItemsCollection& collection, int colu
       // because localization is not taken into account and day-month detection algorithm is weak.
       // (e.g. 01.02.2026 -> Day=02 & Month=01; 02.01.2026 -> Day=01, Month=02)
       // See GitHub: https://github.com/wxWidgets/wxWidgets/issues/3049
+      // 2026-07-23 update: Now using YYYY-MM-DD format, so ParseDate() should be good.
```
dskrvk: I'm confused by what the comments (old and new) are telling me. If `ParseDate` on the next line works fine now, do we really need this long explanation anymore?

src/ui/wxWidgets/GridCtrl.cpp
dskrvk: Nit: `tm` could simply be scoped to this else block, with no need to re-init manually.

dskrvk (approved):

src/ui/wxWidgets/GridCtrl.cpp
nobugshere: I thought it might be useful to keep the history, but if it's confusing I'll remove it.

src/ui/wxWidgets/GridCtrl.cpp
nobugshere: Looking more closely at that part of the code, I realized that now that it expects YYYY-MM-DD, it can be simplified.

src/ui/wxWidgets/GridCtrl.cpp
dskrvk: The history is in the git log :) It can be looked up when needed, but most of the time people just care about the current state.

src/ui/wxWidgets/GridCtrl.cpp
```diff
@@ -713,12 +713,15 @@ template<typename ItemsCollection>
 void GridCtrl::RearrangeItemsDateTimeBased(ItemsCollection& collection, int column, bool dateOnly)
 {
   int row = 0;
-  struct tm tm;
   wxDateTime dt = wxDateTime::Today(); // initialize with valid date
   wxString datetime;
-  wxString dateFormat = wxGetApp().GetLocaleShortDateFormat();
```
rafaelx, line 719: Does `dateFormat` show the correct formatting string for the format with the timestamp in the middle of the date format?

If so, then maybe my suggestion in the next comment could be a solution?!

src/ui/wxWidgets/GridCtrl.cpp
rafaelx, line 740: Is this the reason for the error message `Failed to parse item's date string...`?

If `ParseDateTime` has problems with the date and time format on MacOS, I would suggest using the solution with `strptime` here as well, just like in the `else if (dateOnly)` branch. `strptime` should work much more reliably, so the fallback solution with `ParseDateTime` shouldn't really be necessary. I thought, however, that if `strptime` fails, maybe `ParseDateTime` could still save something. Since the algorithm of `ParseDateTime` seems to be very simple, you probably shouldn't expect any reliable results from this function if `strptime` already hasn't worked.

src/ui/wxWidgets/GridCtrl.cpp
```diff
@@ -740,20 +743,8 @@ void GridCtrl::RearrangeItemsDateTimeBased(ItemsCollection& collection, int colu
...
     */
-    else if (dateOnly) {
-      memset(&tm, 0, sizeof(tm)); // init / reset
-      if (!dateFormat.IsEmpty() && strptime(datetime.c_str(), dateFormat.c_str(), &tm)) {
-        collection.insert(std::pair<wxDateTime, const CItemData*>(wxDateTime(tm), GetItem(row)));
-      }
-      // Fallback to buggy version where day and month are in some cases swapped,
-      // because localization is not taken into account and day-month detection algorithm is weak.
-      // (e.g. 01.02.2026 -> Day=02 & Month=01; 02.01.2026 -> Day=01, Month=02)
-      // See GitHub: https://github.com/wxWidgets/wxWidgets/issues/3049
```
rafaelx, line 751: While working on date-time-based sorting, I noticed issues with `ParseDateTime`. `ParseDateTime` does not rely on a native, mature implementation but rather on a custom algorithm that does not take the locale into account. Given that `pwsafe` uses the Wx framework, it makes sense to leverage Wx's built-in mechanisms. However, since `ParseDateTime` does not behave as expected, I added the comment to alert other developers so they might avoid using this function or at least be aware of the issue.

src/ui/wxWidgets/GridCtrl.cpp
nobugshere: Given the example in the original post, and single stepping through the original code, I found that ParseDateTime calls ParseDate, which would take 10 for the year, then it computes the day of the week as a check (in the year 10, or maybe it assumes it's 1910, I'm not sure) and fails because it wasn't a Wednesday.  I haven't looked into strptime, perhaps if it can use %c for parsing (because that's what is used to produce the string) it might adhere to the same platform/locale conventions.

rafaelx:
Maybe my suggestion isn’t entirely clear from my previous comments, or I just haven’t fully understood the problem yet, which is also possible.

My idea for determining the date-time information without using `wxDateTime::ParseDateTime` is as follows.

We have `PWSafeApp::GetLocaleShortDateFormat` to get the date format depending on the locale. Now, another method could be added to the class that provides the date-time format. For example, `PWSafeApp::GetLocaleDateTimeFormat` using [`wxLocale::GetInfo`](https://docs.wxwidgets.org/3.3/classwx_locale.html#acd6eccc8900847c0a29e7a4598c7d83f) and the [parameter](https://docs.wxwidgets.org/3.3/intl_8h.html#a54d3421c44d428cc5cce0c772d583c34) `wxLocaleInfo::wxLOCALE_LONG_DATE_FMT` or `wxLocaleInfo::wxLOCALE_DATE_TIME_FMT` (this would need to be determined). This method should then hopefully provide the correct date-time formatting, which can then be passed to [`strptime`](https://man7.org/linux/man-pages/man3/strptime.3.html) along with the string to be parsed. `strptime` should then be able to extract the date and time information using the correct format string.

nobugshere:
> Maybe my suggestion isn’t entirely clear from my previous comments, or I just haven’t fully understood the problem yet, which is also possible.

Your suggestion was clear, I've just been busy.  I've done some more experimenting...

Part of the problem (on macOS) is that PWSafeApp::GetLocaleDateTimeFormat does not return the correct sequence; it gives me d/m/y, even though I'm using en_US and the date is displayed in m/d/y order.  

I believe there is confusion between US and GB locales.  I had that problem before; See PR #1375 for code I wrote that works on the Mac, at least as far as I can test.  (I recently noticed that it is needed on FreeBSD as well.)  The original problem was that if you pass "en" to wxLocale it guesses en_GB, which I try to correct.  I now realize that PWSafeApp::m_locale is still wrong, so that might be why PWSafeApp::GetLocaleDateTimeFormat and ParseDateTime don't work for me.

As an experiment, if I hard-code GridCtrl::RearrangeItemsDateTimeBased to use strptime() with %c and %x it all works right on macOS, which makes sense since that's how the strings were produced with strftime.  Buuuuut, Linux insists on displaying the date in d/m/y format (wrong for en_US), so when I run that build on Linux, the date-only column doesn't sort correctly.  (Again, maybe m_locale is at fault.)

To top it all off, I've already noticed a new date ordering problem with wx3.3.x, they have made some changes to the locale handling and don't recommend the use of wxLocale on macOS.  (use wxUILocale instead.)  

So, at this point, I'm rather tired and confused :-). I think the way to proceed is to make minimal changes to fix my original sorting problems on macOS, preferably without breaking Linux.  Then have a larger re-think of how best to handle locale with wx3.2 and 3.3, across platforms and locales.  Maybe I'll start a discussion thread to talk about it.

src/ui/wxWidgets/GridCtrl.cpp
dskrvk, line 735: I think we can fairly easily skip date parsing altogether and deal only with `time_t`. Here `GetItem(row)` gives us the item, and we just need to define a method similar to `CItemData::GetTime`, but that doesn't format the time field.

nobugshere: > I think we can fairly easily skip date parsing altogether and deal only with `time_t`. 

Indeed!  That's so much simpler.

nobugshere:
I'm closing this PR in favor of a much simplified approach in PR #1899.

PR #1888, opened 2026-07-26
Pass parameters as const reference where possible for efficiency

dskrvk:
...to avoid object copying


src/ui/wxWidgets/AddEditPropSheetDlg.cpp
```diff
@@ -1691,7 +1691,7 @@ struct newer {
   }
 };
 
-void AddEditPropSheetDlg::SetupDCAComboBoxes(wxComboBox *pcbox, short &iDCA, bool isShift)
+void AddEditPropSheetDlg::SetupDCAComboBoxes(wxComboBox *pcbox, const short &iDCA, bool isShift)
```
ronys: iDCA should be just a short, not a const reference.

src/ui/wxWidgets/AddEditPropSheetDlg.h
```diff
@@ -340,9 +340,9 @@ class AddEditPropSheetDlg : public wxPropertySheetDialog
   void ApplyFontPreferences();
 
   void ItemFieldsToPropSheet();
-  void SetupDCAComboBoxes(wxComboBox *pcbox, short &iDCA, bool isShift);
+  void SetupDCAComboBoxes(wxComboBox *pcbox, const short &iDCA, bool isShift);
```
ronys: iDCA should be just a short, not a const reference.

src/ui/wxWidgets/DnDPWSafeObject.h
```diff
@@ -59,7 +59,7 @@ class DnDPWSafeObject : public wxDataObjectSimple
     // we need to copy the Object because the one we're handled may be
     // deleted while it's still on the clipboard (for example) - and we
     // reuse the serialisation methods here to copy it
-  DnDPWSafeObject(wxMemoryBuffer *object = (wxMemoryBuffer *) nullptr);
+  DnDPWSafeObject(const wxMemoryBuffer *object = (wxMemoryBuffer *) nullptr);
```
ronys: Shouldn't this be `object = nullptr`? Seems odd to be casting to a non-const.

src/ui/wxWidgets/TreeCtrl.h
```diff
@@ -136,7 +136,7 @@ class TreeCtrlBase : public wxTreeCtrl
   void SetShowGroup(bool v) { m_show_group = v; }
   bool IsShowGroup() const { return m_show_group; }
   
-  void SetSorting(TreeSortType &v) { m_sort = v; }
+  void SetSorting(const TreeSortType &v) { m_sort = v; }
```
ronys: `TreeSortType` is an enum, so just passing it by value is clearer, I think.

src/ui/wxWidgets/AddEditPropSheetDlg.cpp
dskrvk: Fixed

src/ui/wxWidgets/AddEditPropSheetDlg.h
dskrvk: Fixed

src/ui/wxWidgets/DnDPWSafeObject.h
dskrvk: Right, fixed

src/ui/wxWidgets/TreeCtrl.h
dskrvk: Fixed

PR #1893, opened 2026-07-28
Restore alt n functionality

### LINKED ISSUE #1892: Accessibility: Restore Alt-N accelerator
In the Basic tab of the add/edit dialog, the switch to subtabs for Notes and Custom Fields broke the Alt-N accelerator functionality. This should be restored Alt-N (or whatever is defined as the Notes accelerator key in the current language) should switch the keyboard focus to the Notes tab.

ronys:
Resolves #1892 and #1898


src/ui/Windows/AddEdit_Basic_SubPage.cpp
```diff
@@ -29,3 +51,20 @@ void CAddEdit_Basic_SubPage::NotifyChanged()
...
+
+BOOL CAddEdit_Basic_SubPage::PreTranslateMessage(MSG *pMsg)
+{
+  if (pMsg->message == WM_SYSCHAR &&
+      !HasLocalMnemonic(this, static_cast<wchar_t>(pMsg->wParam))) {
+    // None of this page's own controls are a mnemonic for this key, so this
+    // page has nothing useful to do with it. Offer it to our parent directly
+    // instead of just leaving it alone for the framework's own
+    // PreTranslateMessage tree-walk to (maybe) carry further up: that walk
+    // does not reliably reach past this page for every kind of hosted control.
```
dskrvk, line 96: Is it because the various UI components don't always implement command bubbling correctly (which is outside of our control) or because we're using them wrong somehow? I'm not a Windows expert so could be way off track, but seems like we're hacking around something that should be guaranteed by the framework.

src/ui/Windows/AddEdit_Basic_Tabs.cpp
```diff
@@ -96,3 +99,45 @@ bool CAddEdit_Basic_Tabs::IsExternalEditorActive() const
...
+      wchar_t szTabText[256];
+      TCITEM tci = {};
+      tci.mask = TCIF_TEXT;
+      tci.pszText = szTabText;
+      tci.cchTextMax = _countof(szTabText);
+
+      if (pTabCtrl->GetItem(iNotes, &tci)) {
+        const wchar_t mnemonic = WinUtil::GetMnemonicChar(szTabText);
+        if (mnemonic != 0 &&
+            mnemonic == static_cast<wchar_t>(std::towupper(static_cast<wchar_t>(pMsg->wParam)))) {
```
dskrvk: This accelerator button handling in particular looks like it should've been implemented by the tab UI component (kind of surprising that it doesn't do this).

ronys:
Both good points, but what we're doing here is slightly non-standard. The tabs aren't dialog controls that are candidates targets for accelerator keys. We also need to map from the tab to the Note text control itself.
Finally, to get Alt-N to work when the focus is in the list control of the sibling "Custom Fields" tab requires non-standard event handling.

src/ui/Windows/AddEdit_Basic_Tabs.cpp
```diff
@@ -96,3 +99,48 @@ bool CAddEdit_Basic_Tabs::IsExternalEditorActive() const
...
+{
+  // The single home for every subtab's accelerator: both CAddEdit_Basic (when
+  // one of its fields has focus) and our subtabs (via
+  // CAddEdit_Basic_SubPage, when they don't own the accelerator themselves)
+  // route WM_SYSCHAR here.
+  //
+  // Each accelerator letter is marked with a leading '&' in the corresponding
+  // subtab's caption, as read from the tab control for proper I18N support.
+  if (pMsg->message == WM_SYSCHAR) {
+    CAddEdit_Basic_SubPage *const pages[] = { &m_pp_notes, &m_pp_customFields };
```
dskrvk: Is there a way to list the actual child pages instead of hard-coding them?

src/ui/Windows/AddEdit_Basic_Tabs.cpp
```diff
@@ -96,3 +99,48 @@ bool CAddEdit_Basic_Tabs::IsExternalEditorActive() const
...
+      for (CAddEdit_Basic_SubPage *pPage : pages) {
+        const int iPage = GetPageIndex(pPage);
+        if (iPage < 0)
+          continue;
+
+        wchar_t szTabText[256];
+        TCITEM tci = {};
+        tci.mask = TCIF_TEXT;
+        tci.pszText = szTabText;
+        tci.cchTextMax = _countof(szTabText);
```
dskrvk, line 134: We assume that the accelerator mnemonic will be mentioned somewhere at the beginning of the text, so even if it happens to be longer than 256 we won't miss the mnemonic by using a max length? Seems reasonable.

src/ui/Windows/AddEdit_Basic_SubPage.h
```diff
@@ -27,7 +27,20 @@ class CAddEdit_Basic_SubPage : public CPWPropertyPage
   // has finished, and never for a view-only or protected entry.
   void NotifyChanged();
 
+  // Called by CAddEdit_Basic_Tabs right after this page becomes the active tab
+  // via its mnemonic accelerator, so the page can put keyboard focus wherever
+  // makes sense for it (e.g. an edit control, or a list vs. a button depending
+  // on whether the list has any rows).
+  virtual void FocusDefaultControl() = 0;
```
dskrvk, line 34: Can we maybe define an abstract class containing these two methods, to specify the minimal interface required to implement an accessible control? We can't force all future custom controls to be accessible, but we can at least guide the implementers.

ronys: I think this is overengineering, at least for now.
Having this as an abstract method, along with the preceding comment, seems enough for me.
This is also a pretty specific scenario, where MFC's default focus mechanism isn't sufficient.

src/ui/Windows/AddEdit_Basic_CustomFieldsPage.cpp
```diff
@@ -264,6 +278,46 @@ void CAddEdit_Basic_CustomFieldsPage::OnCustomFieldsKeyDown(NMHDR *pNMHDR, LRESU
...
+      default:
+        break;
+    }
+
+    if (next >= 0) {
+      m_customFieldsList.SetItemState(-1, 0, LVIS_SELECTED | LVIS_FOCUSED);
+      m_customFieldsList.SetItemState(next, LVIS_SELECTED | LVIS_FOCUSED,
+                                      LVIS_SELECTED | LVIS_FOCUSED);
+      m_customFieldsList.SetSelectionMark(next);
+      m_customFieldsList.EnsureVisible(next, FALSE);
```
dskrvk: This sequence of 4 calls now appears thrice, would it make sense to define a helper method?

dskrvk (approved):

Issue #1702, opened 2026-02-07
Wayland Support


hdlguy:
Ubuntu 24.04 LTS uses Wayland by default.  The ubuntu package passwordsafe does not run under Wayland.

froggie2523:
It does not run at all, comes with some error message or some feature is not working?

dskrvk:
I have popOS 24.04 with Wayland, and an old locally-built pwsafe 1.17 works mostly fine. There are some UI quirks like some controls/windows have the wrong size, but nothing that's blocking (I've been meaning to log some of those issues). @hdlguy what issues are you seeing?

Issue #1336, opened 2024-07-12
Password Safe does not recognizes system tray, but allows to select options anyway


igor2x:
Password Safe v1.19.1 flatpak on Ubuntu 24.04.

1. Start Password Safe and Master Passwrord Entry make sure "Open as read-only" is unckecked.
2. Manage | Options | System and in System Tray section there is red warning displayed that it appears to be no system tray support.
3. Check the "Put icon in System Tray" and OK button on bottom-right.
4. Click on Lock icon from toolbar.
5. I can see no icon in system tray and also Alt+Tab and there is no Password Safe on the list. Also Password Safe also does not have dot displayed on Ubuntu's Dash icon indicating program is started.
6. Click on Dash Password Safe's icon and new Password Safe is displayed.
7. At Master Password Entry make sure "Open as read-only" is unchecked and type in master password and OK button and error is displayed: "Remove Lock? Lock is done by yourself" with No/Yes buttons.

- Password Safe is still running in some background not visible. 
- Password Safe closed, but did not release a lock.

---

I was just playing around with different settings and forgot about this Sytem Tray checkbox and then after a while I clicked on Lock icon.

In my humble opinion if Password Safe recognizes that there is no system tray (or probably it is, but can't work with it) I think both of the checkbox options from System Trey section should get greyed out.

PR #1895, opened 2026-07-31
Linux: add SNI/AppIndicator tray icon support for Wayland

ykne:
## Problem

On native-Wayland GNOME (`GDK_BACKEND=wayland`), pwsafe's tray icon doesn't appear at all. wx's classic tray-icon path checks `wxGTKImpl::IsX11()`, which is unconditionally false under native Wayland — this isn't a pwsafe bug, it's a gap in wx's non-AppIndicator GTK3 backend.

## Fix

Two coordinated changes, since the tray icon logic lives in wx, not pwsafe:

1. **wxWidgets** (my fork, branches `appindicator-sni-taskbar` for master and `appindicator-sni-taskbar-3.2` for the 3.2 stable branch): adds a new `wxUSE_APPINDICATOR` backend to `src/gtk/taskbar.cpp` using AppIndicator/StatusNotifierItem, which works under both X11 and native Wayland. Also wires the library's `activate` signal to the existing `wxEVT_TASKBAR_LEFT_DCLICK` event, so double-click-to-restore keeps working exactly as before.
2. **pwsafe** (this PR): minimal `SystemTray` changes to use wx's new backend — no behavior changes on X11.

## Dependency

Full click parity (double-click restores, single/right-click show menu — matching pwsafe's existing UX) needs `libayatana-appindicator >= 0.6.0`, which added a real `Activate()` D-Bus method upstream in 2026-06. No patch needed there, just a version floor. Older versions still show the tray icon, just without the double-click distinction.

## Verified working, live, on GNOME/Wayland

- Native build (Fedora RPM, wx 3.2.12 + libayatana-appindicator 0.6.0): full click matrix (right-click, single-left, double-left) all correct.
- Flatpak build (manifest pointing at both forks + an appindicator 0.6.0 module + `--talk-name=org.kde.StatusNotifierWatcher`): same result, confirmed via `org.kde.StatusNotifierWatcher`'s `RegisteredStatusNotifierItems` and manual testing.
- Local dev build against the currently-open PR's exact commit: rebuilt clean, tray icon registers and click behavior confirmed.

## Known limitation

Double-click-to-restore causes a brief minimize/restore blink even when the window is already visible, under focus-follows-mouse WMs. Root-caused to Wayland not exposing minimize state to clients at all — no wx/GTK-visible property distinguishes "just lost focus" from "actually minimized." Documented in a comment at the fix site (`PasswordSafeFrame::UnlockSafe`); not fixable client-side as far as I can tell.

## Related issues

- #999 — Flatpak tray icon "vanished" under Wayland (closed as unreproducible at the time — this is the real cause)
- #1223 — tray enabled + broken under Wayland leaves the app completely unreachable (no Alt+Tab, no dash entry, no icon)
- #1336 — same symptom, flatpak-specific
- #1612 — doc workaround for this exact problem; superseded by this actual fix
- #1702 — general Wayland-support umbrella issue

## Upstream wxWidgets discussion

- wx PR (open, active review): https://github.com/wxWidgets/wxWidgets/pull/26715
- wx PR (earlier iteration, superseded): https://github.com/wxWidgets/wxWidgets/pull/26689

pwsafe couldn't depend on the wx change in a release until #26715 lands upstream, but there's no reason this pwsafe-side change can't be reviewed in parallel.


CMakeLists.txt
```diff
@@ -76,6 +76,38 @@ if (${CMAKE_SYSTEM_NAME} MATCHES "Linux")
   option (USE_ASAN "Set ON to add -fsanitize=address to Debug builds" OFF)
 endif (${CMAKE_SYSTEM_NAME} MATCHES "Linux")
 
+if (${CMAKE_SYSTEM_NAME} MATCHES "Linux")
```
dskrvk: This step adds a dependency on ImageMagick. Can we instead convert the files once and check them into source control? I don't think the XPMs change very often.

CMakeLists.txt
```diff
@@ -76,6 +76,38 @@ if (${CMAKE_SYSTEM_NAME} MATCHES "Linux")
...
+  find_program(IMAGEMAGICK_CMD NAMES magick convert)
+  if(IMAGEMAGICK_CMD)
+    set(TRAY_ICON_APPS_DIR "${CMAKE_BINARY_DIR}/tray-icons/hicolor/32x32/apps")
+    file(MAKE_DIRECTORY ${TRAY_ICON_APPS_DIR})
+    file(WRITE "${CMAKE_BINARY_DIR}/tray-icons/hicolor/index.theme"
+         "[Icon Theme]\nName=hicolor\nDirectories=32x32/apps\n\n[32x32/apps]\nSize=32\nType=Fixed\n")
```
dskrvk: Can we just keep this file in source control? As the cmake logic grows it gets harder to maintain and to find these embedded file fragments in it.

src/ui/wxWidgets/wxUtilities.h
```diff
@@ -480,6 +480,10 @@ namespace wxUtilities
 // on Fedora or Ubuntu
 bool IsTaskBarIconAvailable();
 
+// True if a Wayland session is actually falling back to the X11 GDK backend
+// (forced via GDK_BACKEND=x11, or the Flatpak "--socket=x11" case)
+bool IsXWaylandEnabled();
```
dskrvk: A cleaner design IMO would've been to expand `WhatWindowSystem` to return a new value for XWayland. In a previous PR it was decided that it's not ideal to have individual methods like `IsWayland`, `IsX`, and that's how `WhatWindowSystem` came about.

src/ui/wxWidgets/SystemTray.h
```diff
@@ -41,9 +41,18 @@ class SystemTray : public wxTaskBarIcon
...
+    //overridden from wxTaskBarIcon: supplies the persistent menu the
+    //AppIndicator/SNI backend keeps attached to the indicator; unlike
+    //CreatePopupMenu() this must not pop up a menu as a side effect,
+    //since it's invoked on every SetTrayStatus() call, not just clicks
+    virtual wxMenu* GetPopupMenu();
+#endif
     void ProcessSysTrayMenuItem(int itemId);
 
   private:
+    wxMenu* BuildMenu();
```
dskrvk:
```suggestion
    wxMenu* BuildPopupMenu();
```

src/ui/wxWidgets/SystemTray.h
```diff
@@ -41,9 +41,18 @@ class SystemTray : public wxTaskBarIcon
...
+    void PopulateMenu(wxMenu* menu);
```
dskrvk:
```suggestion
    void PopulatePopupMenu(wxMenu* menu);
```

src/ui/wxWidgets/SystemTray.cpp
```diff
@@ -115,10 +145,17 @@ void SystemTray::SetTrayStatus(TrayStatus status)
...
 {
   wxMenu* menu = new wxMenu;
+  PopulateMenu(menu);
+  return menu;
+}
+
+void SystemTray::PopulateMenu(wxMenu* menu)
+{
+  while (menu->GetMenuItemCount() > 0)
+    menu->Destroy(menu->FindItemByPosition(0));
```
dskrvk: IMO it's cleaner to keep this destroy logic in the only place that needs it, namely `GetPopupMenu`, so `PopulateMenu` remains focused on a single responsibility - to populate a new menu.

Alternatively, this method could check if `menu == nullptr` and then create it (so the caller doesn't have to), otherwise empty the provided one. We could also name it e.g. `RepopulateMenu` to make this idempotent behavior clearer. That would also let us get rid of `BuildMenu` (currently there a bit too many methods with similar names and unclear differences in behavior).

src/ui/wxWidgets/SystemTray.cpp
```diff
@@ -95,6 +106,25 @@ void SystemTray::SetTrayStatus(TrayStatus status)
   if (!IsTaskBarIconAvailable())
     return;
 
+#if wxUSE_APPINDICATOR
+  if (PWSprefs::GetInstance()->GetPref(PWSprefs::UseSystemTray)) {
+    const char* sniName = nullptr;
+    switch (status) {
+      case TrayStatus::LOCKED:
```
dskrvk: Why not the same 3 cases as in the switch below (CLOSED/UNLOCKED/LOCKED)?

dskrvk (approved):

CMakeLists.txt
dskrvk: Looks like #1896 did add `ImageMagick` to mock-build.sh - is it still needed there?

PR #1896, opened 2026-07-31
Misc: modernize mock-build.sh

ykne:
## Summary
- Drop CentOS-7/devtoolset-7 special-casing; target current Fedora only
- Add `-v WXVERSION` option to pin a specific installed wxGTK version instead of always using the latest, copying matching wx* RPMs into the mock chroot and force-installing them
- Clone from the current git remote/branch when run inside a repo, falling back to the upstream pwsafe.git otherwise
- Build with `cmake --build`/`cpack` directly instead of a fixed CentOS toolchain invocation; disable gtest in the mock build
- Add `libayatana-appindicator-gtk3-devel` to the installed build deps
- Mark the script executable

## Test plan
- Exercised the new features locally: ran the script both with and without `-v WXVERSION` to confirm version pinning picks up the matching wx* RPMs, and confirmed the git-remote-aware clone path picks up the current branch/remote when run from within a repo checkout.

### DIFF
--- a/Misc/mock-build.sh
+++ b/Misc/mock-build.sh
@@ -4,18 +4,76 @@
 # With no argument compiles for current system.
 # If added -r mock_arch, where mock_arch is a file name in /etc/mock without .cfg,
 # then rpm will be compiled on specified architecture.
-# Tested on Fedora-30 and CentOS-7
-# Build on or for CentOS-7 requires current mock* for devtoolset-7-gcc-c++ access.
-#
-# To install the resulting RPM on CentOS7, the https://fedoraproject.org/wiki/EPEL repository
-# should be installed.
+# Use -v WXVERSION to pin a specific wxGTK version instead of the latest.
+# Tested on Fedora
+
+usage() {
+    echo "Usage: $0 [-v WXVERSION] [mock options]" >&2
+    echo "  -v WXVERSION  Install this exact wxGTK version instead of the latest" >&2
+    exit 1
+}
+
+WANT_VERSION=""
+
+while getopts ":v:" opt; do
+    case $opt in
+        v) WANT_VERSION="$OPTARG" ;;
+        :) echo "ERROR: -$OPTARG requires an argument" >&2; usage ;;
+        \?) break ;;
+    esac
+done
+shift $((OPTIND - 1))
+
+CLONE_URL="https://github.com/pwsafe/pwsafe.git"
+CLONE_OPTS=""
+RPMDIR=$(rpm --eval '%_rpmdir')
+RPMARCH=$(rpm --eval '%_arch')
+DIST=$(rpm --eval '%{dist}' | sed 's/^\.//')
+
+if git rev-parse --is-inside-work-tree >/dev/null 2>&1; then
+    BRANCH=$(git rev-parse --abbrev-ref HEAD 2>/dev/null)
+    UPSTREAM=$(git rev-parse --abbrev-ref --symbolic-full-name @{u} 2>/dev/null)
+    if [ -n "$UPSTREAM" ]; then
+        REMOTE=$(echo "$UPSTREAM" | cut -d/ -f1)
+        CLONE_URL=$(git remote get-url "$REMOTE")
+        CLONE_OPTS="--branch $BRANCH --single-branch"
+    fi
+fi
 
 mock --init "$@"
-if [ "$2" = epel-7-x86_64 -o \( -z "$2" -a $(readlink /etc/mock/default.cfg) = epel-7-x86_64.cfg \) ];then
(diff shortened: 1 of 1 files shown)


Misc/mock-build.sh
```diff
@@ -4,18 +4,76 @@
...
+        CLONE_URL=$(git remote get-url "$REMOTE")
+        CLONE_OPTS="--branch $BRANCH --single-branch"
+    fi
+fi
 
 mock --init "$@"
-if [ "$2" = epel-7-x86_64 -o \( -z "$2" -a $(readlink /etc/mock/default.cfg) = epel-7-x86_64.cfg \) ];then
-    mock "$@" install cmake gcc-c++ git gtest-devel libXt-devel libXtst-devel libcurl-devel libuuid-devel libyubikey-devel make openssl-devel wxGTK3-devel xerces-c-devel ykpers-devel qrencode-devel file-devel file-devel redhat-lsb-core devtoolset-7-gcc-c++
-    mock "$@" --enable-network --chroot 'su - mockbuild -c "git clone https://github.com/pwsafe/pwsafe.git; mkdir -p pwsafe/build; cd pwsafe/build; scl enable devtoolset-7 -- cmake -D wxWidgets_CONFIG_EXECUTABLE=/usr/libexec/wxGTK3/wx-config ..; scl enable devtoolset-7 -- cpack -G RPM"'
+mock "$@" install cmake dpkg gcc-c++ git libXt-devel libXtst-devel libcurl-devel libuuid-devel libyubikey-devel make openssl-devel wxGTK-devel xerces-c-devel ykpers-devel qrencode-devel file-devel ImageMagick libayatana-appindicator-gtk3-devel
```
dskrvk: Could we list these packages one per line, so it will be easier to see the differences in the future when the list changes?

Misc/mock-build.sh
```diff
@@ -4,18 +4,76 @@
...
+        "$WXBASE" \
+        "$WXBASE_DEVEL" \
+        "$WXGTK" \
+        "$WXGTK_DEVEL" \
+        "$WXGTK_GL" \
+        "$WXGTK_I18N" \
+        "$WXGTK_MEDIA" \
+        "$WXGTK_WEBVIEW" \
+        /tmp/
+    mock "$@" --chroot "rpm -Uvh --force /tmp/wxBase-3*.rpm /tmp/wxBase-devel*.rpm /tmp/wxGTK-3*.rpm /tmp/wxGTK-devel*.rpm /tmp/wxGTK-gl*.rpm /tmp/wxGTK-i18n*.rpm /tmp/wxGTK-media*.rpm /tmp/wxGTK-webview*.rpm"
```
dskrvk: 1. Are `wxBase-3`/`wxGTK-3` meant to be `wxBase-${VER_GLOB}`/`wxGTK-${VER_GLOB}`?
2. It would be nice to have just a list of packages, namely `wxBase wxBase-devel wxGTK ...` and have the script expand that to:
   1. a list of RPM names for the `mock --copyin` command,
   2. a list of directories for the `mock --chroot` command.
Easier to maintain, less error-prone.

Misc/mock-build.sh
ykne: Done in 145c55b43 — the install list is now one package per line.

Misc/mock-build.sh
ykne: Good catch on both — fixed in 145c55b43.

1. Yes, that was a bug: the `rpm -Uvh` step hardcoded `wxBase-3*`/`wxGTK-3*` instead of using `$VER_GLOB`, so it could silently diverge from what `--copyin` actually staged. Both lists are now built from the same resolved RPM paths, so they can't drift apart.
2. Refactored to a package-list-driven loop as suggested. It also now determines the package list by querying what's actually installed in the chroot (`rpm -qa 'wx*'`) rather than hardcoding all 8 known wxGTK subpackages, and treats an ambiguous `-v` match (more than one RPM per package) as an error listing the candidates instead of silently picking one.

Misc/mock-build.sh
```diff
@@ -41,39 +51,77 @@ if git rev-parse --is-inside-work-tree >/dev/null 2>&1; then
...
-        "$WXGTK_DEVEL" \
-        "$WXGTK_GL" \
-        "$WXGTK_I18N" \
-        "$WXGTK_MEDIA" \
-        "$WXGTK_WEBVIEW" \
-        /tmp/
-    mock "$@" --chroot "rpm -Uvh --force /tmp/wxBase-3*.rpm /tmp/wxBase-devel*.rpm /tmp/wxGTK-3*.rpm /tmp/wxGTK-devel*.rpm /tmp/wxGTK-gl*.rpm /tmp/wxGTK-i18n*.rpm /tmp/wxGTK-media*.rpm /tmp/wxGTK-webview*.rpm"
+if [ -n "$WX_RPMS" ]; then
+    mock "$@" --copyin $WX_RPMS /tmp/
+    mock "$@" --chroot "rpm -Uvh --force$WX_TMP_RPMS"
```
dskrvk, line 124: Oops, missing a space here

ykne: Fixed in 737556ca9 — added the space explicitly and stripped the accumulator's own leading space so there's exactly one, deterministically, instead of relying on the hidden leading space in $OVERRIDE_TMP_RPMS.

PR #1902, opened 2026-08-07
mock-build.sh: don't touch %_rpmdir unless -v/-a is given

ykne:
## Problem

`mock-build.sh` with no arguments is documented (and expected) to build against stock Fedora repo packages only. In practice, the default `VER_GLOB="[0-9]*"` matched *any* locally-built RPM under `%_rpmdir`, so if a developer had ever locally built a wx* package there (e.g. while working on the AppIndicator/SNI tray-icon patch from #1895), a plain no-argument run would silently force-install that local build over the repo package instead of leaving the repo package alone. Confirmed live: a no-arg run picked up a locally-built wxGTK 3.3.3 instead of Fedora 42's repo wxGTK-devel 3.2.8, and the pwsafe source failed to compile against wx 3.3's newer `_()` literal-only translation macro.

## Fix

Three small follow-ups to #1896 (already reviewed/merged), continuing on the same branch:

1. Add `-a APPINDICATORVERSION`, generalizing the existing `-v` override mechanism to also cover `libayatana-appindicator*` packages (needed since wxGTK's package now `Requires: libayatana-appindicator-gtk3 >= 0.6.0`, so pinning both families sometimes needs a single combined `rpm -Uvh` transaction).
2. Make the local-build lookup opt-in only: `collect_local_build()` now returns immediately unless a version was explicitly requested via `-v`/`-a`. With no arguments, nothing under `%_rpmdir` is queried or installed — mock's stock repo packages are used as-is. An unresolved `-v`/`-a` version is now a hard error instead of a silent "leave as installed" skip, since silently ignoring an explicit pin request would be more confusing than failing loudly.
3. Fix a missing separating space in the combined `rpm -Uvh --force` chroot command (previously worked by accident because the accumulator string happened to carry a leading space).

## Test plan

- [x] Ran `Misc/mock-build.sh` with no arguments on Fedora 42: confirmed no `--copyin`/`rpm -Uvh` override step ran, `cmake` picked up stock repo `wxGTK-devel-3.2.8`, and the build completed through `cpack -G RPM` producing `passwordsafe-fedora42-1.25-amd64.rpm`.

### DIFF
--- a/Misc/mock-build.sh
+++ b/Misc/mock-build.sh
@@ -1,4 +1,14 @@
 #! /bin/sh
+# Builds a pwsafe RPM, via mock, for an arbitrary Fedora or RHEL-clone
+# target - from master, or from whatever branch is checked out locally -
+# without needing a full pwsafe dev environment installed on this system,
+# only mock itself. Useful as a clean-room sanity build of a local dev
+# branch before pushing it.
+#
+# Can also optionally overlay newer wx*/libayatana-appindicator* packages
+# from a local rpmbuild tree into the mock chroot only (see -v/-a below);
+# this system's own installed packages are never touched either way.
+#
 # Following from https://github.com/pwsafe/pwsafe/issues/543 by ykne.
 # Mock can be installed via dnf or yum.
 # With no argument compiles for current system.
@@ -71,14 +81,12 @@ mock "$@" install \
     libcurl-devel \
     libuuid-devel \
     libyubikey-devel \
-    make \
     openssl-devel \
     wxGTK-devel \
     xerces-c-devel \
     ykpers-devel \
     qrencode-devel \
     file-devel \
-    ImageMagick \
     libayatana-appindicator-gtk3-devel
 
 # version -> glob against local RPM filenames, e.g. "3.2.12" -> "3.2.12-*",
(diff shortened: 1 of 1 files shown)


Misc/mock-build.sh
```diff
@@ -71,57 +81,78 @@ mock "$@" install \
...
+        fi
+        OVERRIDE_RPMS="$OVERRIDE_RPMS $rpm_path"
+        OVERRIDE_TMP_RPMS="$OVERRIDE_TMP_RPMS /tmp/$(basename "$rpm_path")"
+    done
+}
+
+OVERRIDE_RPMS=""
+OVERRIDE_TMP_RPMS=""
+collect_local_build "$WANT_WX_VERSION" 'wx*' '^wx' "$@"
+collect_local_build "$WANT_APPINDICATOR_VERSION" 'libayatana-appindicator*' '^libayatana-appindicator' "$@"
```
dskrvk: Nit: could we simplify this and use a single argument (e.g. `'wx'`) instead of glob and regex_anchor, and have `collect_local_build` append the characters necessary for both cases?

dskrvk (approved):

dskrvk:
Also, if we end up having additional version arguments, it may be time to generalize to some kind of version spec input, e.g. `<package-name1>:<version1>,<package-name2>:<version2>...` or some kind of spec file. This would remove the need for the script to hard-code the globs etc. Probably overkill for now though.

Misc/mock-build.sh
```diff
@@ -71,57 +81,78 @@ mock "$@" install \
...
-        echo "no ${VER_GLOB} rpm found for $pkg, leaving as installed" >&2
-        continue
-    fi
-    WX_RPMS="$WX_RPMS $rpm_path"
-    WX_TMP_RPMS="$WX_TMP_RPMS /tmp/$(basename "$rpm_path")"
-done
+# Collects RPM paths for packages matching $2 (an `rpm -qa` glob, e.g. 'wx*')
+# that are actually installed in the chroot, at local %_rpmdir build $1, into
+# the shared $OVERRIDE_RPMS/$OVERRIDE_TMP_RPMS accumulators below. Does
+# nothing at all if $1 is empty - no query, no accumulation.
```
ronys: Too much detail. Please remove or at least shorten this comment.

Misc/mock-build.sh
ronys, line 12: A more friendly explanation of the purpose of this script would be better, including when it should (or could) be used instead of cmake/cpack.

dskrvk (approved):

PR #1903, opened 2026-08-08
[win] Fixed regression introduced post 3.72.01

ronys:
Due to the way the Group listbox was populated in OnInitDialog(), the tweaks to the event loop processing caused this to be skipped, resulting in a blank group control even for an entry that's in a group.

This fixes the problem and adds some checks in other controls for consistency.

### DIFF
--- a/src/ui/Windows/AddEdit_PasswordPolicy.cpp
+++ b/src/ui/Windows/AddEdit_PasswordPolicy.cpp
@@ -276,6 +276,9 @@ BOOL CAddEdit_PasswordPolicy::OnInitDialog()
 
 void CAddEdit_PasswordPolicy::OnChanged()
 {
+  if (IsChangeIgnored())
+    return;
+
   UpdateData(TRUE);
   NotifyChanged();
 }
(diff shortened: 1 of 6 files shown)


src/ui/Windows/AddEdit_PasswordPolicy.cpp
```diff
@@ -276,6 +276,9 @@ BOOL CAddEdit_PasswordPolicy::OnInitDialog()
 
 void CAddEdit_PasswordPolicy::OnChanged()
 {
+  if (IsChangeIgnored())
+    return;
+
   UpdateData(TRUE);
```
dskrvk: The new check always seems to precede this call - could we move the check into `UpdateData`? Currently nothing is forcing the developer to make the check, and it's easy to miss for new code.

src/ui/Windows/AddEdit_PropertyPage.cpp
```diff
@@ -36,9 +36,14 @@ CAddEdit_PropertyPage::CAddEdit_PropertyPage(CWnd *pParent,
...
 
+bool CAddEdit_PropertyPage::IsChangeIgnored() const
+{
+  return !m_bInitdone || m_AEMD.uicaller == IDS_VIEWENTRY || m_AEMD.ucprotected != 0;
+}
+
 void CAddEdit_PropertyPage::NotifyChanged()
 {
-  if (!m_bInitdone || M_uicaller() == IDS_VIEWENTRY || M_protected() != 0)
+  if (IsChangeIgnored())
```
dskrvk, line 46: We now seem to always do the same check before calling `NotifyChanged` - is this an additional guardrail? What's the downside of only making the check here?

dskrvk (approved):

PR #1904, opened 2026-08-08
[wx] More unused variables

nobugshere:
Removes a couple of unused variables that are only flagged some of the time.  In this case, when I build with wx 3.3.3.1 and PR #1852.


dskrvk (approved):

src/ui/wxWidgets/AddEditPropSheetDlg.cpp
ronys, line 1: Actually, it would be better to use the result of the switch statement to set the dialog box's text.
Currently the text is that of SYMBOL_AUTOPROPSHEETDLG_TITLE: "Add, Edit or View Entry", which , while accurate, is less user friendly.

ronys (changes requested):

src/ui/wxWidgets/AddEditPropSheetDlg.cpp
nobugshere: Well, now that you mention it ... :-)  
Done.

dskrvk (approved):

PR #1907, opened 2026-08-11
[wx] Set wxDEBUG_LEVEL=0 in cmake release builds

nobugshere:
While working on something else, I discovered that `__WXDEBUG__` is always defined in Linux and FreeBSD builds.  The wxWidgets library build defaults to `wxDEBUG_LEVEL=1`, which cause `__WXDEBUG__` to be defined unless we compile with `wxDEBUG_LEVEL=0`.  The repo packages must have used the default build.

1. Add `wxDEBUG_LEVEL=0` to the CMakeLists.txt file for release builds.  The Xcode project already has this.
2. I only found `__WXDEBUG__` in a few places in PWSafeApp.cpp, so I replaced them with `DEBUG/_DEBUG`

Tested on:
macOS M1Max 15.7.9 Sequoia,                 Xcode 16.4, wx 3.2.10 and 3.2.11

### DIFF
--- a/src/ui/wxWidgets/PWSafeApp.cpp
+++ b/src/ui/wxWidgets/PWSafeApp.cpp
@@ -1059,10 +1059,8 @@ void PWSafeApp::OnHelp(wxCommandEvent& evt)
       wxHtmlModalHelp help(window, helpFileNamePath, itr->second, wxHF_DEFAULT_STYLE);
     }
     else {
-#if defined(_DEBUG) || defined(DEBUG)
       msg << _("Please inform the developers.");
       wxMessageBox(msg, _("Help Undefined"), wxOK | wxICON_EXCLAMATION);
-#endif
     } // keyName not found in map
   }
   else {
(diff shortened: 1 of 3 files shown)


dskrvk (approved):

src/ui/wxWidgets/PWSafeApp.cpp
```diff
@@ -1059,7 +1059,7 @@ void PWSafeApp::OnHelp(wxCommandEvent& evt)
       wxHtmlModalHelp help(window, helpFileNamePath, itr->second, wxHF_DEFAULT_STYLE);
     }
     else {
-#ifdef __WXDEBUG__
+#if defined(_DEBUG) || defined(DEBUG)
```
ronys: Actually, I think that the conditional here doesn't make much sense, and should be removed.

nobugshere: Agreed.

PR #1911, opened 2026-08-15
Sf1614 pref cleanup

ronys:
Resolves https://sourceforge.net/p/passwordsafe/bugs/1614/

### DIFF
--- a/src/core/PWSprefs.cpp
+++ b/src/core/PWSprefs.cpp
@@ -121,7 +121,7 @@ const PWSprefs::boolPref PWSprefs::m_bool_prefs[NumBoolPrefs] = {
   {_T("ShowDragbar"), true, ptApplication},                 // application
   {_T("ClearClipboardOnMinimize"), true, ptApplication},    // application
   {_T("ClearClipboardOnExit"), true, ptApplication},        // application
-  {_T("ShowFindToolBarOnOpen"), false, ptApplication},      // application
+  {_T("ShowFindToolBarOnOpen"), false, ptDeprecatedApp},    // deprecated - superseded by FindToolBarActive
   {_T("NotesWordWrap"), false, ptApplication},              // application
   {_T("LockDBOnIdleTimeout"), true, ptDatabase},            // database
   {_T("HighlightChanges"), true, ptApplication},            // application
@@ -226,6 +226,13 @@ const PWSprefs::stringPref PWSprefs::m_string_prefs[NumStringPrefs] = {
   {_T("TreeSort"), _T("group"), ptApplication},                     // application
   {_T("ActiveFilterName"), _T(""), ptDatabase},                     // database
 };
+
+namespace {
+  // A deprecated pref is read for backward compatibility but never written.
+  inline bool IsDBPrefType(PWSprefs::PrefType pt) {
+    return pt == PWSprefs::ptDatabase || pt == PWSprefs::ptDeprecatedDB;
+  }
+} // namespace
 
 PWSprefs *PWSprefs::GetInstance()
 {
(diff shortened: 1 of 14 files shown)


src/core/PWSprefs.cpp
```diff
@@ -1554,7 +1561,7 @@ void PWSprefs::SaveApplicationPreferences()
       }
       m_boolChanged[i] = false;
     }
-    if (m_bool_prefs[i].ptype == ptObsolete) {
+    if (m_bool_prefs[i].ptype == ptObsolete || m_bool_prefs[i].ptype == ptDeprecatedApp) {
```
dskrvk, line 1564: What's the difference between obsolete and deprecated?

src/core/PWSprefs.h
```diff
@@ -192,8 +193,13 @@ class PWSprefs
...
 
-  // Preference types - values are powers of 2, except ptAll = sum of previous values
-  enum PrefType {ptObsolete = 0, ptDatabase = 1, ptApplication = 2, ptAll = 3};
+  // Preference types.
+  // Values must stay in sync with the mirrored enum + stringTypes[] in Tools/Windows/Minidumps/.../Copy_PWS_prefs.h/.cpp.
+  // ptDeprecatedDB/ptDeprecatedApp preferences are read but cleaned up on the next save
+  // Deprecated preferences should be obsoleted within a release or two of being deprecated.
+  // ptLast must always alias to the last value, for use as a range-check bound.
+  enum PrefType {ptObsolete = 0, ptDatabase = 1, ptApplication = 2, ptAll = 3,
+                 ptDeprecatedDB = 4, ptDeprecatedApp = 5, ptLast = ptDeprecatedApp};
```
dskrvk, line 202: If the value of `ptLast` is going to keep changing as we add new types, why can't we just keep using `ptAll` for the same purpose (change its value to `ptDeprecatedApp`)?

src/core/PWSprefs.cpp
ronys: Obsolete means that the preference is neither read nor written.
Deprecated means that it's read, but not written. Higher-level code is responsible for managing the migration.

src/core/PWSprefs.h
ronys: Keeping the names separate allows for cleaner code where it's tested. ptLast is used in a sanity check (assert) on the value.

src/core/PWSprefs.cpp
dskrvk: Aha, so the only reason to keep obsolete pref names in the code is so that this method can delete them if seen? That's the only interaction with obsolete prefs, whereas deprecated ones still have to be read until they graduate to obsolete. Got it.

src/core/PWSprefs.h
dskrvk: Oh, I missed this part: `ptAll = sum of previous values`. So `ptAll` is used to indicate "any valid, non-deprecated, app or db preference", and now we need another value to indicate the end of the list.

src/core/PWSprefs.cpp
dskrvk: Should we add another method like `IsActivePrefType` that excludes all of (obsolete, deprecatedDB, deprecatedApp)?

dskrvk (approved):

src/core/PWSprefs.h
```diff
@@ -192,8 +193,13 @@ class PWSprefs
...
     DisplayModeDark = 2, maxDisplayMode = 2};
 
-  // Preference types - values are powers of 2, except ptAll = sum of previous values
-  enum PrefType {ptObsolete = 0, ptDatabase = 1, ptApplication = 2, ptAll = 3};
+  // Preference types.
+  // Values must stay in sync with the mirrored enum + stringTypes[] in Tools/Windows/Minidumps/.../Copy_PWS_prefs.h/.cpp.
+  // ptDeprecatedDB/ptDeprecatedApp preferences are read but cleaned up on the next save
+  // Deprecated preferences should be obsoleted within a release or two of being deprecated.
+  // ptLast must always alias to the last value, for use as a range-check bound.
+  enum PrefType {ptObsolete = 0, ptDatabase = 1, ptApplication = 2, ptAll = 3,
```
dskrvk, line 201: Maybe for clarity better to say

```suggestion
  enum PrefType {ptObsolete = 0, ptDatabase = 1, ptApplication = 2, ptAll = ptDatabase + ptApplication,
```
?

src/core/PWSprefs.cpp
ronys: If this predicate appeared more than once or twice, then perhaps. as-is, I think that would be overkill.

PR #1912, opened 2026-08-16
Misc: mock-build.sh: fail fast on mock errors, validate branch name, add -h

ykne:
## Summary
- Add explicit failure checks after every mock invocation that lacked one (`--init`, `install`, `--copyin`, override `--chroot`, build `--chroot`, `--copyout`), instead of a blanket `set -e` which would break the intentionally-tolerant no-upstream-tracking fallback path (the script needs to tolerate `git rev-parse --abbrev-ref --symbolic-full-name @{u}` failing when the current branch has no upstream).
- Validate the current git branch name against the same safe-charset pattern already used for `-v`/`-a` before interpolating it unquoted into the chroot clone command string; git branch names may legally contain shell metacharacters.
- Add `-h`/`--help` handling.

## Test plan
- Ran the script (with these changes layered on top of other in-flight branches) against real local `%_rpmdir` RPMs with `-v 3.3.4 -a 0.6.0`, confirming: the new fail-fast checks correctly abort with a clear `ERROR:` message on both an injected mock failure and a genuine build failure instead of silently falling through; normal successful builds still complete and produce the RPM via `--copyout`.

### DIFF
--- a/Misc/mock-build.sh
+++ b/Misc/mock-build.sh
@@ -25,14 +25,15 @@ usage() {
     echo "Usage: $0 [-v WXVERSION] [-a APPINDICATORVERSION] [mock options]" >&2
     echo "  -v WXVERSION            Override wx* packages with this local %_rpmdir build" >&2
     echo "  -a APPINDICATORVERSION  Override libayatana-appindicator* packages with this local %_rpmdir build" >&2
-    exit 1
+    exit "${1:-1}"
 }
 
 WANT_WX_VERSION=""
 WANT_APPINDICATOR_VERSION=""
 
-while getopts ":v:a:" opt; do
+while getopts ":hv:a:" opt; do
     case $opt in
+        h) usage 0 ;;
         v) WANT_WX_VERSION="$OPTARG" ;;
         a) WANT_APPINDICATOR_VERSION="$OPTARG" ;;
         :) echo "ERROR: -$OPTARG requires an argument" >&2; usage ;;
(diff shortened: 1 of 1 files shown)


dskrvk:
I think `set -e` is a better choice, since there are (presumably) only a few places that may fail and have fallbacks paths (easy enough to handle those selectively). Most failed commands should cause the script to fail, and it would be tedious to manually add `exit 1` everywhere (and the script may grow in the future).

ykne:
Good point, switched to `set -e` in 38fa12323 - it does make the script more robust as it grows, agreed.

Two spots needed an explicit `|| true` to keep working correctly, since they involve commands that are expected to legitimately "fail" without that meaning the script should abort:

- `UPSTREAM=$(git rev-parse --abbrev-ref --symbolic-full-name @{u} 2>/dev/null)` - exits non-zero whenever the current branch has no upstream configured, which is a normal case here (falls through to the default clone URL), not an error.
- `count=$(printf '%s\n' "$candidates" | grep -c .)` - `grep -c` exits 1 on zero matches even though it printed a valid `0`. Zero local RPM candidates for a package is a legitimate, already-handled outcome (the caller prints its own "no rpm found" message), not a failure `set -e` should act on.

I verified both actually abort the script under `set -e` with isolated repros before adding the guards (not just reasoning from shell semantics), and re-ran the mock build end-to-end afterward (including the ambiguous-version-match path) to confirm nothing regressed.

dskrvk (approved):

PR #1913, opened 2026-08-17
Add cross-platform FFlush and separate it from FClose

huven:
## Summary

This supersedes the original macOS-only `FClose` change with consistent
cross-platform file-closing semantics.

- Add a platform-specific `FFlush()` that flushes stdio buffers and
  synchronizes data to storage:
  - macOS: `fflush()` followed by `F_FULLFSYNC`
  - Unix: `fflush()` followed by `fsync()`
  - Windows: `fflush()` followed by `FlushFileBuffers()`
- Move the platform-independent `FClose()` implementation into
  `os/file.h`.
- Add `FFlushAndClose()`, which always closes the stream and returns
  `EOF` if either operation fails.
- Use `FFlushAndClose()` when closing writable safe files.
- Update other callers to use the new `FClose()` signature.
- Add tests for null streams and for closing the descriptor after a
  flush failure.

This keeps ordinary closing separate from durable flushing while ensuring
that database writes explicitly request both operations.

## Testing

- `ninja test` on Linux
- 2/2 tests passed, including the new `fsync` failure-path test

### DIFF
--- a/src/core/ItemAtt.cpp
+++ b/src/core/ItemAtt.cpp
@@ -185,7 +185,7 @@ int CItemAtt::Import(const stringT &fname)
 
   auto flen = static_cast<size_t>(pws_os::fileLength(fhandle));
   if (flen > CItemAtt::MAX_SIZE) {
-    pws_os::FClose(fhandle, false);
+    pws_os::FClose(fhandle);
     return PWScore::MAX_SIZE_EXCEEDED;
   }
 
@@ -195,12 +195,12 @@ int CItemAtt::Import(const stringT &fname)
 
   size_t nread = fread(data, flen, 1, fhandle);
   if (nread != 1) {
-    pws_os::FClose(fhandle, false);
+    pws_os::FClose(fhandle);
     status = PWScore::READ_FAIL;
     goto done;
   }
 
-  if (pws_os::FClose(fhandle, true) != 0) {
+  if (pws_os::FClose(fhandle) != 0) {
     status = PWScore::READ_FAIL;
     goto done;
   }
--- a/src/core/PWSfile.cpp
+++ b/src/core/PWSfile.cpp
@@ -71,7 +71,7 @@ PWSfile *PWSfile::MakePWSfile(const StringX &a_filename, const StringX &passkey,
           status = CANT_OPEN_FILE;
           break;
         }
-        pws_os::FClose(fd, false);
+        pws_os::FClose(fd);
         [[fallthrough]];
       default:
         status = FAILURE;
@@ -158,7 +158,10 @@ int PWSfile::Close()
   int rc(SUCCESS);
 
   if (m_fd != nullptr) {
-    rc = pws_os::FClose(m_fd, m_rw == Write);
+    if (m_rw == Write)
+      rc = pws_os::FFlushAndClose(m_fd);
+    else
+      rc = pws_os::FClose(m_fd);
     m_fd = nullptr;
   }
 
@@ -415,7 +418,7 @@ bool PWSfile::Encrypt(const stringT &fn, const StringX &passwd, stringT &errmess
     goto exit;
   } // catch
 
-  status = (pws_os::FClose(out, true) == 0); out = nullptr;
+  status = (pws_os::FClose(out) == 0); out = nullptr;
 
  exit:
   if (!status)
@@ -423,8 +426,8 @@ bool PWSfile::Encrypt(const stringT &fn, const StringX &passwd, stringT &errmess
   delete fish;
   trashMemory(buf, BUFSIZ);
   delete[] ivthing;
-  pws_os::FClose(in, false);
-  pws_os::FClose(out, true);
+  pws_os::FClose(in);
+  pws_os::FClose(out);
   return status;
 }
 
@@ -451,7 +454,7 @@ bool PWSfile::Decrypt(const stringT &fn, const StringX &passwd, stringT &errmess
   file_len = pws_os::fileLength(in);
 
   if (file_len < (8 + sizeof(randhash) + 8 + SaltLength)) {
-    pws_os::FClose(in, false);
+    pws_os::FClose(in);
     LoadAString(errmess, IDSC_FILE_TOO_SHORT);
     return false;
   }
@@ -474,7 +477,7 @@ bool PWSfile::Decrypt(const stringT &fn, const StringX &passwd, stringT &errmess
 
   GenRandhash(passwd, randstuff, temphash);
   if (memcmp(reinterpret_cast<char *>(randhash), reinterpret_cast<char *>(temphash), SHA1::HASHLEN) != 0) {
-    pws_os::FClose(in, false);
+    pws_os::FClose(in);
     LoadAString(errmess, IDSC_BADPASSWORD);
     return false;
   }
@@ -558,8 +561,8 @@ bool PWSfile::Decrypt(const stringT &fn, const StringX &passwd, stringT &errmess
   delete[] ivthing;
   if (!status)
     errmess = ErrorMessages();
-  pws_os::FClose(in, false);
-  pws_os::FClose(out, true);
+  pws_os::FClose(in);
+  pws_os::FClose(out);
 
   return status;
 }
--- a/src/os/file.h
+++ b/src/os/file.h
@@ -32,7 +32,23 @@ namespace pws_os {
   extern void TryUnlockFile(const stringT &filename, HANDLE &lockFileHandle);
 
   extern std::FILE *FOpen(const stringT &filename, const TCHAR *mode);
-  extern int FClose(std::FILE *fd, const bool &bIsWrite);
+  // Flush stdio and OS buffers to storage. A null stream is a no-op.
+  // Returns zero on success and EOF on error.
+  extern int FFlush(std::FILE *fd);
+
+  inline int FClose(std::FILE *fd)
+  {
+    return fd != nullptr ? std::fclose(fd) : 0;
+  }
+
+  inline int FFlushAndClose(std::FILE *fd)
+  {
+    // Closing is unconditional, including when flushing fails.
+    const int flushResult = FFlush(fd);
+    const int closeResult = FClose(fd);
+    return flushResult == 0 && closeResult == 0 ? 0 : EOF;
+  }
+
   extern size_t fileLength(std::FILE *fp);
   extern bool GetFileTimes(const stringT &filename,
       time_t &ctime, time_t &mtime, time_t &atime);
--- a/src/test/ImportXmlTest.cpp
+++ b/src/test/ImportXmlTest.cpp
@@ -32,12 +32,12 @@ std::string readFileUtf8(const stringT &filename)
   const size_t length = pws_os::fileLength(fp);
   EXPECT_NE(length, static_cast<size_t>(-1));
   if (length == static_cast<size_t>(-1)) {
-    pws_os::FClose(fp, false);
+    pws_os::FClose(fp);
     return {};
   }
   std::vector<char> buffer(length);
   EXPECT_EQ(length, fread(buffer.data(), 1, length, fp));
-  pws_os::FClose(fp, false);
+  pws_os::FClose(fp);
   return std::string(buffer.begin(), buffer.end());
 }
 
@@ -46,7 +46,7 @@ void writeFileUtf8(const stringT &filename, const std::string &content)
   std::FILE *fp = pws_os::FOpen(filename, _T("wb"));
   ASSERT_NE(fp, nullptr);
   ASSERT_EQ(content.size(), fwrite(content.data(), 1, content.size(), fp));
-  ASSERT_EQ(0, pws_os::FClose(fp, true));
+  ASSERT_EQ(0, pws_os::FClose(fp));
 }
 
 void removeSensitiveElement(const stringT &filename)
--- a/src/test/OSTest.cpp
+++ b/src/test/OSTest.cpp
@@ -9,6 +9,13 @@
 
 #include "os/media.h"
 #include "os/dir.h"
+#include "os/file.h"
+
+#ifndef WIN32
+#include <fcntl.h>
+#include <unistd.h>
+#endif
+
 #include "gtest/gtest.h"
 
 TEST(OSTest, testMedia)
@@ -45,3 +52,29 @@ TEST(OSTest, testPath)
   out_path = pws_os::makepath(in_drive, in_dir, in_file, in_ext);
   EXPECT_EQ(in_path, out_path);
 }
+
+TEST(OSTest, testNullFileOperations)
+{
+  EXPECT_EQ(0, pws_os::FFlush(nullptr));
+  EXPECT_EQ(0, pws_os::FClose(nullptr));
+  EXPECT_EQ(0, pws_os::FFlushAndClose(nullptr));
+}
+
+#ifndef WIN32
+TEST(OSTest, testFlushAndCloseAfterFlushFailure)
+{
+  int pipeDescriptors[2];
+  ASSERT_EQ(0, pipe(pipeDescriptors));
+
+  std::FILE *fd = fdopen(pipeDescriptors[1], "w");
+  ASSERT_NE(nullptr, fd);
+  const int fileDescriptor = fileno(fd);
+  ASSERT_NE(-1, fileDescriptor);
+
+  ASSERT_EQ(1U, fwrite("x", 1, 1, fd));
+  EXPECT_EQ(EOF, pws_os::FFlushAndClose(fd));
+  EXPECT_EQ(-1, fcntl(fileDescriptor, F_GETFD));
+
+  close(pipeDescriptors[0]);
+}
+#endif
--- a/src/test/OSTest.cpp
+++ b/src/test/OSTest.cpp
@@ -72,6 +72,8 @@ TEST(OSTest, testFlushAndCloseAfterFlushFailure)
   ASSERT_NE(-1, fileDescriptor);
 
   ASSERT_EQ(1U, fwrite("x", 1, 1, fd));
+  // The stdio flush succeeds, but the OS-level sync fails because a pipe
+  // has no backing storage to synchronize.
   EXPECT_EQ(EOF, pws_os::FFlushAndClose(fd));
   EXPECT_EQ(-1, fcntl(fileDescriptor, F_GETFD));
 
--- a/src/os/mac/file.cpp
+++ b/src/os/mac/file.cpp
@@ -329,14 +329,9 @@ std::FILE *pws_os::FOpen(const stringT &filename, const TCHAR *mode)
   return retval;
 }
 
-int pws_os::FClose(std::FILE *fd, const bool &bIsWrite)
+int pws_os::FClose(std::FILE *fd, const bool &)
 {
   if (fd != NULL) {
-    if (bIsWrite) {
-      // Flush the data buffers
-      fflush(fd);
-    }
-    // Now close file
     return fclose(fd);
   }
   return 0;
--- a/src/os/mac/file.cpp
+++ b/src/os/mac/file.cpp
@@ -329,12 +329,19 @@ std::FILE *pws_os::FOpen(const stringT &filename, const TCHAR *mode)
   return retval;
 }
 
-int pws_os::FClose(std::FILE *fd, const bool &)
+int pws_os::FFlush(std::FILE *fd)
 {
-  if (fd != NULL) {
-    return fclose(fd);
-  }
-  return 0;
+  if (fd == nullptr)
+    return 0;
+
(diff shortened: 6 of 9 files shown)


ronys:
Hm.

- I think the semantics of `pws_os::FClose()` should be consistent across all os platforms.
- `fflush()` actually doesn't guarantee writing to the disk, it only flushes the userspace buffer. 
- I suggest using `fsync(fileno(fd))` before `fclose()` to ensure actually writing to disk before closing, returning an error if that fails and making sure that the error is handled properly in all calls to `pws_os::FClose()`

huven:
Adding fsync would be an unexpected side-effect for the caller, unless the call is renamed to FSyncAndClose or something alike?

huven:
Putting flush/sync/close in one method would look like this:

```cpp
int pws_os::FClose(std::FILE *fd, const bool &bIsWrite)
{
  if (fd == nullptr)
    return 0;

  int result = 0;
  int saved_errno = 0;

  if (bIsWrite) {
    // Move stdio-buffered data into the kernel before synchronizing it.
    if (fflush(fd) != 0) {
      result = EOF;
      saved_errno = errno;
    } else if (fcntl(fileno(fd), F_FULLFSYNC) == -1) {
      result = EOF;
      saved_errno = errno;
    }
  }

  // Always close the stream, even if flushing or synchronization failed.
  if (fclose(fd) != 0 && result == 0) {
    result = EOF;
    saved_errno = errno;
  }

  // fclose may have overwritten errno; expose the first failure.
  if (result != 0)
    errno = saved_errno;

  return result;
}
```

Would this be the preferred macOS behavior? It flushes the stdio buffer, requests persistence using `F_FULLFSYNC` (as fsync is not enough on macOS...), always closes the stream, and preserves the first error for the caller.

Linux could use the same structure with `fsync(fileno(fd))`, while Windows could propagate failures from `fflush()` and `FlushFileBuffers()` similarly. Separate PRs please.

This would make `FClose(fd, true)` explicitly mean a durable close across platforms. I'd still rename the method to reflect that contract though.

ronys:
I think that calling `fclose()` if `fflush()` or `fcntl(fileno(fd), F_FULLFSYNC)` fail is a mistake, as this would preclude retrying the call to `pws_os::FClose()`, which make sense, since these failures are potentially retryable.
So perhaps the signature of `pws_os::FClose()` should change to indicate whether or not a failure is retryable, and, and allow the caller to retry accordingly. Alternatively, we could implement, say, 3 retries internally before returning a failure, perhaps with a half-second wait between attempts.
I'm leaning towards the latter, but am willing to be convinced either way..

ronys:
As to the function name change, perhaps
```
int pws_os::FClose(std::FILE *fd, bool ensureFlush) // set ensureFlush to true when fd is open for write
```
would be cleaner.
In any case, the `bool` should be passed as a `bool`. `const &` is silly for a bool.

dskrvk:
> Alternatively, we could implement, say, 3 retries internally before returning a failure, perhaps with a half-second wait between attempts.

I think as long as the overall retry time is not more than a couple of seconds this is a good balance between usability and control. If the quick inline retries didn’t succeed in fixing the transient problem the user can retry manually, perhaps after restoring the connection to the io device or fixing whatever the issue was.

huven:
I'm a bit struggling here, doesn't feel right to add retry logic to close and not to all other I/O operations. What's so special about close?

huven:
From "OS" library design perspective, I would keep it simple and define FClose as "close and return success/error", and maybe add FSync (sync and return success/error) considering the wish to force persistence. Then all OS implementations can simply provide that. If magic wrapper is wanted, it can go to non-OS level, preventing every OS to re-implement the same logic.

dskrvk:
I don't think fclose should be special in how retries are handled. As for which layer to do this in (assuming @ronys thinks we need retries at all), you're right that we don't want to implement the same generic mechanism for every OS. Looks like the main consumers of pws/os are CItemAtt and PWSFile both of which do multiple OS calls per operation, so to keep the code clean(ish) a separate wrapper would be needed.

huven:
Exactly. Let me try pushing my luck a bit further...

If we define this in "os"

```cpp
namespace pws_os {
  int FSync(std::FILE *fd);
}
```

then each platform implements that differently:

- macOS: `fflush()` + `F_FULLFSYNC`
- Linux/BSD: `fflush()` + `fsync()`
- Windows: `fflush()` + `_commit()`

and we can _remove_ FClose from "os", as all code can just call std::fclose() (checking non-null of course).

The split would also clean up existing incorrect call-site classifications: several read-only streams currently pass true to FClose(). With separate operations, read paths simply close, while selected database/output paths explicitly synchronize.

ronys:
Good discussion. My inputs:

- Background: Explicitly flushing buffers to disk before closing the file is best practice - but I suspect we all agree on this.
- Retries if/when a flush fail make sense to me, because this is the only place where we're sure we're interacting with the hardware, but perhaps that's on bridge too far, as in over-engineering
- I'd rather not mix pws_os::...(FILE *) and std::...(FILE *) functions in core and above. This puts mental load on maintainers as to when to call what.

So here's what I suggest:
- The "flush+sync+close" sequence can be captured in an inline function in os/file.h, something like:
```
// pseudo code - I know it won't compile...
FFlushAndClose(fd) { // Or FCloseWriteFile(fd)?
  if (!pws_os::FFlush(fd)) { // system dependent, as @huven described (can be called FSync, although 'flush' is the common function and the more accurate description of what we're doing)
    // retry logic or just fail and return failure reason
  } else {
   return pws_os::FClose(fd) // system dependent, or just an inline to std::fclose(fd)
  }
```

nobugshere:
Would this preclude the possibility of a UI response, such as an error dialog?  Should it at least return the result of the sync/retries?

huven:
> * Background: Explicitly flushing buffers to disk before closing the file is best practice - but I suspect we all agree on this.

I wouldn't force this on all files. For the database, definitely yes, but e.g. for temporary files and exports this could severely hit performance. Your FFlushAndClose proposal fits this nicely, so each call-site can choose between FClose and FFlushAndClose.

> * Retries if/when a flush fail make sense to me, because this is the only place where we're sure we're interacting with the hardware, but perhaps that's on bridge too far, as in over-engineering

This can be added at any time, if deemed useful. Personally, I'd leave retries to the lower layer (like network retries on NAS storage).

> * I'd rather not mix pws_os::...(FILE *) and std::...(FILE *) functions in core and above. This puts mental load on maintainers as to when to call what.

Good point, FClose can just be alias of std::fclose (you also suggested below I see), it could also guard for nullptr as convenience.

> So here's what I suggest:
> 
> * The "flush+sync+close" sequence can be captured in an inline function in os/file.h, something like:
> 
> ```
> // pseudo code - I know it won't compile...
> FFlushAndClose(fd) { // Or FCloseWriteFile(fd)?
>   if (!pws_os::FFlush(fd)) { // system dependent, as @huven described (can be called FSync, although 'flush' is the common function and the more accurate description of what we're doing)
>     // retry logic or just fail and return failure reason
>   } else {
>    return pws_os::FClose(fd) // system dependent, or just an inline to std::fclose(fd)
>   }
> ```

I prefer FFlushAndClose name, and think the func should always close fd, not just when flush succeeds, otherwise the caller is clueless what to do next (still need to close fd or not?).

Putting it in the header is nice, prevents duplicate implementation across platforms.

dskrvk:
For retries, I agree we should keep that discussion (and if necessary, implementation) separate.

huven:
> Would this preclude the possibility of a UI response, such as an error dialog? Should it at least return the result of the sync/retries?

All three methods (FClose, FFlush and FFlushAndClose) need to return `int`, `0` on success, allowing caller to respond to errors.

FClose could be something like `fd != nullptr ? std::fclose(fd) : 0`

ronys:
I think that FFlushAndClose() should *not* close fd if FFlush() failed, since this means that it can never be retried.
Even if we don't want to implement a retry mechanism now, I suggest the following:
1. Pass the FILE * as a reference, that is `int FFlushAndClose(FILE *&fp)`
2. If everything succeeds, set fp to nullptr and return zero.
3. If FFlush() fails, return non-zero and leave fp untouched.
4. If FFclose() fails, return non-zero and set fp to nullptr.

This also enables us to provide more accurate diagnostics, either as logs or error messages to users.

huven:
Hm, leaving fd open in FFlushAndClose has high potential of leaking descriptors if caller forgets to check fd on failure.

IMO, a caller interested in handling FFlush failures in some magic way (like retrying), should simply not call FFlushAndClose, but use FFlush directly:

``` cpp
if (FFlush(fd) != 0) {
  // do magic here, fall through if desired or not
}
if (FClose(fd) != 0) {
  // other failure handling
}
```

Let's see if others chime in..

nobugshere:
> Hm, leaving fd open in FFlushAndClose has high potential of leaking descriptors if caller forgets to check fd on failure.

I guess I would expect such a function to close the file.  After a call to fclose() the stream pointer is unusable, whether it returned an error or not.  So, this should propagate the error code to inform the caller, but there's nothing more that can be done.

> IMO, a caller interested in handling FFlush failures in some magic way (like retrying), should simply not call FFlushAndClose, but use FFlush directly:

That makes sense to me.

dskrvk:
I like @ronys's proposed logic. It tries to provide convenience, but bails out after the first failure, letting the caller either bubble up the error or retry automatically. An automated retry handler could go something like this:
```c++
FILE *fp;
int err;
while ((err = FFlushAndClose(fp)) != 0) {
  if (fp == nullptr) {
    // flushed but not properly closed, nothing to do -> let caller decide or inform the user
    break;
  }
  // attempt to flush again
  // sleep
}
return err;
```
I don't see a need for some callers to invoke `FFlush` directly unless they want to keep the file open - don't think there's anything(?) they could do differently to handle flush errors.

huven:
So we are not afraid maintainers will use constructs like

``` cpp
if (FFlushAndClose(fp) != 0) {
  // Inform user
  return FAILURE;
}
```

leaking the file descriptor on flush error? @ronys mentioned not wanting to put mental load on maintainers as to when to call what..

nobugshere:
If FFlushAndClose() bails out early the caller would not know whether or not close had been called, unless it defines its' own error codes.  It may be better to have it always call close and the caller just knows it cannot do anything about it.  More elaborate error handling would likely involve calling FFlush multiple times, anyway.

huven:
I spent some time on researching if retries are even allowed, and this looks non-trivial.

POSIX reference (https://pubs.opengroup.org/onlinepubs/9799919799/functions/fflush.html) does not state anything at all about retries. It does not specify the state of the output buffer after the failure. There is no portable guarantee that the remaining data is still buffered for a subsequent fflush.

Microsoft reference (https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/fflush?view=msvc-170) is more explicit:

> If fflush returns EOF, data may have been lost due to a write failure.

ronys:
I'm convinced - simplicity wins. Here's my latest suggestion:
1. We keep the interface simple and always close the fd (or at least try), regardless of what failed internally. 
2. The failure be propagated up to the user, suggesting a retry or write to another file.

In other words, we don't try to be smarter than the OS, and actionably report to the user when we failed to save password data. (Meaning that the user can try something useful with the information)

huven:
Perfect. I'll have a go at it this weekend, will merge the changes in this PR so the changes stay alongside the discussion.

huven:
I've applied the changes as discussed. All three methods accept nullptr as no-op, and return either 0 on success or EOF on error.

Database writes use the new FFlushAndClose, other previous FClose callers do just FClose.

Tests on Linux/macOS could take a bit longer as database changes are now sync-ed, as intended.
Depending on storage this might not be noticeable.

Fire away! 😄

src/os/file.h
```diff
@@ -32,7 +32,23 @@ namespace pws_os {
...
+  {
+    return fd != nullptr ? std::fclose(fd) : 0;
+  }
+
+  inline int FFlushAndClose(std::FILE *fd)
+  {
+    // Closing is unconditional, including when flushing fails.
+    const int flushResult = FFlush(fd);
+    const int closeResult = FClose(fd);
+    return flushResult == 0 && closeResult == 0 ? 0 : EOF;
```
dskrvk, line 49: How would the caller know which part failed? Handling either failure might call for different steps.

src/core/PWSfile.cpp
```diff
@@ -558,8 +561,8 @@ bool PWSfile::Decrypt(const stringT &fn, const StringX &passwd, stringT &errmess
   delete[] ivthing;
   if (!status)
     errmess = ErrorMessages();
-  pws_os::FClose(in, false);
-  pws_os::FClose(out, true);
+  pws_os::FClose(in);
+  pws_os::FClose(out);
```
dskrvk, line 565: Doesn't `FClose(out, true)` now correspond to `FFlushAndClose(out)`?

src/core/ItemAtt.cpp
```diff
@@ -195,12 +195,12 @@ int CItemAtt::Import(const stringT &fname)
...
   size_t nread = fread(data, flen, 1, fhandle);
   if (nread != 1) {
-    pws_os::FClose(fhandle, false);
+    pws_os::FClose(fhandle);
     status = PWScore::READ_FAIL;
     goto done;
   }
 
-  if (pws_os::FClose(fhandle, true) != 0) {
+  if (pws_os::FClose(fhandle) != 0) {
```
dskrvk, line 203: I wonder what the original author intended by using true here, even though the file is open with "rb". But the current code makes sense.

src/core/PWSfile.cpp
```diff
@@ -415,16 +418,16 @@ bool PWSfile::Encrypt(const stringT &fn, const StringX &passwd, stringT &errmess
     goto exit;
   } // catch
 
-  status = (pws_os::FClose(out, true) == 0); out = nullptr;
+  status = (pws_os::FClose(out) == 0); out = nullptr;
```
dskrvk, line 421: Should be FFlushAndClose?

src/core/PWSfile.cpp
```diff
@@ -415,16 +418,16 @@ bool PWSfile::Encrypt(const stringT &fn, const StringX &passwd, stringT &errmess
...
  exit:
   if (!status)
     errmess = ErrorMessages();
   delete fish;
   trashMemory(buf, BUFSIZ);
   delete[] ivthing;
-  pws_os::FClose(in, false);
-  pws_os::FClose(out, true);
+  pws_os::FClose(in);
+  pws_os::FClose(out);
```
dskrvk, line 430: Not writing here since this goto handles error conditions?

src/test/ImportXmlTest.cpp
```diff
@@ -46,7 +46,7 @@ void writeFileUtf8(const stringT &filename, const std::string &content)
   std::FILE *fp = pws_os::FOpen(filename, _T("wb"));
   ASSERT_NE(fp, nullptr);
   ASSERT_EQ(content.size(), fwrite(content.data(), 1, content.size(), fp));
-  ASSERT_EQ(0, pws_os::FClose(fp, true));
+  ASSERT_EQ(0, pws_os::FClose(fp));
```
dskrvk, line 49: FFlushAndClose?

src/test/OSTest.cpp
```diff
@@ -44,4 +51,30 @@ TEST(OSTest, testPath)
...
+  int pipeDescriptors[2];
+  ASSERT_EQ(0, pipe(pipeDescriptors));
+
+  std::FILE *fd = fdopen(pipeDescriptors[1], "w");
+  ASSERT_NE(nullptr, fd);
+  const int fileDescriptor = fileno(fd);
+  ASSERT_NE(-1, fileDescriptor);
+
+  ASSERT_EQ(1U, fwrite("x", 1, 1, fd));
+  EXPECT_EQ(EOF, pws_os::FFlushAndClose(fd));
```
dskrvk, line 77: Why EOF?

src/os/file.h
huven: The caller doesn't know. If the caller wants to respond differently to either failure, it should just call FFlush and FClose separately. FFlushAndClose is convenience if caller doesn't care (most common case?).

src/core/PWSfile.cpp
huven: Not necessarily, FFlushAndClose is for when code really wants to enforce storage persistence on close, regular FClose is usually enough. That depends on the context of course, so that's why I made database writes use FFlushAndClose.

src/core/ItemAtt.cpp
huven: Yeah, there were more like that 😄

src/core/PWSfile.cpp
dskrvk: Sure, but in this case out is a writable stream - don't we want to attempt to flush?

src/core/PWSfile.cpp
huven: Not necessarily, we can choose to. This method is, I think, only used when using command-line file encryption (-e option), not sure if anybody uses that?

src/core/PWSfile.cpp
huven: Just not forcing persistent storage.

src/test/ImportXmlTest.cpp
huven: not needed here

src/core/PWSfile.cpp
dskrvk: Same as https://github.com/pwsafe/pwsafe/pull/1913#discussion_r3836665761. Also, we shouldn't assume this method is unused and that it's OK to make it less reliable.

src/test/OSTest.cpp
huven: sync is not supported on pipe, so it must fail. This tests the convenience wrapper testing and returning that.

dskrvk: Maybe add a comment?

src/core/PWSfile.cpp
huven: I wouldn't do it, this is just a file conversion. For databases, fsync is common. Utils like cp, sed, awk etc. don't sync.

As long as return value is not checked, there is no real point of doing it anyway.

src/core/PWSfile.cpp
huven: For a file conversion this is really uncommon. Suppose you have a disk with 1 million files you want to encrypt or decrypt. Syncing after each file would slow the process down immensely. One final sync after the job makes more sense.

I don't mind changing this btw if consensus is that this call-site should sync, just trying to explain why I think it should be weighed carefully for each close invocation.

src/test/OSTest.cpp
huven: Will do, good suggestion.

dskrvk:
On unix fclose always does fflush automatically - isn't that the case everywhere? So the only valid use case for calling FFlush seems to be when we *don't* intend to immediately close the file (for incremental writes), but this use case doesn't seem to exist right now. Seems like we're complicating the interface for no reason - since for input and read-only streams flush should be safe why do we even need FFlushAndClose?

huven:
FFlush does two things

1. std::fflush, to force stdio buffer into kernel
2. calling platform specific method to force kernel buffer to persistent storage (sync)

So, yes, if 2. is not needed, then just calling FClose is sufficient, and most common.

FFlush is needed only for a few operations that needed high guarantees, like database writing.
FFlushAndClose is a small wrapper around it, nothing more.

src/test/OSTest.cpp
huven: done

src/core/PWSfile.cpp
dskrvk: So the rationale is: don't sync if we we're writing to a new file (like here), but for PWSfile::Close we're presumably modifying an existing file, so we want reliability? Now it makes sense.

dskrvk:
Then it's a naming thing: fflush is not the part we care about (it's always done anyway), it's fsync. FSyncAndClose?

src/core/PWSfile.cpp
huven: The sync is usually required when there is risk of incorrect behavior (e.g. strict ordering of operations) or data loss. Transaction logs, database files, repositories, things like that.

Like with the encrypt/decrypt option, if the OS crash-es 1ms after completing the command (before kernel buffers are flushed), there is still the original file.

It's usually the critical processes that use fsync, as the performance impact is real.

huven:
> Then it's a naming thing: fflush is not the part we care about (it's always done anyway), it's fsync. FSyncAndClose?

That was my first choice too, but @ronys preferred Flush instead of Sync (long way up in this discussion).

dskrvk:
Well, I vote for FSyncAndClose or smth like CommitToDeviceAndClose since it's more accurate. Otherwise the PR looks good to me.

PR #1915, opened 2026-08-17
Linux: adapt SystemTray to wxWidgets 3.3's AppIndicator API changes

ykne:
wxWidgets' upcoming 3.3 series (tracked here against `vadz/use-appindicator`, which merges Vadim Zeitlin's native AppIndicator implementation of `wxTaskBarIcon`) changes how `wxTaskBarIcon`'s AppIndicator backend manages the tray icon/menu. This branch adapts to that.

**`wxTaskBarIcon`'s new AppIndicator backend handles icon/menu internally**

wx 3.3's AppIndicator implementation of `wxTaskBarIcon` (built with `wxUSE_APPINDICATOR`) now manages the StatusNotifierItem icon and menu entirely on its own: any `wxIcon` passed to the existing `SetIcon()` call is saved to a temp PNG and registered with AppIndicator directly, and `GetPopupMenu()`/`CreatePopupMenu()` are already wired up the way `SystemTray` expects. The custom `SetSNIIconThemePath()`/`SetSNIIconName()` calls belonged to an earlier, now-superseded fork API and no longer exist on this branch, so they're dropped - the plain `SetIcon()` calls already in `SystemTray::SetTrayStatus()` do the same job. A follow-up commit removes the now-dead build-tree icon staging that only existed to feed that old lookup path (`CMakeLists.txt` custom commands, `install/graphics/32x32/*`).

**Backward compatibility**

Verified via `Misc/mock-build.sh` in this repo that the branch still builds cleanly against stock Fedora 44 packages (plain `wxGTK-devel`/`libayatana-appindicator-gtk3-devel` from the repo, no overrides). Installed the resulting RPM on stock Fedora 44 and confirmed it works as expected.

---

Edit: split the `_()`/`wxGetTranslation()` translation fix out into #1916 per review below - this PR is now scoped to just the AppIndicator/`wxTaskBarIcon` API adaptation.


dskrvk:
Could you split the translation fixes into a separate PR? Also, @sauravg took a swing at that earlier in #1852 - not sure what the status is.

src/ui/wxWidgets/SystemTray.cpp
```diff
@@ -91,12 +86,6 @@ SystemTray::SystemTray(PasswordSafeFrame* frame) : m_TrayIconWithOverlay(false),
                                                    m_frame(frame),
                                                    m_status(TrayStatus::CLOSED)
 {
-#if wxUSE_APPINDICATOR
-  wxFileName exePath(wxStandardPaths::Get().GetExecutablePath());
-  wxString trayIconsDir = exePath.GetPath() + wxT("/tray-icons");
-  if (wxDirExists(trayIconsDir))
-    SetSNIIconThemePath(trayIconsDir);
-#endif
```
dskrvk, line 99: So with wxWidgets<3.3 this works as before - the appindicator dependency change only affects 3.3+?

ykne:
Split the translation fixes out into #1916 - this PR is now scoped to just the AppIndicator/wxTaskBarIcon API adaptation. #1916 references #1852 for coordination since it looks like it covers similar ground with broader scope.

dskrvk:
Earlier implementation for reference: https://github.com/pwsafe/pwsafe/pull/1895

src/ui/wxWidgets/SystemTray.cpp
ykne: Confirmed directly from stock Fedora 44's wxGTK-devel-3.2.9 setup.h - wxUSE_APPINDICATOR doesn't exist there at all, it's not in any released wxWidgets. So on wx <3.3, the #if wxUSE_APPINDICATOR block in SystemTray.cpp (this PR's change included) is compiled out entirely and pwsafe falls back to the classic wxTaskBarIcon - completely unaffected by this PR. The change only takes effect once building against a wx that ships AppIndicator support, which today is only vadz/use-appindicator (targeting 3.3).

ykne:
Thanks for the pointer - to clarify, #1895 predates this PR and was built against a custom AppIndicator API from my own wxWidgets fork. That API has since been superseded upstream: wxWidgets now has its own native AppIndicator implementation of wxTaskBarIcon (vadz/use-appindicator, targeting the 3.3 series), with a different interface. This PR adapts pwsafe to that newer, upstream API, so the implementation #1895 required updates to match new API.

dskrvk:
Right, I only linked it here for convenience, as I was checking what other changes it introduced.

nobugshere:
> Split the translation fixes out ....

I don't mean to nit-pick, but could you also update the title, please?

dskrvk (approved):

ykne:
Good catch, done - title updated to reflect the AppIndicator-only scope.

PR #1916, opened 2026-08-17
Fix wxWidgets 3.3 compatibility: use wxGetTranslation() for non-literal strings

ykne:
wxWidgets 3.3 enforces via `static_assert` that the `_()` translation macro only accepts string literals (for `xgettext` extraction), rejecting any pointer/`wxString` argument. Several call sites pass runtime values (array/map lookups, function results) instead of literals, which fails to compile against wxWidgets 3.3. This replaces those with `wxGetTranslation()`, the underlying runtime translation call, which is unaffected by the literal-only check.

Affected files: `StringX.cpp`, `DragBarCtrl.cpp`, `OptionsPropertySheetDlg.cpp`, `PWFiltersEditor.cpp`, `PasswordSafeFrame.cpp`, `wxUtilities.cpp`.

Split out of #1915 per review - that PR now only covers the AppIndicator/`wxTaskBarIcon` API adaptation.

Related: #1852 also addresses wx 3.3 translation issues, with broader scope (also touches `GridCtrl.cpp`, `TreeCtrl.cpp`, `Misc/rc2cpp.pl`) but is currently showing merge conflicts. Happy to close this in favor of #1852 if that one gets rebased, or coordinate scope - opening this mainly to keep #1915 unblocked and API-focused as requested.


src/ui/wxWidgets/PasswordSafeFrame.cpp
```diff
@@ -3086,7 +3086,7 @@ void PasswordSafeFrame::UpdateStatusBar()
       menu->SetItemLabel(m_core.IsReadOnly() ? _("Change to R/W") : _("Change to R-O"));
   }
   else { // no open file
-    m_statusBar->SetStatusText(_(PWSprefs::GetDCAdescription(-1)), StatusBar::Field::DOUBLECLICK);
+    m_statusBar->SetStatusText(wxGetTranslation(PWSprefs::GetDCAdescription(-1)), StatusBar::Field::DOUBLECLICK);
```
dskrvk: I think this translation call is redundant: https://github.com/pwsafe/pwsafe/pull/1852#discussion_r3558297963

src/ui/wxWidgets/wxUtilities.cpp
```diff
@@ -227,7 +227,7 @@ SafeCombinationCtrl* wxUtilities::CreateLabeledSafeCombinationCtrl(wxWindow* par
   auto *sizer = new wxBoxSizer(wxVERTICAL);
   parent->GetSizer()->Add(sizer, 0, wxBOTTOM|wxLEFT|wxRIGHT|wxEXPAND, 12);
 
-  auto *labelCtrl = new wxStaticText(parent, wxID_STATIC, _(label), wxDefaultPosition, wxDefaultSize, 0);
+  auto *labelCtrl = new wxStaticText(parent, wxID_STATIC, wxGetTranslation(label), wxDefaultPosition, wxDefaultSize, 0);
```
dskrvk: And this: https://github.com/pwsafe/pwsafe/pull/1852#discussion_r3558289120

src/ui/wxWidgets/PasswordSafeFrame.cpp
ronys: Nice catch!

src/ui/wxWidgets/wxUtilities.cpp
ronys: This is indeed redundant the way CreateLabeledSafeCombinationCtrl() is currently invoked, but unlike the previous comment, this can't be ensured in the future, if someone calls this function differently.
Aside from a nice comment in the header file stating this assumption, I don't see a good fix.

nobugshere:
I did a quick diff between this PR and #1852.  The working directory is based on PR #1852 with the current master merged-in.  The additions (+) are from 1852 (green if you can see it in color.)  Looks like there are still a number of places where wxTRANSLATE would be more appropriate than _().  

[diff-1852-vs-1916.txt](https://github.com/user-attachments/files/31192138/diff-1852-vs-1916.txt)

src/ui/wxWidgets/wxUtilities.cpp
dskrvk: We could instead require that the label text not be translated by the caller (e.g. declare it as `const StringX& englishLabel`, though that won't provide any guarantees of course) and handle the translation in the method itself. There are only 4 call sites to update, so not a big change.

ykne: This was an attempt to revive #1852, which had already been approved but severely deviated from master by the time it could be rebased. If that approach is no longer considered the way to go, I can try the current suggestions instead.

dskrvk: Further improvements are a bit outside of scope, and it would be good to merge this change and not let it linger. Were you able to test it?

ronys:
I did a quick test under Debian 13 and it seems to be OK. Merging this in per @dskrvk 's suggestion, as indeed "better is the enemy of good enough".

ykne:
Tested on Fedora 44, built against the latest wxWidgets AppIndicator implementation (vadz/use-appindicator) and libayatana-appindicator 0.6.0. Verified the tray icon, including double-click support, works as expected.

PR #1918, opened 2026-08-20
[wx] Locale selection rework

### LINKED ISSUE #1711: Dialog windows still in German
Latest version of pwsafe 1.23 for OSX  still has dialog windows in German even after changing language preferences to English
See attached image

Image

nobugshere:
Resolves #1711
Related to #1712

I’m posting this as a draft in the hope of getting more testing, especially in non-english locales and unusual language/locale combinations.

This is a re-work of the locale selection code.  Per current recommendations it replaces wxLocale with wxUILocale which separates language selection from locale selection.  (The use of wxTranslations to select the language is largely unchanged.)  It also tries to keep the WX UI and C library locales in sync.  This is particularly problematic on macOS, but applies to Unix/Linux as well.  Also addresses a date format problem seen while testing with wx3.3.x on macOS (it always used d/m/y format)

Some of the trouble stems from pwsafe setting the language (e.g. en, fr, etc.) then wxLocale chooses a default locale for it (en_GB for English), which may not be what the user wants, especially the date format.  This code will use the user’s system configured locale (region code e.g. US or GB), provided the language part matches (en).

wxUILocale first appeared in wx3.1.6 and some of the logic requires 3.2.2.  Pwsafe still supports some older distros that have 3.0.5 (e.g. Ubuntu 22).  As it happens, most of the work is done with static class member functions that are the same between the two classes.  That allows for an implementation that defines a new class, PWSLocale, that is derived from either wxLocale or wxUILocale, depending on the WX version used for the build, thus allowing compatibility with the “old style” while minimizing the use of ifdef’s in the code.  Builds using a WX version prior to 3.2.2 should behave the same as before.

A few highlights:
- Existing macOS code to select an appropriate locale based on the selected language, system settings, and environment variables now applies to all platforms with wx3.2.2 or newer.
- As a side effect, this resolved an issue of German language dialogs in older versions of macOS. Issue #1711
- Disabled the ability to use the LC_PATH environment variable to locate language catalog files in Release builds
- Revised some Xcode project debug settings

macOS known issue:
- Per the wxWidgets documentation: wxUILocale and setlocale do not affect macOS native controls, e.g. date picker, and the system locale cannot be changed by the application.  If necessary, a Mac user can set the system settings as desired or use the defaults command to set the locale specifically for pwsafe.  Note, the language part needs to match the pwsafe language selection.  For example, to use Pwsafe in French, but with Canadian date format, while the system is set to en_US, use the following two commands: 
```
defaults write org.pwsafe.pwsafe AppleLocale fr_CA.   # The CA region is used 
defaults write org.pwsafe.pwsafe AppleLanguages '("fr-CA”)’   # (- instead of _) The fr language is used 
```

Then, within Pwsafe, select French for the language.

Tested on:
```
macOS M1Max 15.7.9 Sequoia,                 Xcode 16.4, wx 3.2.10, 3.2.11, and 3.3.3.1
Ubuntu 24.04.4 ARM64 VMware, Gnome/Wayland, gcc 13.3.0, wx 3.2.4, GTK 3.24.41

```

### DIFF
--- a/src/os/mac/utf8conv.cpp
+++ b/src/os/mac/utf8conv.cpp
@@ -31,6 +31,13 @@ using namespace std;
 class Startup {
 public:
   Startup() {
+    /*
+      All programs start in the "C" locale.
+      On macOS, a GUI app is typically started without any LANG/LC_* environment variables.
+      Therefore, the "" argumant treats the "system locale" as "C" and this doesn't actually change anything.
+      However, there are ways a power user can change that, so, maybe, this is worth doing.
+      It will be overridden very soon in PWSafeApp::OnInit(), anyway.
+    */
     char *sl = setlocale(LC_ALL, "");
     if (sl == NULL)
       throw "Couldn't initialize locale - bailing out";
--- a/src/os/mac/utf8conv.cpp
+++ b/src/os/mac/utf8conv.cpp
@@ -32,12 +32,12 @@ class Startup {
 public:
   Startup() {
     /*
-      All programs start in the "C" locale.
-      On macOS, a GUI app is typically started without any LANG/LC_* environment variables.
-      Therefore, the "" argumant treats the "system locale" as "C" and this doesn't actually change anything.
-      However, there are ways a power user can change that, so, maybe, this is worth doing.
-      It will be overridden very soon in PWSafeApp::OnInit(), anyway.
-    */
+     * All programs start in the "C" locale.
+     * On macOS, a GUI app is typically started without any LANG/LC_* environment variables.
+     * Therefore, the "" argumant treats the "system locale" as "C" and this doesn't actually change anything.
+     * However, there are ways a power user can change that, so, maybe, this is worth doing.
+     * It will be overridden very soon in PWSafeApp::OnInit(), anyway.
+     */
     char *sl = setlocale(LC_ALL, "");
     if (sl == NULL)
       throw "Couldn't initialize locale - bailing out";
--- a/src/ui/wxWidgets/PWSLocale.h
+++ b/src/ui/wxWidgets/PWSLocale.h
@@ -41,13 +41,18 @@
   public:
     PWSLocale() {};
     ~PWSLocale() { m_pwslocale = nullptr; }
-    static PWSLocale *m_pwslocale;
 
     static bool UseDefault();
     static void ChooseLocale(wxLanguage language);
     wxString PWSGetCurrentName() { return wxLocale::GetCanonicalName(); };
+    bool Init(wxLanguage lang)
+    {
+      if (m_pwslocale == nullptr) m_pwslocale = this;
+      return wxLocale::Init(lang);
+    }
 
   private:
+    static PWSLocale *m_pwslocale;
     static wxString appendUTF8(const wxString& ev);
   };
 #endif // LOCALE_WX322
--- a/src/ui/wxWidgets/PWSLocale.h
+++ b/src/ui/wxWidgets/PWSLocale.h
@@ -32,6 +32,11 @@ public:
   class PWSLocale : public wxUILocale, public PWSMacLocale
   {
   public:
+    static bool UseDefault() {
+      auto ret = wxUILocale::UseDefault();
+      setMacLocale(PWSGetCurrentName());
+      return ret;
+    }
     static wxString PWSGetCurrentName() { return wxUILocale::GetCurrent().GetName(); };
   };
 
@@ -42,7 +47,12 @@ public:
   {
   public:
     PWSLocale() {};
-    static bool UseDefault() { return false; }; //no-op for compatibility
+    static bool UseDefault() {
+      // Because the old version did this, possibily as problem work arounds.
+      setlocale(LC_CTYPE, "");
+      setlocale(LC_TIME, "");
+      return false;
+    };
     wxString PWSGetCurrentName() { return wxLocale::GetCanonicalName(); };
   };
 #endif // LOCALE_WX322
--- a/src/ui/wxWidgets/PWSLocale.h
+++ b/src/ui/wxWidgets/PWSLocale.h
@@ -1,0 +1,50 @@
+/*
+ * Initial version created by Richard Powell
+ *
+ * Copyright (c) 2003-2026 Rony Shapiro <ronys@pwsafe.org>.
+ * All rights reserved. Use of the code is allowed under the
+ * Artistic License 2.0 terms, as specified in the LICENSE file
+ * distributed with this code, or available from
+ * http://www.opensource.org/licenses/artistic-license-2.0.php
+ */
+
+#ifndef _PWSLOCALE_H_
+#define _PWSLOCALE_H_
+
+// wx3.2.2 and later can use wxUILocale and some improved logic
+#if wxCHECK_VERSION(3, 2, 2)
+#define LOCALE_WX322
+#endif
+
+class PWSMacLocale
+{
+public:
+#ifdef __WXMAC__
+  static void setMacLocale(const char *loc);
+#else
+  static void setMacLocale(const char *) {}; // no-op if not macOS
+#endif
+};
+
+#ifdef LOCALE_WX322
+#include <wx/uilocale.h>
+
+  class PWSLocale : public wxUILocale, public PWSMacLocale
+  {
+  public:
+    static wxString PWSGetCurrentName() { return wxUILocale::GetCurrent().GetName(); };
+  };
+
+#else // LOCALE_WX322
+
+  // Pre-wx3.2.2 compatible version
+  class PWSLocale : public wxLocale, public PWSMacLocale
+  {
+  public:
+    PWSLocale() {};
+    static bool UseDefault() { return false; }; //no-op for compatibility
+    wxString PWSGetCurrentName() { return wxLocale::GetCanonicalName(); };
+  };
+#endif // LOCALE_WX322
+
+#endif // _PWSLOCALE_H_
--- a/src/ui/wxWidgets/PWSLocale.h
+++ b/src/ui/wxWidgets/PWSLocale.h
@@ -12,24 +12,16 @@
 #define _PWSLOCALE_H_
 
 // wx3.2.2 and later can use wxUILocale and some improved logic
+// All supported macOS versions work best with wxWidgets 3.2.4 and above.
+// Unix/Linux is compatible as far back as wxWidgets 3.0.5
 #if wxCHECK_VERSION(3, 2, 2)
 #define LOCALE_WX322
 #endif
 
-class PWSMacLocale
-{
-public:
-#ifdef __WXMAC__
-  static void setMacLocale(const char *loc);
-#else
-  static void setMacLocale(const char *) {}; // no-op if not macOS
-#endif
-};
-
 #ifdef LOCALE_WX322
 #include <wx/uilocale.h>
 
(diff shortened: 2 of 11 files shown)


src/ui/wxWidgets/PWSafeApp.cpp
```diff
@@ -236,7 +238,7 @@ void PWSafeApp::Init()
 void PWSafeApp::OnAssertFailure(const wxChar *file, int line, const wxChar *func,
                 const wxChar *cond, const wxChar *msg)
 {
-  if (m_locale)
+  if (m_PWSLocaleIsSet)
```
dskrvk, line 241: Not sure why there was a condition here. The [docs for OnAssertFailure](https://docs.wxwidgets.org/latest/classwx_app_console.html#a5f91d78bbfe3c9ef28e8e3c8fdf29989) don't mention a requirement for locale to be initialized.

src/os/mac/utf8conv.cpp
```diff
@@ -31,6 +31,13 @@ using namespace std;
...
 public:
   Startup() {
+    /*
+     * All programs start in the "C" locale.
+     * On macOS, a GUI app is typically started without any LANG/LC_* environment variables.
+     * Therefore, the "" argumant treats the "system locale" as "C" and this doesn't actually change anything.
+     * However, there are ways a power user can change that, so, maybe, this is worth doing.
+     * It will be overridden very soon in PWSafeApp::OnInit(), anyway.
+     */
     char *sl = setlocale(LC_ALL, "");
```
dskrvk, line 41: I see there's a fancy version that wx recommends: https://docs.wxwidgets.org/latest/classwx_app_console.html#a9a02416d2eeaf63b0cb9313dad69bb0f

> Calling this function is roughly equivalent to calling
> setlocale(LC_ALL, "");
> but performs additional toolkit-specific tasks under some platforms and so should be used instead of setlocale() itself.

src/ui/wxWidgets/PWSafeApp.cpp
```diff
@@ -712,40 +674,43 @@ bool PWSafeApp::ActivateLanguage(wxLanguage language, bool tryOnly)
...
-      }
-#endif // wxCHECK_VERSION
-      // This value must be set for mac OS starting with version 11, but is no problem for earlier versions, see
-      // https://github.com/wxWidgets/wxWidgets/issues/19023
-      // https://docs.wxwidgets.org/3.2/classwx_locale.html
-      setlocale(LC_NUMERIC, "C");
-#else // __WXMAC__
+      envString += ".UTF-8";
+      PWSLocale::UseLocaleName(envString);
+      PWSLocale::setMacLocale(envString);
```
dskrvk: Why not do this as part of `UseLocaleName`, when applicable? Ideally the public interface should not have any OS-specific parts.

src/ui/wxWidgets/PWSafeApp.cpp
```diff
@@ -754,6 +719,24 @@ bool PWSafeApp::ActivateLanguage(wxLanguage language, bool tryOnly)
 #ifdef __WXMAC__
 // macOS specific functions for UI interactions...
 
+void PWSMacLocale::setMacLocale([[maybe_unused]] const char *loc)
```
dskrvk: Defined in the wrong file?

src/ui/wxWidgets/PWSafeApp.cpp
```diff
@@ -754,6 +719,24 @@ bool PWSafeApp::ActivateLanguage(wxLanguage language, bool tryOnly)
...
+    pws_os::Trace(L"Failed to set locale to: %s", loc);
+  }
+#endif // LOCALE_WX322
+
+  // This value must be set for mac OS starting with version 11, but is no problem for earlier versions, see:
+  // https://trac.wxwidgets.org/ticket/19023
+  // https://docs.wxwidgets.org/3.2/classwx_locale.html
+  int major, minor;
+  wxGetOsVersion(&major, &minor);
+  if (major == 11 || (major == 12 && minor < 3)) {
```
dskrvk: Why limit to 12.3? Can we just do this for all v. >=11?

src/ui/wxWidgets/PWSLocale.h
```diff
@@ -0,0 +1,60 @@
...
+
+#ifndef _PWSLOCALE_H_
+#define _PWSLOCALE_H_
+
+// wx3.2.2 and later can use wxUILocale and some improved logic
+#if wxCHECK_VERSION(3, 2, 2)
+#define LOCALE_WX322
+#endif
+
+class PWSMacLocale
```
dskrvk: Opinion: maybe an overkill to define a class - just have a single definition of `PWSLocale::setMacLocale` with the body enclosed in `ifdef`?

src/ui/wxWidgets/PWSafeApp.cpp
nobugshere: I'm not sure either.  I considered removing it, but I assumed it's there to solve a past problem, so I left it.  Maybe it's obsolete now, should I remove it?

src/os/mac/utf8conv.cpp
nobugshere: Yes, but the OS code is generally not wxWidgets aware.  It will be overridden soon enough.  I wonder if it was related to the OnAssertFailure issue?  

Maybe I'll use the wx version in other places.

src/ui/wxWidgets/PWSafeApp.cpp
nobugshere: Good idea, I'll do that.

src/ui/wxWidgets/PWSafeApp.cpp
nobugshere: As I understand it, the bug was supposedly fixed in 12.3.  Beyond that we can use the chosen locale instead of forcing it to "C".

src/ui/wxWidgets/PWSLocale.h
nobugshere: You may be right.  I think I started down a different path where that class had more to do.  I'll look into simplifying it.

src/ui/wxWidgets/PWSafeApp.cpp
nobugshere: Done.

src/ui/wxWidgets/PWSafeApp.cpp
nobugshere: Moved into PWSLocale.h.

src/ui/wxWidgets/PWSLocale.h
nobugshere: Removed the Mac specific class.

src/ui/wxWidgets/PWSafeApp.cpp
dskrvk: Any clues in git blame?

nobugshere: Looks like it originated here, 13 years ago...
https://sourceforge.net/p/passwordsafe/bugs/1105/

Using international characters doesn't cause any apparent problems or assertions for me, either with master or this branch. (On macOS)

dskrvk: Just speculating, maybe this is in case an assertion fails very early in the program's execution, before locale is applied, which was causing a crash in wx's internals? Probably safer to keep this.

nobugshere: Yes, that’s the only thing that makes sense to me, and it must have actually happened to cause someone to worry about it.  I think I’ll change the ifdef back to __WXDEBUG__ to match the OnAssertFailure doc.

src/ui/wxWidgets/PWSafeApp.cpp
ronys, line 605: If we've already created a PWSLocale class, wouldn't it make more sense to move this logic there as well?

src/ui/wxWidgets/PWSafeApp.cpp
```diff
@@ -712,40 +674,39 @@ bool PWSafeApp::ActivateLanguage(wxLanguage language, bool tryOnly)
     isHelpActivated = ActivateHelp(language);
 
     const wxLanguageInfo *langInfo = nullptr;
-    langInfo = wxLocale::GetLanguageInfo(language);
+    langInfo = PWSLocale::GetLanguageInfo(language);
```
dskrvk: Nit: we could combine declaration and initialization

src/ui/wxWidgets/PWSafeApp.cpp
```diff
@@ -712,40 +674,39 @@ bool PWSafeApp::ActivateLanguage(wxLanguage language, bool tryOnly)
...
+      /*
+       * Some languages have multiple locale variations.  (e.g. en_US, en_GB, etc.)
+       * Just using the two letter language identifier (e.g. en.UTF-8) does not work
+       * in some cases; there needs to be a region as well (e.g. en_US.UTF-8), not doing
+       * so causes some inconsistent results. Specifically, the date format seems to
+       * default to en_GB in WX but not in native macOS controls, such as the date picker.
+       * This checks the user environment setting, if the language matches, use the env locale.
+       * If not, use whatever WX guessed.
+       */
       wxString envString;
```
dskrvk: Maybe we should refactor the logic of determining envName into `wxLocale` (e.g. a separate private method called from `UseLocaleName`, so the user doesn't have to pass this as a parameter)?

nobugshere: I moved the logic into PWSLocale and renamed the function to PWSLocale::ChooseLocale().  I thought that was more descriptive and differentiates it from wxUILocale.  I still need to do more testing.

src/ui/wxWidgets/PWSafeApp.cpp
nobugshere: I'm not so sure.  At a high level, I wanted to have some separation between language and locale, though they are intertwined. Maybe the language handling should be its own class, including the menu handling?  If you feel strongly about it, I'll pull more of it in here.

src/ui/wxWidgets/PWSafeApp.cpp
```diff
@@ -219,24 +214,31 @@ PWSafeApp::~PWSafeApp()
 void PWSafeApp::Init()
 {
   pws_os::install_cleanup_handler(cleanup_handler, &m_core);
-  m_locale = new wxLocale;
-  wxLocale::AddCatalogLookupPathPrefix(L"/usr/share/locale");
-  wxLocale::AddCatalogLookupPathPrefix(L"/usr");
-  wxLocale::AddCatalogLookupPathPrefix(L"/usr/local");
+
+#ifndef LOCALE_WX322
+  m_locale = PWSLocale::m_pwslocale = new PWSLocale;
```
ronys: This seems a bit redundant. Do we need both m_locale and PWSLocale::m_pwslocale?
I think the singleton design pattern might be more appropriate here, and consistent with how we work with PWSprefs.

src/ui/wxWidgets/PWSafeApp.cpp
ronys: I see your point.
Let's leave this as-is then, at least for this PR.

src/ui/wxWidgets/PWSafeApp.cpp
nobugshere: I've already changed this.  There was a bit of an issue using dynamic members in static functions.  I think it's better now.

PR #1919, opened 2026-08-21
Linux: repaint tray icon on next interaction after a transient IsAvailable() drop

### LINKED ISSUE #1905: System tray icon doesn't repaint on idle-timeout auto-lock (stuck on unlocked) when screen is DPMS-blanked
## Summary

On Linux (wxWidgets UI), when "Lock password database after being idle" fires *while the screen is DPMS-blanked* (i.e. the display is off due to its own, independent idle timeout at that exact moment), the database is correctly locked, but the system tray icon does not repaint to the locked state. It stays showing the last-painted (unlocked) icon indefinitely, even though internal state (`SystemTray::m_status`, `PasswordSafeFrame::IsLocked()`) is correctly `LOCKED`. If the screen is on (not DPMS-blanked) when the idle-lock fires, or for a manual "Lock" (menu/toolbar/tray), the icon repaints correctly.

## Root cause

`SystemTray::SetTrayStatus()` has an early return guarded by `IsTaskBarIconAvailable()`:

https://github.com/pwsafe/pwsafe/blob/master/src/ui/wxWidgets/SystemTray.cpp#L102-L107

```cpp
void SystemTray::SetTrayStatus(TrayStatus status)
{
  m_status = status;

  if (!IsTaskBarIconAvailable())
    return;
  ...
```

`IsTaskBarIconAvailable()` (`wxUtilities.cpp`) ultimately calls `wxTaskBarIcon::IsAvailable()`, which on X11/XWayland checks live ownership of the `_NET_SYSTEM_TRAY_S<screen>` selection (`src/gtk/taskbar.cpp` in wxWidgets). This guard was added in 77c1814a2 ("Wrokaround wx bug") to avoid a wx crash on platforms/WMs that never support a system tray at all — a permanent-environment check.

The problem: this same check is re-evaluated on *every* `SetTrayStatus()` call, including transient states — e.g. while the screen is DPMS-blanked (screensaver/idle), the tray-manager selection can transiently report unowned. When the idle-lock timer fires while the screen is blanked, `SetTrayStatus(LOCKED)` sets `m_status` correctly but the early return skips `SetIcon()`/`SetSNIIconName()`, so the visible icon is never repainted to reflect the lock. Nothing re-triggers a repaint later, so the icon can remain stuck on "unlocked" until some unrelated status transition happens to succeed.

## Repro (confirmed via a locally instrumented debug build with added trace logging)

1. Open a database, enable system tray icon.
2. Options → Security → enable "Lock Database" after N minutes idle.
3. Set desktop idle/DPMS screen blank to a shorter time than N (so the screen is off when the DB lock fires).
4. Wait for the idle lock to fire.

Trace output at the moment of lock:
```
LockDb: entry
LockDb: SaveAndClearDatabaseOnLock()=1
SetTrayStatus: status=2 (LOCKED), IsTaskBarIconAvailable()=0
```
DB is locked (`SaveAndClearDatabaseOnLock()` succeeds), but `IsTaskBarIconAvailable()` returns `0`, so the repaint is skipped. A manual Lock (tray/menu), tested with the screen on, goes through the same code path with `IsTaskBarIconAvailable()=1` and repaints correctly — confirming the divergence is this transient-availability check, not a separate code path for idle-lock vs. manual lock.

## Environment
- Fedora 42, GNOME Shell on Xorg (X11 session, confirmed via `loginctl show-session -p Type` = `x11`), with `appindicatorsupport@rgcjonas.gmail.com` extension bridging the legacy XEmbed tray icon
- wxWidgets 3.2.8, pwsafe master (`c53ad7872`)
- `gsettings get org.gnome.desktop.session idle-delay` = 600s (screen blank only, session lock disabled)

## Possible fixes (would appreciate maintainer input before submitting a PR)

- **Retry with backoff**: when the repaint is skipped due to unavailability, keep `m_status` current and arm a short-lived timer to retry the repaint until it succeeds (capped, so it gives up cleanly on WMs with no tray at all).
- **Repaint on next natural event**: re-assert `SetTrayStatus(m_status)` whenever the user next interacts with the window/tray (restore, unlock, click) — simpler, but the icon stays stale until interaction even if the screen wakes on its own.

Happy to submit a PR once there's agreement on which approach (or another) is preferred, since `77c1814a2` was a deliberate crash workaround and I'd rather not weaken/remove it carelessly.

ykne:
## Summary
Fixes #1905. `SetTrayStatus()` re-checks `IsTaskBarIconAvailable()` on
every call but skips `SetIcon()` when it's false, and never gets called
again once availability returns (since the tracked status itself didn't
change) — leaving the tray icon showing stale state, e.g. after the
screen is DPMS-blanked during idle-timeout auto-lock.

Per the discussion on the issue, this implements the "repaint on next
natural event" option: `SetTrayStatus(m_status)` is re-run at the top of
the handlers that only fire when the user actually interacts with the
tray icon, so the icon self-heals without caching `IsAvailable()`'s
result (which risks the crash `77c1814a2` fixed) or adding a recurring
polling timer.

Which handlers are reachable depends on the backend:
- `OnTaskBarLeftDoubleClick()` and `OnSysTrayMenuItem()` cover
  double-click and menu-item selection on every backend.
- `CreatePopupMenu()` additionally covers right-click, but only when
  there's no AppIndicator/SNI backend in play — e.g. stock Fedora 42's
  wx 3.2, which doesn't compile in `wxUSE_APPINDICATOR` at all. On that
  path, `wxTaskBarIconBase::OnRightButtonDown()` falls back to
  `CreatePopupMenu()` since `GetPopupMenu()` isn't overridden.

Right-click can *not* be caught this way under the AppIndicator/SNI
backend itself (e.g. wx 3.3 on Wayland): the menu is exported once via
a one-time D-Bus dump inside `app_indicator_set_menu()`, and the shell
renders it from that without ever calling back into the app —
`GetPopupMenu()` is only invoked internally by `SetIcon()`, never by an
actual click. So on that backend, a stale icon only self-heals via
double-click or by actually selecting a menu item, not by merely
opening the menu with a right-click. (The AppIndicator/StatusNotifierItem
protocol has no hover/mouse-move signal at all either — only
`Activate`/`SecondaryActivate`/`Scroll` — so there's no better hook
available there.)

## Test plan
- [x] Clean mock build against stock Fedora 42 repo packages
- [x] Clean mock build overlaying wx 3.3.4-4 + libayatana-appindicator 0.6.0-1
- [x] Manual repro on Wayland (wx 3.3/AppIndicator): locked DB via
      idle-timeout while screen was DPMS-blanked, woke the screen,
      confirmed the icon stayed stale until double-clicking, at which
      point it repainted correctly
- [x] Manual repro on Fedora 42 (no AppIndicator): confirmed left-click
      and right-click (opening the menu) both repaint a stale icon

### DIFF
--- a/src/ui/wxWidgets/SystemTray.cpp
+++ b/src/ui/wxWidgets/SystemTray.cpp
@@ -90,12 +90,13 @@ SystemTray::SystemTray(PasswordSafeFrame* frame) : m_TrayIconWithOverlay(false),
 
 void SystemTray::SetTrayStatus(TrayStatus status)
 {
+  if (m_inSetTrayStatus)
+    return;
+  m_inSetTrayStatus = true;
+
   m_status = status;
 
-  if (!IsTaskBarIconAvailable())
-    return;
-
-  if (PWSprefs::GetInstance()->GetPref(PWSprefs::UseSystemTray)) {
+  if (IsTaskBarIconAvailable() && PWSprefs::GetInstance()->GetPref(PWSprefs::UseSystemTray)) {
      switch(status) {
        case TrayStatus::CLOSED:
          SetIcon(m_IconClosed, wxTheApp->GetAppName());
@@ -113,6 +114,8 @@ void SystemTray::SetTrayStatus(TrayStatus status)
          break;
      }
   }
+
+  m_inSetTrayStatus = false;
 }
 
 wxMenu* SystemTray::RepopulateMenu(wxMenu* menu)
@@ -203,10 +206,12 @@ wxMenu* SystemTray::GetPopupMenu()
   // Also deliberately does not re-run SetTrayStatus() the way the other
   // handlers do to self-heal a stale icon: this is called *from inside*
   // SetIcon() itself (via SetTrayStatus() -> SetIcon() -> here), so doing
-  // so would recurse. It's also moot here either way - opening the menu
-  // under this backend never reaches this method in the first place (the
-  // shell renders it from a one-time D-Bus export, with no callback into
-  // the app), so there's no interaction to catch.
+  // so would recurse (m_inSetTrayStatus guards against that actually
+  // happening, but it's not a reason to add the call - see below). It's
+  // also moot here either way - opening the menu under this backend never
+  // reaches this method in the first place (the shell renders it from a
+  // one-time D-Bus export, with no callback into the app), so there's no
+  // interaction to catch.
   m_sniMenu = RepopulateMenu(m_sniMenu);
   return m_sniMenu;
 }
--- a/src/ui/wxWidgets/SystemTray.cpp
+++ b/src/ui/wxWidgets/SystemTray.cpp
@@ -264,6 +264,13 @@ void SystemTray::OnSysTrayMenuItem(wxCommandEvent& evt)
 {
   EventHandlerDisabler ehd(this);
 
+  // IsTaskBarIconAvailable() can transiently report false (e.g. while the
+  // screen is DPMS-blanked), which makes SetTrayStatus() skip SetIcon()
+  // and leave the icon/menu stale once availability returns, since nothing
+  // else re-triggers a repaint. Re-run it here so the icon self-heals the
+  // next time the user actually interacts with it, without polling.
+  SetTrayStatus(m_status);
+
   const int id = evt.GetId();
   if (IsRUECommand(id)) {
     RUEOperation opn = GetRUEOperation(id);
@@ -363,5 +370,8 @@ void SystemTray::ProcessSysTrayMenuItem(int itemId)
 
 void SystemTray::OnTaskBarLeftDoubleClick(wxTaskBarIconEvent& WXUNUSED(evt))
 {
+  // See the comment in OnSysTrayMenuItem() re: self-healing a stale icon.
+  SetTrayStatus(m_status);
+
   wxTheApp->CallAfter([this](){ ProcessSysTrayMenuItem(ID_SYSTRAY_RESTORE); });
 }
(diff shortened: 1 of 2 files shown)


src/ui/wxWidgets/SystemTray.cpp
```diff
@@ -192,6 +199,14 @@ wxMenu* SystemTray::GetPopupMenu()
   // the same instance reused/repopulated rather than a fresh one each
   // time. Deliberately does not go through CreatePopupMenu(), whose
   // reparent-and-display-now logic exists only for actual click handling.
+  //
+  // Also deliberately does not re-run SetTrayStatus() the way the other
+  // handlers do to self-heal a stale icon: this is called *from inside*
+  // SetIcon() itself (via SetTrayStatus() -> SetIcon() -> here), so doing
+  // so would recurse. It's also moot here either way - opening the menu
```
dskrvk: Is there any safety check we can do to prevent accidental recursion (in case things change in the future)?

ykne: Good catch — right now that's only a comment, not an actual guard. Added a small reentrancy guard: a `m_inSetTrayStatus` bool that `SetTrayStatus()` checks and sets on entry, clearing on exit. If anything ever did call `SetTrayStatus()` from inside `GetPopupMenu()` (or anywhere else in that call chain) in the future, it'd now be a silent no-op instead of infinite recursion.

For context on why it isn't reachable today: `GetPopupMenu()` is only ever invoked by wx's own `Private::SetIcon()`, from inside `SetTrayStatus()`'s call to `SetIcon()` — never directly by a user click, since under the AppIndicator/SNI backend the menu is exported once via a one-time D-Bus dump and the shell renders it without calling back into the app. So there's no existing code path that recurses; the guard is purely insurance against a future change reintroducing one. Pushed the update now — let me know if you'd rather see it done differently.

dskrvk (approved):

PR #1921, opened 2026-08-21
[wx] Release build fails on macOS with Yubikey disabled

froggie2523:
This PR provides a fix that allows the build to compile successfully on macOS with Yubikey support disabled (Release configuration).

macOS (Release configuration) - fails with an error

% xcodebuild -project pwsafe-xcode6.xcodeproj -scheme pwsafe -configuration Release
...
In file included from /Users/ja/Documents/pwsafe/src/ui/wxWidgets/DbSelectionPanel.cpp:28:
/Users/ja/Documents/pwsafe/src/ui/wxWidgets/DbSelectionPanel.h:82:7: error: private field 'm_confirmationButtonId' is not used
      [-Werror,-Wunused-private-field]
   82 |   int m_confirmationButtonId;
      |       ^
**1 error generated.**

macOS (Debug-no-yubi configuration), FreeBSD and OpenBSD - succeeds with warning

Example - FreeBSD:
$ cmake -D wxWidgets_CONFIG_EXECUTABLE=/usr/local/bin/wxgtk3u-3.2-config -D NO_YUBI=ON ..
$ gmake
/home/ja/Documents/pwsafe/src/ui/wxWidgets/DbSelectionPanel.h:82:7: warning: private field 'm_confirmationButtonId' is not used [-Wunused-private-field]
   82 |   int m_confirmationButtonId;
      |       ^
1 warning generated.

### DIFF
--- a/src/ui/wxWidgets/DbSelectionPanel.h
+++ b/src/ui/wxWidgets/DbSelectionPanel.h
@@ -79,7 +79,7 @@ private:
   SafeCombinationCtrl* m_sc;
   bool m_bAutoValidate;
   PWScore* m_core;
-  int m_confirmationButtonId;
+  [[maybe_unused]] int m_confirmationButtonId;
   StringX m_yubiCombination; // needed to adjust TransferDataFromWindow()
 };
 
--- a/src/ui/wxWidgets/DbSelectionPanel.h
+++ b/src/ui/wxWidgets/DbSelectionPanel.h
@@ -79,7 +79,13 @@ private:
   SafeCombinationCtrl* m_sc;
   bool m_bAutoValidate;
   PWScore* m_core;
+#if defined(__clang__)
   [[maybe_unused]] int m_confirmationButtonId;
+#elif defined(__GNUC__)
+  int m_confirmationButtonId;
+#else
+  int m_confirmationButtonId;
+#endif
   StringX m_yubiCombination; // needed to adjust TransferDataFromWindow()
 };
 
--- a/src/ui/wxWidgets/DbSelectionPanel.h
+++ b/src/ui/wxWidgets/DbSelectionPanel.h
@@ -81,8 +81,6 @@ private:
   PWScore* m_core;
 #if defined(__clang__)
   [[maybe_unused]] int m_confirmationButtonId;
-#elif defined(__GNUC__)
-  int m_confirmationButtonId;
 #else
   int m_confirmationButtonId;
 #endif


dskrvk (approved):

froggie2523:
I've pushed the change to address the failing build (ubuntu-22.04) job. Please review it again.

The solution is to use a compiler-specific annotation. Eventually, it will work as follows:
macOS (yubikey support disabled): Uses Clang, which reports the unused private field as a warning. All warnings are treated as errors.
Ubuntu 22: Uses GCC 11.4, which reports the [[maybe_unused]] attribute as a warning. Since -Werror is enabled, the warning is promoted to an error.
FreeBSD and OpenBSD (yubikey support disabled): Use Clang; there is no longer a warning.
Others: no change

rafaelx:
It seems like `m_confirmationButtonId` is always initialized with `wxID_OK` and the flexibility to use a custom button ID in the application isn't being used. Wouldn't it be simpler to just remove `m_confirmationButtonId` as well as `buttonConfirmationId` from the implementation of `DbSelectionPanel`?

nobugshere:
> Wouldn't it be simpler to just remove ....

Since it's not really being used as intended, I'd say that removing it is a cleaner solution to the original issue than ifdef'ing the compiler version.

dskrvk:
Not quite, it's initialized to a different value here: https://github.com/pwsafe/pwsafe/blob/master/src/ui/wxWidgets/CompareDlg.cpp#L227-L232

rafaelx:
> Not quite, it's initialized to a different value here: https://github.com/pwsafe/pwsafe/blob/master/src/ui/wxWidgets/CompareDlg.cpp#L227-L232

Ah, a multi-line constructor call. I didn't take that into account when using grep. What does the ID 1 stand for?

dskrvk:
> What does the ID 1 stand for?

It's a row separator size.

nobugshere:
> Ah, a multi-line constructor call. I didn't take that into account when using grep. 

Yup. I missed it, too.

src/ui/wxWidgets/DbSelectionPanel.h
```diff
@@ -79,7 +79,13 @@ class DbSelectionPanel : public wxPanel, private YubiMixin
   SafeCombinationCtrl* m_sc;
   bool m_bAutoValidate;
   PWScore* m_core;
+#if defined(__clang__)
+  [[maybe_unused]] int m_confirmationButtonId;
+#elif defined(__GNUC__)
   int m_confirmationButtonId;
+#else
+  int m_confirmationButtonId;
```
nobugshere: Do you really need the #elif clause?

froggie2523: The __GNUC__ branch isn't actually necessary if GCC is the only other compiler, but in theory, one could compile the wxWidgets version for Windows using Visual Studio.

nobugshere: Sure, but at least for now, the #elif and #else branches are the same.

rafaelx: To avoid some branching or reduce the branching the implementation could check what the used compiler supports.

Solution A)

```
#if defined(__has_cpp_attribute) && __has_cpp_attribute(maybe_unused)
  [[maybe_unused]] int m_confirmationButtonId;
#else
  int m_confirmationButtonId;
#endif
```

To completely avoid branches and improve readability, a macro for multiple uses could be introduced.

Solution B)

```
#if defined(__has_cpp_attribute)
#  if __has_cpp_attribute(maybe_unused)
#    define PWSAFE_MAYBE_UNUSED [[maybe_unused]]
#  else
#    define PWSAFE_MAYBE_UNUSED
#  endif
#else
#  define PWSAFE_MAYBE_UNUSED
#endif

PWSAFE_MAYBE_UNUSED int m_confirmationButtonId;
```

and maybe with:

```
#if defined(PWSAFE_MAYBE_UNUSED)
#  undef PWSAFE_MAYBE_UNUSED
#endif
```

nobugshere (approved):

Issue #1923, opened 2026-08-22
[wx] Assertions during drag-and-drop operations


froggie2523:
Since PR #1907, I've started seeing crashes/assertions. I haven't had time to investigate further, but the issue is easily reproducible.

The assertions/silent crashes occur across several drag-and-drop mechanisms, from time to time:
-moving items within a wxTree
-copying entries between two app instances using the D&D toolbar
-using the D&D toolbar to copy data from an entry

Versions tested: the latest source code, v1.24 and 1.21.1 with the PR #1907 changes applied.

Tested on:
Ubuntu 22.04 + wxWidgets 3.0.5
Debian 13 + wxWidgets 3.2.8
macOS 15 + wxWidgets 3.2.9

<img width="1966" height="1066" alt="Image" src="https://github.com/user-attachments/assets/aeaf9ae3-cd5e-406c-81c4-7defd7d082f4" />

nobugshere:
Edit: This problem is unrelated to this issue and will get its own PR.

Using a release build from the current master w/wx3.2.10 I couldn't reproduce a problem like yours.  But, using a debug build, I got this when I use one of the drag bar buttons to drag a field to a text editor, then come back and click a second drag-bar button:

<img width="704" height="345" alt="Image" src="https://github.com/user-attachments/assets/69afff19-bd7d-48fe-9df5-6f31b3b015ac" />

Clicking the "see details" button just crashes the program.
macOS M1Max 15.7.9 Sequoia,                 Xcode 16.4, wx 3.2.10 (local build)

PR #1924, opened 2026-08-22
[wx] Fix for assertions during drag-and-drop operations

froggie2523:
Resolves issue #1923, related to PR #1907.

According to the wxWidgets documentation (e.g. https://docs.wxwidgets.org/3.2/group__group__funcmacro__debug.html):
Starting with wxWidgets 2.9.1, debugging support in wxWidgets is always compiled in by default, you need to explicitly define wxDEBUG_LEVEL as 0 to disable it completely. However, by default debugging macros are dormant in the release builds. 

My understanding is that wxDEBUG_LEVEL=0 can only be set when both wxWidgets and Password Safe are built together, which is true only in some cases, such as during the Flatpak build or building a package on macOS.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -276,7 +276,7 @@ endif(HAVE_YKPERS_H)
 if (NOT MSVC)
    # Following is because (a) -O3 breaks the test and application, and
    # (b) -O3 is the default for cmake
-   set (CMAKE_CXX_FLAGS_RELEASE "-O2 -DNDEBUG -DwxDEBUG_LEVEL=0")
+   set (CMAKE_CXX_FLAGS_RELEASE "-O2 -DNDEBUG")
 endif (NOT MSVC)
 
 include(pws-version)


nobugshere:
You might be right about wxDEBUG_LEVEL (it defaults to 1).  But, if the lib and app are built with level == 1, I’m not sure what you mean by:

> However, by default debugging macros are dormant in the release builds.

Edit: I see.  That means with NDEBUG defined.

nobugshere:
The wxWidgets documentation also says this:
"Also notice that it is possible to build your own application with a different value of wxDEBUG_LEVEL than the one which was used for wxWidgets itself. E.g. you may be using an official binary version of the library which will have been compiled with default
[wxDEBUG_LEVEL](https://docs.wxwidgets.org/3.2.4/group__group__funcmacro__debug.html#ga7ce424d27d39bb92c26c80b6a1903725) == 1 
but still predefine wxDEBUG_LEVEL as 0 for your own code."
https://docs.wxwidgets.org/3.2.11/overview_debugging.html

However, after some testing on Fedora 44, I agree that the proposed change makes the assertions go away (NDEBUG), but it may just be masking the issue.  When I do a full debug build (cmake --fresh -DCMAKE_BUILD_TYPE=Debug) I get the same GTK assertion you posted in #1923, but it looks like something between WX and GTK.  It's strange that adding wxDEBUG_LEVEL=0 enables an assertion in GTK, but there it is.

I do not see this assert on FreeBSD.

Anyway, I agree we can merge this change to fix the Linux release builds.  The other changes in PR #1907 cover my concerns over using `__WXDEBUG__` to gate our debug code.

```

```

PR #1926, opened 2026-08-25
Fixed ATTIV value (0x72, instead of 0x71)

medo64:
In V4 documentation, both AttAK and AttAK share the same value (0x71)
```
AttEK                       0x70        32 bytes      Y              [8]
AttAK                       0x71        32 bytes      Y              [9]
ATTIV                       0x71        binary        Y              [10]
Content                     0x73        4 byte+binary Y              [11]
```

I believe ATTIV was intended to have value 0x72.


ronys:
Nice catch - thanks!

PR #1928, opened 2026-08-26
Add VKShowTooltips Boolean preference

ronys:
Users can now turn of the tooltips displayed in the Virtual Keyboard by clearing the new "Display Tooltips" checkbox. This is useful when the tooltips hide some of the buttons.
(As requested by Benjamin via email)


dskrvk:
Hm, I wonder whether this is the right solution. If I generally use tooltips, going into settings to turn tooltips off and on when accessing certain screens would be cumbersome. Two alternatives:
1. See if we can affect the tooltip positioning so it never obscures the element it belongs to.
2. Add a "tooltips" checkbox (without persistence) just on the VK screen and not introduce a global preference until there's a clear need for it. Considering that in this PR we can in any case only control tooltips on a single screen this proposed approach might be less confusing to users.

Thoughts?

ronys:
Thanks for the feedback.

I'm keeping the new preference using the general prefs mechanism for convenience. The "VK" prefix in the name hints to the fact that it's only relevant to the virtual keyboard, similar to the existing `VKPlaySound` preference.
For the regular PasswordSafe tooltips, I don't think this is an issue, as the dialogs are large enough for the tooltips not to get in the way. The virtual keyboard, however, is pretty crowded (by design), and I can see how the tooltips there both get in the way, and are pretty much useless to a regular user of this feature. That's why I opted for a persistent way to turn them off specifically in the virtual keyboard window.

Affecting the tooltip positioning requires a (very) deep dive into the internals, which I don't think is justified.

Another option I've considered, which is simpler to implement, is to change the time delay between when the cursor is on a control and the tooltip is displayed. However, I don't see a way to find a widely acceptable "sweet spot" for all users.

docs/ReleaseNotes.md
```diff
@@ -15,6 +15,10 @@ Bugs Fixed in 3.72.2
 * [GH1889](https://github.com/pwsafe/pwsafe/issues/1889) Regression: 'Apply' button now enabled only when a change is made to the entry.
 * [GH1092](https://github.com/pwsafe/pwsafe/issues/1092) Regression: Last size and position maintained correctly across scaling other than 100%.
 
+New Features in 3.72.2
+-------------------------
+* Users can now turn of the tooltips displayed in the Virtual Keyboard by clearing the new "Display Tooltips" checkbox. This is useful when the tooltips hide some of the buttons.
```
dskrvk, line 20:
```suggestion
* Users can now turn off the tooltips displayed in the Virtual Keyboard by clearing the new "Display Tooltips" checkbox. This is useful when the tooltips hide some of the buttons.
```

src/core/PWSprefs.cpp
```diff
@@ -141,6 +141,7 @@ const PWSprefs::boolPref PWSprefs::m_bool_prefs[NumBoolPrefs] = {
   {_T("ExcludeFromClipboardHistory"), true, ptDatabase},    // database
   {_T("FindToolBarActive"), false, ptApplication},          // application
   {_T("ExcludeFromScreenCapture"), true, ptDatabase},       // database
+  {_T("VKShowTooltips"), true, ptApplication},             // application
```
dskrvk, line 144: Why do we need these comments at the end?

src/ui/Windows/VirtualKeyboard/VKresource.h
```diff
@@ -101,5 +101,6 @@
 
 #define IDC_SAVEKLID                   8100
 #define IDC_KEYPRESS_PLAYSOUND         8101
+#define IDC_SHOWTOOLTIPS               8104
```
dskrvk: Can we keep these ordered by numeric value?

dskrvk:
>  The virtual keyboard, however, is pretty crowded (by design), and I can see how the tooltips there both get in the way, and are pretty much useless to a regular user of this feature.

Maybe we could just turn them off unconditionally on that screen? Is there any reason to need them? I don't think they serve any a18y purpose?

src/ui/Windows/VirtualKeyboard/VKresource.h
ronys: Fixed - thanks!

src/core/PWSprefs.cpp
ronys: You're correct, these predate the pt* enum, but I'll leave this for a separate 'cleanup' PR.

PR #1929, opened 2026-08-27
[wx 3.3.x] Assertion in YubiCfgDlg.cpp

nobugshere:
With a YibiKey inserted, select menu Manage->YibiKey...
I get the attached assert when built with wx 3.3.3.1 (does not happen with wx 3.2.x)

The changelog for wx 3.3.x says this under the heading "INCOMPATIBLE CHANGES SINCE 3.2.x:"

- Calling wxTextCtrl::SetLabel() now consistently does nothing and asserts
under all platforms instead of behaving as SetValue() under MSW and doing
nothing elsewhere.

I have a feeling we may find more of these, but I haven't yet :-)

Screenshot 2026-08-27 at 12 58 25 PM

Found and fixed with a debug build on :
macOS M1Max 15.7.9 Sequoia, Xcode 16.4, wx 3.3.3.1

### DIFF
--- a/src/ui/wxWidgets/YubiCfgDlg.cpp
+++ b/src/ui/wxWidgets/YubiCfgDlg.cpp
@@ -350,7 +350,8 @@ void YubiCfgDlg::yubiInserted(void)
     m_yksk = wxEmptyString;
   ReadYubiSN();
   dynamic_cast<wxTextCtrl*>(FindWindow(ID_YK_SERNUM))->ChangeValue(m_yksernum);
-  Validate(); TransferDataToWindow();
+  Validate();
+  TransferDataToWindow();
 }
 
 void YubiCfgDlg::yubiRemoved(void)
--- a/src/ui/wxWidgets/YubiCfgDlg.cpp
+++ b/src/ui/wxWidgets/YubiCfgDlg.cpp
@@ -349,7 +349,7 @@ void YubiCfgDlg::yubiInserted(void)
   } else
     m_yksk = wxEmptyString;
   ReadYubiSN();
-  dynamic_cast<wxTextCtrl*>(FindWindow(ID_YK_SERNUM))->SetValue(m_yksernum);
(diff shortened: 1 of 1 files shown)


rafaelx:
It's worth keeping in mind the difference between SetValue and ChangeValue when making changes to the code. [ChangeValue](https://docs.wxwidgets.org/trunk/classwx_text_entry.html#a8c52ab71f51c8f80556c2c8e763cbca1) marks the control as not-modified and doesn't generate a wxEVT_EVENT event. For internal use, this can be the better choice. SetValue behaves more as if a user had entered text interactively, which the application should be informed about.

dskrvk (approved):
Looks like this is the only such occurrence for a `wxTextCtrl`.

src/ui/wxWidgets/YubiCfgDlg.cpp
```diff
@@ -349,7 +349,7 @@ void YubiCfgDlg::yubiInserted(void)
   } else
     m_yksk = wxEmptyString;
   ReadYubiSN();
-  FindWindow(ID_YK_SERNUM)->SetLabel(m_yksernum);
+  dynamic_cast<wxTextCtrl*>(FindWindow(ID_YK_SERNUM))->ChangeValue(m_yksernum);
   Validate(); TransferDataToWindow();
```
dskrvk: Nit: would be good to not put these calls on separate lines

nobugshere: I'm not sure what you're referring to.  Did you mean _**to put**_ Validate and TransferDataToWindow on separate lines?

dskrvk: Oops, yes, exactly

nobugshere: Okay.  Done.

PR #1940, opened 2026-09-04
Fix missing XML schema error when reading filter files

dskrvk:
It's possible these calls relied on some defaults, but on my system they
 result in an XML "schema not found" error.
Changed them to align with existing calls that specify the schema explicitly.
Also, remove an obsolete version workaround.

The original support was added in https://github.com/pwsafe/pwsafe/pull/746 - perhaps it only covered pugi?

## Testing

Linux only (xercesc 3.2.4, wxwidgets 3.2.10), I don't have access to Win/Mac:

* Import/export XML filters
* Import/export pwsafe files as XML
* Autoload filters load when a file is opened

### DIFF
--- a/src/core/XML/Xerces/XFilterXMLProcessor.cpp
+++ b/src/core/XML/Xerces/XFilterXMLProcessor.cpp
@@ -121,11 +121,6 @@ bool XFilterXMLProcessor::Process(const bool &bvalidation,
   pSAX2Parser->setContentHandler(pSAX2Handler);
   pSAX2Parser->setErrorHandler(pSAX2Handler);
 
-  // Workaround/bypass until Xerces supports retrieving version from the
-  // <xs:schema ...> statement!
-  // Set 'dummy' schema version to arbitrary value > 1
-  pSAX2Handler->SetSchemaVersion(99);
-
   pSAX2Handler->SetVariables(m_pAsker, &m_MapXMLFilters, m_FPool, m_bValidation);
 
   // instantiate converter out of if/else to be sure that string will be valid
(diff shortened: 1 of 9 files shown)


src/core/XML/Xerces/XFilterXMLProcessor.cpp
```diff
@@ -121,11 +121,6 @@ bool XFilterXMLProcessor::Process(const bool &bvalidation,
   pSAX2Parser->setContentHandler(pSAX2Handler);
   pSAX2Parser->setErrorHandler(pSAX2Handler);
 
-  // Workaround/bypass until Xerces supports retrieving version from the
-  // <xs:schema ...> statement!
-  // Set 'dummy' schema version to arbitrary value > 1
-  pSAX2Handler->SetSchemaVersion(99);
```
dskrvk, line 127: Checked the release history, but couldn't find when this support was added. Seems to work with my version (3.2.4). The workaround existed since 2008 - seems time to remove it?

src/ui/wxWidgets/PWSafeApp.cpp
```diff
@@ -549,8 +549,9 @@ void PWSafeApp::FinishInit() {
   
   if (pws_os::FileExists(wsAutoLoad)) {
     stringT strErrors = L"";
+    stringT XSDFilename = PWSdirs::GetXMLDir() + PWSFilters::schema_filename;
     int rc = m_frame->ImportFilterXMLFile(FPOOL_AUTOLOAD, L"", wsAutoLoad,
-                                          L"", strErrors, &anAsker, nullptr); // Only summary will be reported
+                                          XSDFilename, strErrors, &anAsker, nullptr); // Only summary will be reported
```
dskrvk, line 554: The commit this was added in says

> Import/export XML filter is working now without XSD file for wxWidgets.

Perhaps the author only finished the pugi-based implementation which doesn't support schemas?

https://github.com/pwsafe/pwsafe/commit/8b5ae2d075b7016e6ab55753b00d67a73c16f0d5#diff-be4bd76224af765f460a625eb581c24cec2c4bdbc4e1b24c6d7975e010bb671aR517

Issue #1938, opened 2026-09-03
Allow an alias to have its own TOTP code


ronys:
(From Bob, via email)

> I found something today in Password Safe that I think may be a bug.  I have a base entry that is configured with username, password, and authenticator secret key.  That works great.  I configured an alias of that base entry for a different system that I log into with the same username/password, but that system gave me a different authenticator secret key, which I entered in the alias.  The problem is that both entries (the base and the alias) both return the TOTP code that is in the base.  The authenticator code that I put in the alias seems to be ignored.  Both the base entry and the alias entry show different authenticator secret keys, but they both return the same TOTP value.
Can you fix (or enhance this) so that Password Safe works so that the alias entries can have an authenticator secret key that is different from the base entry?

@rafaelx previously (#1464) suggested disabling the ability to configure TOTP in an alias. This is an actual use-case for supporting this capability.

PR #1941, opened 2026-09-04
[wx] Allow TOTP configuration for an alias

rafaelx:


rafaelx:
Related to #1938

rafaelx:
If the authentication key is removed from an alias in the "Additional" tab of the Add/Edit dialog and the user switches to the "Basic" tab, the authentication code from the "Base" item is displayed and can be copied to the clipboard via a button. Should the authentication code perhaps not be displayed in this case while the alias is being edited?
Personally, I find it quite practical. :)

src/ui/wxWidgets/PasswordSafeFrame.cpp
```diff
@@ -1458,16 +1458,19 @@ CItemData *PasswordSafeFrame::GetSelectedEntry() const
...
 /**
- * If the selected item is an 'Alias' or a 'Shortcut' the item's base
- * is returned. If the selected item is a 'Normal' item (not of type
- * 'Alias' or 'Shortcut') the item itself is returned.
+ * The selected item is returned if it is a 'Normal' item or a
+ * 'Base' item.
+ * If the selected item is a 'Shortcut' or 'Alias' without TOTP
+ * configuration the item's base is returned.
+ * If the selected item is an 'Alias' with TOTP configuration the
+ * 'Alias' is returned.
```
dskrvk: In the future there may be other fields that could differ in a shortcut entry. Maybe keep this more general, e.g. "If the selected item is a 'Shortcut' or 'Alias' that overrides some fields from the base entry, then the item (rather than its base) is returned."?

src/ui/wxWidgets/PasswordSafeFrame.cpp
```diff
@@ -2083,15 +2086,19 @@ const CItemData* PasswordSafeFrame::GetTotpItem(const CItemData *item) const
...
+  }
+  // Item is 'Normal', 'Base' or 'Alias' with TOTP configuration
+  else {
     return item;
   }
-  return m_core.GetBaseEntry(item);
 }
 
-bool PasswordSafeFrame::IsItemNormalOrBase(const CItemData *item) const
+bool PasswordSafeFrame::IsItemShortcut(const CItemData *item) const
```
dskrvk: Side note: I don't think this method is very valuable. In each instance where it's used we could simply call  `item->IsShortcut()` (the nullptr check is done previously). Even if the consensus is that it's still useful, `PasswordSafeFrame` just isn't the right place - a static method in `CItemData` would be a better choice.

src/ui/wxWidgets/PasswordSafeFrame.cpp
dskrvk, line 2078: Same thing - why should this method be in the UI layer? Could be in PWSCore.

src/ui/wxWidgets/PasswordSafeFrame.cpp
```diff
@@ -2083,15 +2086,19 @@ const CItemData* PasswordSafeFrame::GetTotpItem(const CItemData *item) const
...
 {
-  return item == nullptr ? false : item->IsNormal() || item->IsBase();
+  return item == nullptr ? false : item->IsShortcut();
 }
 
 bool PasswordSafeFrame::HasItemTwoFactorKey(const CItemData *item) const
```
dskrvk: Same about the UI layer

src/ui/wxWidgets/AddEditPropSheetDlg.h
```diff
@@ -393,7 +393,7 @@ class AddEditPropSheetDlg : public wxPropertySheetDialog
   void DisableAuthenticationCodeControls();
   const PasswordSafeFrame* GetPwSafe() const { return wxGetApp().GetPasswordSafeFrame(); }
   bool HasItemTwoFactorKey() const { return GetPwSafe()->HasItemTwoFactorKey(&m_ItemTotp); };
-  bool IsItemNormalOrBase() const { return GetPwSafe()->IsItemNormalOrBase(&m_ItemTotp); }
+  bool IsItemShortcut() const { return GetPwSafe()->IsItemShortcut(&m_ItemTotp); }
   const CItemData *GetTotpItem() const { return GetPwSafe()->GetTotpItem(&m_ItemTotp); };
   int GetTotpCountdownInterval() const { return GetPwSafe()->GetTotpCountdownInterval(); }
```
dskrvk, line 395: IMO, the last 5 methods here are just an unnecessary layer of indirection. We could directly call the corresponding business logic method without going through PWSafeFrame.

src/ui/wxWidgets/AddEditPropSheetDlg.cpp
```diff
@@ -1908,7 +1908,7 @@ void AddEditPropSheetDlg::ItemFieldsToPropSheet()
   m_Email = m_Item.GetEmail().c_str();
   m_Password = m_Item.GetPassword();
 
-  if (IsItemNormalOrBase() && HasItemTwoFactorKey()) {
+  if (!IsItemShortcut() && HasItemTwoFactorKey()) {
     auto twoFactorKey = m_Item.GetTwoFactorKey();
```
dskrvk, line 1912: And this is just an illustration for the previous comment: why does `HasItemTwoFactorKey` has to go through PWSafeFrame, when the cousin - `GetTwoFactorKey` - is invoked directly on `m_Item`?

rafaelx:
1. Review comments implemented as suggested.
2. Renamed 'CItemData::IsTotpActive' to 'CItemData::HasTwoFactorKey'. Does this affect the Windows version?
3. If the Alias in the "Add/Edit" dialog does not have an authentication key, no authentication code is shown; this better reflects the Alias's configuration. Displaying the Base's authentication code could be unexpected from the user's perspective if no authentication key is provided.

dskrvk:
> Renamed 'CItemData::IsTotpActive' to 'CItemData::HasTwoFactorKey'. Does this affect the Windows version?

I'm not finding any usages in the Win codebase, so we should be good.

dskrvk (approved):

PR #1947, opened 2026-09-15
Fix semver strings

### LINKED ISSUE #1946: Displayed version is confusing and not semver compliant
(from Sal vial email)

The version is displayed as "3.72.02", but referred to as "3.72.2".

Investigation reveals that the former isn't compliant with https://semver.org/#semantic-versioning-200

Leading zeroes in version components should be suppressed,

ronys:
Resolves #1946


src/core/PWScore.cpp
```diff
@@ -144,12 +144,8 @@ void PWScore::SetApplicationNameAndVersion(const stringT &appName,
   int nMajor = HIWORD(dwMajorMinor);
   int nMinor = LOWORD(dwMajorMinor);
```
dskrvk, line 145: Side note: seems overcomplicated to use `MAKELONG(MINORVERSION, MAJORVERSION)` at the (only) call site only to have the method split it back into two numbers. It's not like there's a limit on the number of arguments...

src/ui/Windows/AboutDlg.cpp
dskrvk, line 67: Another side note: IMO it would make sense to make `PWSversion` available in core (making it cross-platform) and have it contain `FormatVersionString` and its variants (with different formats) currently located in the UI layer. "Utils" classes are an anti-pattern after all...

src/ui/Windows/AboutDlg.cpp
```diff
@@ -73,19 +74,12 @@ BOOL CAboutDlg::OnInitDialog()
...
-                        static_cast<LPCWSTR>(Revision));
-  } else {
-    m_appversion.Format(L"%s%sV%d.%02d.%02d%s (%s)", AfxGetAppName(),
-                        static_cast<LPCWSTR>(cs2go),
-                        m_nMajor, m_nMinor, m_nBuild,
-                        static_cast<LPCWSTR>(SpecialBuild),
-                        static_cast<LPCWSTR>(Revision));
-  }
+  const CString version(PWSUtil::FormatVersionString(m_nMajor, m_nMinor, m_nBuild).c_str());
+  m_appversion.Format(L"%s%sV%s%s (%s)", AfxGetAppName(),
```
dskrvk, line 78: I think this could be extracted into Utils/Version as a `FormatVersionString` overload that takes additional components (assuming we can keep things generic between win/wx).

src/ui/Windows/PasskeyEntry.cpp
```diff
@@ -82,11 +82,9 @@ CPasskeyEntry::CPasskeyEntry(CWnd* pParent, const CString& a_filespec, int index
   int nBuild = pPWSver->GetBuild();
   CString csSpecialBuild = pPWSver->GetSpecialBuild();
 
-  if (nBuild == 0)
-    m_appversion.Format(L"V%d.%02d%s", nMajor, nMinor, static_cast<LPCWSTR>(csSpecialBuild));
-  else
-    m_appversion.Format(L"V%d.%02d.%02d%s", nMajor, nMinor, nBuild,
-                        static_cast<LPCWSTR>(csSpecialBuild));
+  const CString version(PWSUtil::FormatVersionString(nMajor, nMinor, nBuild).c_str());
+  m_appversion.Format(L"V%s%s", static_cast<LPCWSTR>(version),
```
dskrvk, line 86: Could we optionally include the special build component in `FormatVersionString` or is it too context-dependent?

src/core/PWScore.cpp
dskrvk: Nvm, looks like the Win version relies on this call style

src/ui/wxWidgets/SafeCombinationEntryDlg.cpp
```diff
@@ -29,6 +29,8 @@
 #include "core/core.h"
 #include "core/PWSdirs.h"
 #include "core/PWSprefs.h"
+#include "core/PWSversion.h"
+#include "core/Util.h"
```
dskrvk, line 33: Unused?

src/ui/wxWidgets/Version.cpp
dskrvk, line 1: Maybe some other time we could get rid of this file now that there's one in core.

dskrvk (approved):

src/ui/wxWidgets/Version.cpp
ronys: Indeed. This PR turned out to be quite a rabbit hole... I'm trying to limit the scope to something reasonable.

src/ui/wxWidgets/Version.cpp
```diff
@@ -23,12 +25,8 @@ const wchar_t *debstr = L"[debug] ";
...
-#if defined(REVISION) && (REVISION != 0)
-const wxString pwsafeVersionString = wxString::Format(wxString(_T("v%d.%.2d.%d (%ls) %ls%ls")),
-                                                      MAJORVERSION, MINORVERSION, REVISION,
-                                                      _T(VCS_VERSION), debstr, SPECIALBUILD);
-#else
-const wxString pwsafeVersionString = wxString::Format(wxString(_T("v%d.%.2d (%ls) %ls%ls")),
-                                                      MAJORVERSION, MINORVERSION,
-                                                      _T(VCS_VERSION), debstr, SPECIALBUILD);
-#endif
+const wxString pwsafeVersionString = wxString::Format(wxString(_T("v%ls%ls (%ls) %ls")),
```
nobugshere, line 28: Should there be a space between the version string and SPECIALBUILD?  (i.e. "v%ls %ls").  The SafeCombinationEntry dialog has one.

src/ui/wxWidgets/SafeCombinationEntryDlg.cpp
```diff
@@ -256,14 +258,10 @@ void SafeCombinationEntryDlg::CreateControls()
...
-#if (REVISION == 0)
-  m_version->SetLabel(wxString::Format(wxT("V%d.%d %ls"),
-                                       MAJORVERSION, MINORVERSION, SPECIALBUILD));
-#else
-  m_version->SetLabel(wxString::Format(wxT("V%d.%d.%d %ls"),
-                                       MAJORVERSION, MINORVERSION,
-                                       REVISION, SPECIALBUILD));
-#endif
+  const PWSversion *pPWSver = PWSversion::GetInstance();
+  m_version->SetLabel(wxString::Format(wxT("V%ls %ls"),
```
nobugshere, line 262: Lower case V here?

nobugshere:
Sorry for the late review :-)

src/ui/wxWidgets/Version.cpp
dskrvk: Yeah, this got hairy...

src/ui/wxWidgets/Version.cpp
ronys: I don't think there should be a space, e.g., "1.26.1pre" rather than "1.26.1 pre". I'll fix SafeCombinationEntry. This will also be consistent with Windows.

src/ui/wxWidgets/SafeCombinationEntryDlg.cpp
ronys: Leaving it uppercase.

PR #1950, opened 2026-09-21
Gh1938 alias totp win

ronys:
This implements #1938 for Windows, as well as:
- Minor refactoring to core for consistency in wx and Windows usage
- Added unit test
- Updated English online help (Windows and wx)
- Updated release notes (Windows and wx)


huven:
I'm trying to think of a use-case, can somebody help me? I read https://github.com/pwsafe/pwsafe/issues/1938 but did not convince me that TOTP deserves special treatment over password.

So, maybe allowing entry of alias-specific secret-key is what needs fixing? So TOTP for alias always refers to base, just like password. That makes sense to me.

For #1938, best option would be to use regular entry, separate password for both logins, which is usually best practice for different IdP.

But it could very well be I'm overlooking a perfectly valid use-case here, so looking forward to feedback.

src/ui/Windows/DboxMain.cpp
```diff
@@ -3576,8 +3576,11 @@ int DboxMain::OnUpdateMenuToolbar(const UINT nID)
       break;
     case ID_MENUITEM_COPY2FAAUTHCODE:
     case ID_MENUITEM_VIEW2FAAUTHCODE:
-      iEnable = bGroupSelected || !pci || pci->IsFieldValueEmpty(CItemData::TWOFACTORKEY, pbci) ? FALSE : TRUE;
+    {
+      const CItemData *pTotpItem = pci ? m_core.GetCredentialEntry(pci) : nullptr;
+      iEnable = (bGroupSelected || !pTotpItem || pTotpItem->GetTwoFactorKeyLength() == 0) ? FALSE : TRUE;
```
dskrvk: Nit: would be more readable this way:

```suggestion
      iEnable = !bGroupSelected && pTotpItem && pTotpItem->HasTwoFactorKey() ? TRUE : FALSE;
```

rafaelx: `pTotpItem->GetTwoFactorKeyLength() == 0` could be replaced by `!pTotpItem->HasTwoFactorKey()`

src/core/PWScore.cpp
```diff
@@ -2653,7 +2653,8 @@ CItemData* PWScore::GetCredentialEntry(const CItemData* pAny)
   ASSERT(pAny != nullptr);
   if (!pAny)
     return nullptr;
-  const CUUID uuidCredential = pAny->IsDependent() ? pAny->GetBaseUUID() : pAny->GetUUID();
+  const bool bUseBase = pAny->IsShortcut() || (pAny->IsAlias() && !pAny->HasTwoFactorKey());
```
rafaelx: This is just the second occurrence here, but it could optionally be replaced by a single function call on a CItemData that reflects the logic `IsShortcut() || (IsAlias() && !HasTwoFactorKey())`. - Just a thought.

src/ui/Windows/AddEdit_Basic.cpp
```diff
@@ -1298,15 +1298,16 @@ void CAddEdit_Basic::OnTwoFactorCodeStaticClicked()
...
-  CSecString twoFactorKey;
-  if (!M_pci() || !M_pci()->IsAlias())
-    twoFactorKey = M_twofactorkey();
-  else {
-    const CItemData* pcbi = M_pcore()->GetBaseEntry(M_pci());
-    if (pcbi != nullptr)
-      twoFactorKey = M_pci()->GetEffectiveFieldValue(CItem::TWOFACTORKEY, pcbi);
-  }
-  return twoFactorKey;
+  if (!M_pci() || M_pci()->IsNormal() || M_pci()->IsBase())
```
rafaelx, line 1301: I'm not very familiar with the Windows code and haven't really tried to understand it in detail. I have a question about the `!M_pci()`.
Why should the authentication key be returned when there is no CItem element? Is that the case when a new element is created through the Add/Edit dialog?

src/ui/Windows/AddEdit_Basic.cpp
```diff
@@ -1298,15 +1298,16 @@ void CAddEdit_Basic::OnTwoFactorCodeStaticClicked()
...
+    return M_twofactorkey();
+
+  if (M_pci()->IsAlias() && !M_twofactorkey().IsEmpty())
```
rafaelx, line 1304: Something like `M_twofactorkey().IsNotEmpty()` would read a bit more naturally, but the current version with `!M_twofactorkey().IsEmpty()` is maybe typical for C++.

src/ui/Windows/AddEdit_PropertyPage.cpp
```diff
@@ -95,7 +95,9 @@ CItemData* CAddEdit_PropertyPage::M_pci_credential()
 {
   if (!M_pci())
     return nullptr;
-  if (!M_pci()->IsAlias())
+  if (M_pci()->IsNormal() || M_pci()->IsBase())
     return M_pci();
-  return M_pcore()->GetBaseEntry(M_pci());
+  if (M_pci()->IsAlias() && !M_twofactorkey().IsEmpty())
```
rafaelx, line 100: Same here about `!M_twofactorkey().IsEmpty()` as in the previous comment.

src/ui/Windows/MainMenu.cpp
```diff
@@ -886,7 +886,8 @@ void DboxMain::CustomiseMenu(CMenu *pPopupMenu, const UINT uiMenuID,
       pPopupMenu->AppendMenu(MF_ENABLED | MF_STRING,
                              ID_MENUITEM_PASSWORDSUBSET, tc_dummy);
 
-      if (!pci->IsFieldValueEmpty(CItemData::TWOFACTORKEY, pbci)) {
+      const CItemData *pTotpItem = m_core.GetCredentialEntry(pci);
+      if (pTotpItem && pTotpItem->GetTwoFactorKeyLength() > 0) {
```
rafaelx: `pTotpItem->GetTwoFactorKeyLength() > 0` could be replaced by `pTotpItem->HasTwoFactorKey()`.

src/ui/Windows/MainMenu.cpp
```diff
@@ -1607,7 +1608,8 @@ void DboxMain::OnContextMenu(CWnd * /* pWnd */, CPoint screen)
     if (pci->IsFieldValueEmpty(CItemData::USER, pbci))
       pPopup->RemoveMenu(ID_MENUITEM_COPYUSERNAME, MF_BYCOMMAND);
 
-    if (pci->IsFieldValueEmpty(CItemData::TWOFACTORKEY, pbci)) {
+    const CItemData *pTotpItem = m_core.GetCredentialEntry(pci);
+    if (!pTotpItem || pTotpItem->GetTwoFactorKeyLength() == 0) {
```
rafaelx: `pTotpItem->GetTwoFactorKeyLength() == 0` could be replaced by `!pTotpItem->HasTwoFactorKey()`.

src/ui/Windows/MainView.cpp
```diff
@@ -548,7 +548,8 @@ void DboxMain::UpdateToolBarForSelectedItem(const CItemData *pci)
       mainTBCtrl.EnableButton(ID_MENUITEM_COPYUSERNAME, TRUE);
     }
 
-    if (pci_entry == NULL || pci_entry->IsFieldValueEmpty(CItemData::TWOFACTORKEY, pbci)) {
+    const CItemData *pTotpItem = pci_entry ? m_core.GetCredentialEntry(pci_entry) : nullptr;
+    if (!pTotpItem || pTotpItem->GetTwoFactorKeyLength() == 0) {
```
rafaelx: `pTotpItem->GetTwoFactorKeyLength() == 0` could be replaced by `!pTotpItem->HasTwoFactorKey()`.

src/ui/wxWidgets/PasswordSafeFrame.cpp
```diff
@@ -2077,18 +2077,7 @@ CItemData* PasswordSafeFrame::GetBaseEntry(const CItemData *item) const
...
-  }
-  // Item is 'Shortcut' or 'Alias' without TOTP configuration
-  if (item->IsShortcut() || (item->IsAlias() && !item->HasTwoFactorKey())) {
-    return m_core.GetBaseEntry(item);
-  }
-  // Item is 'Normal', 'Base' or 'Alias' with TOTP configuration
-  else {
-    return item;
-  }
+  return item ? m_core.GetCredentialEntry(item) : nullptr;
```
rafaelx: Really nice improvement due to the changes in the core. 👍

src/ui/wxWidgets/PasswordSafeFrame.cpp
```diff
@@ -2140,9 +2129,10 @@ PWSTotp::TOTP_Result PasswordSafeFrame::GetTwoFactorAuthenticationCode(const CIt
 void PasswordSafeFrame::OnTotpCountdownTimer(wxTimerEvent& WXUNUSED(event))
 {
   auto item = GetSelectedEntry();
-  // No item selected or item with
-  // no TOTP configuration selected
-  if (item == nullptr || !item->HasTwoFactorKey()) {
+  auto totpItem = GetTotpItem(item);
```
rafaelx: 👍

src/ui/wxWidgets/PasswordSafeFrame.cpp
```diff
@@ -2155,17 +2145,17 @@ void PasswordSafeFrame::OnTotpCopyAuthCodeTimer(wxTimerEvent& WXUNUSED(event))
   static StringX s_LatestAuthCode(L"");
   auto isAuthCodeInClipboard = Clipboard::GetInstance()->HasData(s_LatestAuthCode);
   auto item = GetSelectedEntry();
-  // No item selected or item with
-  // no TOTP configuration selected
-  // or new item selected then stop
-  // updating the auth code in clipboard.
+  auto totpItem = GetTotpItem(item);
```
rafaelx: 👍

dskrvk:
> I'm trying to think of a use-case, can somebody help me? I read https://github.com/pwsafe/pwsafe/issues/1938 but did not convince me that TOTP deserves special treatment over password.

These are good points (though I personally haven't yet found a need to use an alias, so may be missing something here). Two database entries either represent the same resource with different names (e.g. one for easy searching by title, the other for automatic matching based on the URL domain like iOS does; or the SSO example from the Help docs) or they represent different resources that happen to share some fields _at a point in time_, but may diverge in the future. This case could be served either by the "duplicate entry" feature or by some kind of default entry template (with maybe other fields besides the default login, as supported today), but either way the two entries would be independent. However, I can't think of a use case where two independent resources would keep the passwords in sync but have separate auth keys. (Note that #1938 talks about _two different systems_.) If the user chose to reuse an existing password (against best practices) they would generally be responsible for manually updating it in both places (whereas aliases make it look like this happens automatically in those external systems).

tl;dr I don't understand the use case for aliases, with or without sharing TOTP keys.

src/core/PWScore.cpp
dskrvk: Right, otherwise we have to repeat this condition in way too many places (what if it changes in the future?). All logic that determines the effective fields (or the presence of overrides) of an ItemData should be contained in that class.

src/ui/wxWidgets/PasswordSafeFrame.cpp
dskrvk: But with this I no longer see the point of defining `PasswordSafeFrame::GetTotpItem`.

rafaelx: Indeed, with the new code in the core, it's basically just a renaming of `GetCredentialEntry` to `GetTotpItem` in the Wx version.

src/ui/wxWidgets/PasswordSafeFrame.cpp
rafaelx: Based on @dskrvk's suggestion, we could just use `m_core.GetCredentialEntry(item)` here.

src/ui/wxWidgets/PasswordSafeFrame.cpp
rafaelx: Based on @dskrvk's suggestion, we could just use `m_core.GetCredentialEntry(item)` here.

huven:
> tl;dr I don't understand the use case for aliases, with or without sharing TOTP keys.

I get the benefit of an alias when using SSO, for instance aliasing LinkedIn, Azure, DevOps, MPC etc. to one M365 login, so you get to copy/autotype the same password and 2FA when viewing each entry, while allowing each entry to carry its own title, URL, notes, etc.

I expect that most users don't use aliases, as they are somewhat less intuitive than shortcuts. Allowing TOTP override but not password override could add to the confusion. And no, I don't think password override should be added for aliases, it is better to just duplicate the entry, as you suggested too.

nobugshere:
> I get the benefit of an alias when using SSO....

Indeed.  The US gov. has been slowly migrating many of their sites to a common authentication service.  I like that I can convert an existing entry to an alias and retain the separate URL and notes.

dskrvk:
But in case of true SSO the participating resources would always rely on a shared TOTP secret key, right?

nobugshere:
I should have been clear: In my case, yes there is just one TOTP key, in the base entry.

ronys:
Thanks to all for a good review - I'll implement your comments shortly.

I agree that the "same password but different auth key" is a bit odd. The main reason I'm OK with implementing this is that it costs very little in terms of UI and has zero backwards compatibility issues.

Discussion point: How do we want to handle Passkey credentials and aliases?

(My own use-case for aliases is where I have 2 separate credit cards issued by the same company, so the login credentials are the same, even though the custom fields are different.)

huven:
> I agree that the "same password but different auth key" is a bit odd. The main reason I'm OK with implementing this is that it costs very little in terms of UI and has zero backwards compatibility issues.

Agree with that. I'm mainly concerned about aliases becoming more difficult to reason about. Each time I encounter them I need to think hard how they differ from regular entries and shortcuts, and this PR makes it harder. But that could be just me 😄 

> Discussion point: How do we want to handle Passkey credentials and aliases?

IIRC, I implemented those to behave similarly as passwords / TOTP, being replacement credentials on their own.

ronys:
> Agree with that. I'm mainly concerned about aliases becoming more difficult to reason about. Each time I encounter them I need to think hard how they differ from regular entries and shortcuts, and this PR makes it harder. But that could be just me 😄
> 

I think of shortcuts as a convenient way to put the same entry under different groups, e.g., GitHub under 'pwsafe' and 'repos'. Aliases are for more advanced users, and therefore not expressed in the menus.

> > Discussion point: How do we want to handle Passkey credentials and aliases?
> 
> IIRC, I implemented those to behave similarly as passwords / TOTP, being replacement credentials on their own.

Make sense. Worth adding to the unit tests?

rafaelx:
According to the alias model, an entry becomes an alias because its password field points to another entry. This also means that an alias can't have its own password as long as it remains an alias: assigning its own password breaks the reference, and the entry becomes a normal entry again.  

The changes in this pull request give users more flexibility with the TOTP fields. The TOTP secret can be overwritten. If it isn’t overwritten, the default behavior applies and the TOTP configuration is inherited.

- if the alias has no TOTP configuration, the base entry's TOTP can be used;
- if the alias has its own TOTP configuration, that configuration can be used;
- if the alias is assigned its own password, it ceases to be an alias and becomes a normal entry.

Regardless of recommended practices for handling login credentials, I see the benefit here as the added flexibility for the user. The alias concept remains intact and compatible with existing workflows, while now covering additional use cases involving TOTP.

This added flexibility is also documented, so users do not have to figure out how it works implicitly or through trial and error. This clarifies the distinction between inherited and alias-specific TOTP configurations, providing users with the information needed to utilize the feature effectively.

src/ui/Windows/MainEdit.cpp
```diff
@@ -2057,7 +2055,7 @@ void DboxMain::CopyDataToClipBoard(ClipboardDataSource cds, const bool bSpecial)
       ASSERT(0);
     }
   } else if (cds.IsDerived() && cds.GetDerivedType() == ClipboardDataSource::AuthCode) {
-    GetTwoFactoryAuthenticationCode(*pci_credential, sxData);
+    GetTwoFactoryAuthenticationCode(*m_core.GetCredentialEntry(pci), sxData);
```
rafaelx: Typo: "GetTwoFactoryAuthentication", instead of "GetTwoFactorAuthentication"

rafaelx (approved):

src/ui/Windows/MainEdit.cpp
ronys: Nice catch! (this typo was in the code since 2024)

huven:
> > > Discussion point: How do we want to handle Passkey credentials and aliases?
> > 
> > 
> > IIRC, I implemented those to behave similarly as passwords / TOTP, being replacement credentials on their own.
> 
> Make sense. Worth adding to the unit tests?

AliasShortcutTest.cpp already tests

```
 for (const CItemData::FieldType ft : passkey_fields) {
    EXPECT_EQ(base.GetEffectiveFieldValue(ft, nullptr),
              al2.GetEffectiveFieldValue(ft, &base));
  }
```

Do you have any other tests in mind?

huven:
Thanks for summarizing the post-PR alias behavior @rafaelx, I appreciate that.

Don't you think the added flexibility might work against the target audience? From an end-user perspective using aliases for SSO, I would expect the application to update all entries (ie. the base entry) whenever I change the credentials, without needing to worry that I need to edit the base entry specifically.

Last year I had to replace multiple 2FA codes since an app I used in the past did not allow exporting them. So there's a real-world use-case where I would expect the application not to fork one entry if I happen to update the totp code in an alias instead of the base entry.

Still haven't found a real-world use case of two systems sharing the same password provider, yet having different 2FA.

As I'm probably fighting a losing battle here, I'll take my chances and argue that the alias-concept should be removed entirely. Just imagine the amount of code cleanup that would mean, less code is less bugs and less maintenance.

As a replacement, the application could check on a password/totp/passkey update whether other entries have the same username+credentials and if so prompt the user if he/she wants to update those too (select/deselect), as those are likely SSO entries. No base entry to maintain/select, no more "you can't delete this entry as it is base". All entries are autonomous, linked by having the same creds.
Maybe there are even smarter schemes, this one came to mind first.

And there's a straightforward migration path for existing aliases.

I realize it's probably not going to happen, though these thoughts might resonate with more readers.

Thanks for listening.

rafaelx:
> Thanks for summarizing the post-PR alias behavior @rafaelx, I appreciate that.
> 
> Don't you think the added flexibility might work against the target audience? From an end-user perspective using aliases for SSO, I would expect the application to update all entries (ie. the base entry) whenever I change the credentials, without needing to worry that I need to edit the base entry specifically.

In the desktop version, an alias and a base entry are independent elements with their own properties in the database. If I remember correctly, as someone who doesn’t use aliases, no properties of a base entry can be changed through an alias. An alias, therefore, is not a kind of clone of another element, nor a reference to all the properties of another element. If the relationship between the two elements is undone by giving the alias its own password, the other properties of both elements remain. - Is it implemented differently in the app version?

To make changes to a base entry, the user can select "Go to base entry" from the context menu of an alias in the desktop version, and then edit it.

> Last year I had to replace multiple 2FA codes since an app I used in the past did not allow exporting them. So there's a real-world use-case where I would expect the application not to fork one entry if I happen to update the totp code in an alias instead of the base entry.

This case would also be perfectly covered if all login data is managed in the base entry. Overwriting the TOTP configuration is optional and would probably only be the use case for some users, like in the case of #1938. The workflow here would also be: right-click on the alias and 'Go to Base Entry' to change the TOTP configuration.

> Still haven't found a real-world use case of two systems sharing the same password provider, yet having different 2FA.
> 
> As I'm probably fighting a losing battle here, I'll take my chances and argue that the alias-concept should be removed entirely. Just imagine the amount of code cleanup that would mean, less code is less bugs and less maintenance.
> 
> As a replacement, the application could check on a password/totp/passkey update whether other entries have the same username+credentials and if so prompt the user if he/she wants to update those too (select/deselect), as those are likely SSO entries. No base entry to maintain/select, no more "you can't delete this entry as it is base". All entries are autonomous, linked by having the same creds. Maybe there are even smarter schemes, this one came to mind first.
> 
> And there's a straightforward migration path for existing aliases.

This would mean that while the code for handling aliases would be removed, new code for update logic would have to be added. I would argue that the alias-related code has proven itself over time, is well integrated, and hasn't shown up as a source of bugs or problems.

I think the alias concept comes from a time when the main purpose of the application was just to manage passwords. So back then, there was only the password to consider as a login concept. With TOTP and passkeys, the situation has changed, and the question that arises from this discussion is how these login concepts should be integrated into the application or the existing data model. Should all login data be treated the same way (symmetric data model, management only through the base entry; alias just inherits these properties) or should it be possible to override certain login data (asymmetric data model)? I think it's a matter of decision.

> I realize it's probably not going to happen, though these thoughts might resonate with more readers.

Maybe a survey among pwsafe users would be helpful. Even if only a minority of users use the alias concept, should they just be ignored? Maybe, for some users, the alias functionality was actually a reason to choose pwsafe as their password manager, because it allows them to implement their password management strategy.

> Thanks for listening.

Thanks for sharing your thoughts.

huven:
> In the desktop version, an alias and a base entry are independent elements with their own properties in the database. If I remember correctly, as someone who doesn’t use aliases, no properties of a base entry can be changed through an alias. An alias, therefore, is not a kind of clone of another element, nor a reference to all the properties of another element. If the relationship between the two elements is undone by giving the alias its own password, the other properties of both elements remain. - Is it implemented differently in the app version?

The iOS app treats aliases the same as the desktop version (for maximum compatibility), with one exception: changes made to the password (or TOTP for that matter) of the alias are automatically forwarded to the base. So the app will not break the alias relationship during entry editing.

> To make changes to a base entry, the user can select "Go to base entry" from the context menu of an alias in the desktop version, and then edit it.

Indeed. The additional cognitive load of a base entry will keep this an advanced feature. The more simple scheme I suggested in my previous message would make automatic SSO updating available to all users.

> I think the alias concept comes from a time when the main purpose of the application was just to manage passwords. So back then, there was only the password to consider as a login concept. With TOTP and passkeys, the situation has changed, and the question that arises from this discussion is how these login concepts should be integrated into the application or the existing data model. Should all login data be treated the same way (symmetric data model, management only through the base entry; alias just inherits these properties) or should it be possible to override certain login data (asymmetric data model)? I think it's a matter of decision.

Fully agree. Back then, it was also common to reuse passwords for different systems/services, requiring some other way to add SSO-relationship between records.

> Maybe a survey among pwsafe users would be helpful. Even if only a minority of users use the alias concept, should they just be ignored? Maybe, for some users, the alias functionality was actually a reason to choose pwsafe as their password manager, because it allows them to implement their password management strategy.

Yep, agree. If aliases are removed, the newly introduced SSO handling should be superior. So it's not about leaving anybody behind, but making SSO handling available to a larger audience.

I'll keep an eye on this PR. If it's merged I'll change the iOS app accordingly for compatibility.
I might add the more simple scheme in the iOS app too 😀 No core or database changes needed.

rafaelx:
> The iOS app treats aliases the same as the desktop version (for maximum compatibility), with one exception: changes made to the password (or TOTP for that matter) of the alias are automatically forwarded to the base. So the app will not break the alias relationship during entry editing.

Ah, some properties of the base entry can be changed via the alias. So this kind of conflicts a bit with the idea in the desktop version. I think I understand your objections better now.

Just out of curiosity. If a password change in the alias is delegated to the base element, how is an alias then created, or the relationship between two elements? (Maybe I should try out the app?!)

huven:
> Just out of curiosity. If a password change in the alias is delegated to the base element, how is an alias then created, or the relationship between two elements? (Maybe I should try out the app?!)

Good question. The iOS app does not offer an option to create an alias. It does for shortcuts, not for aliases.

To get an idea of how much interest there is, I just checked the full ticket history (14 years): there are 2 related questions about the iOS app missing that option.

ronys:
Aliases were indeed designed and implemented a *long* time ago, and if we were starting from scratch, we'd certainly do it differently today (V5, anyone?).

As it's a relatively complex feature to explain to users, we've made the deliberate decision not to support creating aliases directly via the UI (as opposed to shortcuts). The thinking was that if a user reads about it in the online help and understands it well enough to use, then that user is sophisticated enough to benefit from it. Elitist? Perhaps.

I can confirm @huven 's data - the proportion of support requests related to aliases over the years is quite small. Given that, and, as @rafaelx has written, that the alias code is "battle proven", I'd leave it alone for now. There are more interesting things to work on.

PR #1951, opened 2026-09-22
Fix most code scanning issues

dskrvk:
Unsafe XML parser options, wrong format specifiers.
Left a couple of places where the format specifier is `x` since there doesn't seem to be a signed counterpart for it:

1. https://github.com/pwsafe/pwsafe/security/code-scanning/59
2. https://github.com/pwsafe/pwsafe/security/code-scanning/50


ronys:
@dskrvk note that CodeQL raised an alert. Can you address it?

dskrvk:
Yes, will look into it and maybe fix the remaining two issues.

-------- Original Message --------

dskrvk:
@ronys all issues should be addressed now.

